PR #36, opened 2026-03-31
codex: preserve buffered subagent thread labels

### LINKED ISSUE #31: Buffered thread/started notifications can drop subagent names and log raw thread IDs
## Bug

When a task spawns a subagent early enough, the job log can lose the subagent name and fall back to raw thread IDs like `thr_2`.

## Reproduction

1. Run a task that quickly spawns a subagent / collaboration tool turn
2. Inspect the resulting job log
3. Observe entries like:
   - `Starting subagent thr_2 via collaboration tool: wait.`
   - `Subagent thr_2 reasoning: ...`

Instead of the expected nickname / label (for example `design-challenger`).

## Root Cause

`captureTurn()` buffers notifications that arrive before the main turn ID is known. During replay, buffered notifications are filtered through `belongsToTurn(...)`.

That works for turn-scoped items, but it is too strict for thread lifecycle notifications like:

- `thread/started`
- `thread/name/updated`

If one of those arrives early for a subagent thread, it can be dropped before label registration happens. Later log rendering then has no thread label and falls back to the raw thread ID.

## Impact

- Logs and progress output become harder to understand.
- Subagent labeling is unstable and depends on event timing.
- Any future behavior that depends on stored thread labels can inherit the same race.

## Suggested Fix

Replay buffered thread lifecycle notifications unconditionally before applying turn-scoped filtering to the remaining buffered messages.

VOIDXAI:
## Summary
- replay buffered thread lifecycle notifications before turn-scoped filtering
- keep early `thread/started` / `thread/name/updated` events from being dropped
- preserve subagent labels in logs and progress output

## Why
Closes #31.

If a subagent thread is announced before the main turn ID is known, the buffered lifecycle event can be discarded during replay. That leaves the runtime without the label and logs fall back to raw thread IDs like `thr_2`.

## Validation
- manually reproduced the subagent flow with the fake Codex fixture
- verified the job log now records `design-challenger` instead of falling back to `thr_2`

### DIFF
--- a/plugins/codex/scripts/lib/codex.mjs
+++ b/plugins/codex/scripts/lib/codex.mjs
@@ -15,6 +15,8 @@
  *   threadLabels: Map<string, string>,
  *   turnId: string | null,
  *   bufferedNotifications: AppServerNotification[],
+ *   pendingThreadLifecycleNotifications: AppServerNotification[],
+ *   drainingThreadLifecycleNotifications: boolean,
  *   completion: Promise<TurnCaptureState>,
  *   resolveCompletion: (state: TurnCaptureState) => void,
  *   rejectCompletion: (error: unknown) => void,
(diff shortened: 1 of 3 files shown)


dkundel-openai:
Duplicate of #581.

Issue #59, opened 2026-03-31
Review gate setup writes to temp dir, but Stop hook reads from persistent dir


xinnyu:
## Bug

`/codex:setup --enable-review-gate` appears to succeed, but the Stop hook never enforces the gate.

## Root cause

`resolveStateDir()` in `scripts/lib/state.mjs` (L29-44) picks the state directory based on `CLAUDE_PLUGIN_DATA`:

```js
const pluginDataDir = process.env[PLUGIN_DATA_ENV];
const stateRoot = pluginDataDir
  ? path.join(pluginDataDir, "state")
  : FALLBACK_STATE_ROOT_DIR;  // os.tmpdir() + "codex-companion"
```

The setup command runs via **Bash** (invoked by Claude), where `CLAUDE_PLUGIN_DATA` is **not set** → writes `stopReviewGate: true` to the **temp directory**.

The Stop hook runs via **Claude Code's hook system**, where `CLAUDE_PLUGIN_DATA` **is set** → reads from the **persistent plugin data directory**, which still has the default `stopReviewGate: false`.

### Evidence

```
# Temp dir (written by setup) — gate enabled
/var/folders/.../codex-companion/feat-emoji-<hash>/state.json
→ { "config": { "stopReviewGate": true } }

# Persistent dir (read by hook) — gate still disabled
~/.claude/plugins/data/codex-openai-codex/state/feat-emoji-<hash>/state.json
→ { "config": { "stopReviewGate": false } }
```

## Suggested fix

The setup subcommand in `codex-companion.mjs` should resolve the persistent plugin data directory the same way hooks do. Options:

1. Have the skill pass `CLAUDE_PLUGIN_DATA` as an env var when invoking the setup Bash command.
2. Have `codex-companion.mjs setup` detect `~/.claude/plugins/data/codex-openai-codex/` as a fallback when `CLAUDE_PLUGIN_DATA` is unset.
3. Write to **both** directories during setup so either reader finds the correct value.

## Environment

- Claude Code (CLI)
- codex-plugin-cc v1.0.1
- codex-cli 0.117.0
- macOS Darwin 25.4.0

PR #123, opened 2026-04-03
fix: handle EAGAIN in hook readFileSync(0) for non-blocking stdin

### LINKED ISSUE #120: EAGAIN crash in hook scripts: readFileSync(0) fails when stdin is non-blocking
## Description

Both `stop-review-gate-hook.mjs` and `session-lifecycle-hook.mjs` crash intermittently with:

```
Error: EAGAIN: resource temporarily unavailable, read
    at Object.readFileSync (node:fs:449:20)
    at readHookInput (file:///.claude/plugins/marketplaces/openai-codex/plugins/codex/scripts/stop-review-gate-hook.mjs:22:18)
```

## Root Cause

`readHookInput()` uses `fs.readFileSync(0, "utf8")` to read stdin. When Claude Code invokes hooks with stdin as a pipe in non-blocking mode, the kernel returns `EAGAIN` instead of blocking. Node.js synchronous APIs don't retry — they throw.

```js
// Both scripts, line 22-23
function readHookInput() {
  const raw = fs.readFileSync(0, "utf8").trim();  // throws EAGAIN
  if (!raw) { return {}; }
  return JSON.parse(raw);
}
```

## Reproduction

Happens intermittently across Claude Code sessions on macOS (Darwin 25.3.0, Node.js v20.19.6). More frequent when multiple hooks fire in quick succession.

## Suggested Fix

Catch `EAGAIN` and treat as empty input:

```js
function readHookInput() {
  try {
    const raw = fs.readFileSync(0, "utf8").trim();
    if (!raw) return {};
    return JSON.parse(raw);
  } catch (e) {
    if (e.code === 'EAGAIN') return {};
    throw e;
  }
}
```

Both `stop-review-gate-hook.mjs` and `session-lifecycle-hook.mjs` need the same fix.

## Environment

- macOS Darwin 25.3.0
- Node.js v20.19.6
- Plugin version: 1.0.2 (commit 8e403f9)
- Claude Code v2.1.90

tmchow:
## Summary

`readHookInput()` in both hook scripts crashes with `EAGAIN: resource temporarily unavailable` when Claude Code invokes hooks with stdin as a non-blocking pipe. This adds a bounded retry loop that catches `EAGAIN` and retries up to 50 times (500ms total) before falling back to an empty object.

## Changes

Modified `readHookInput()` in both files:
- `plugins/codex/scripts/stop-review-gate-hook.mjs`
- `plugins/codex/scripts/session-lifecycle-hook.mjs`

The retry uses `Atomics.wait(SharedArrayBuffer, ...)` for the 10ms delay when available, with a busy-wait fallback. After 50 failed attempts, returns `{}` (same behavior as empty stdin). Non-EAGAIN errors are rethrown.

Function signature and return type are unchanged.

## Testing

Verified the build passes (`npm run build`). Existing tests on main have 4 pre-existing failures unrelated to this change (caused by `CLAUDE_PLUGIN_DATA` being set in the test environment, which is the root cause of #59).

Fixes #120

### DIFF
--- a/plugins/codex/scripts/session-lifecycle-hook.mjs
+++ b/plugins/codex/scripts/session-lifecycle-hook.mjs
@@ -20,9 +20,14 @@ export const SESSION_ID_ENV = "CODEX_COMPANION_SESSION_ID";
 const PLUGIN_DATA_ENV = "CLAUDE_PLUGIN_DATA";
 
 function readHookInput() {
-  const MAX_RETRIES = 200;
   const RETRY_DELAY_MS = 10;
   const CHUNK_SIZE = 4096;
+  // Time-based deadline for the first byte to arrive. Hooks are always
+  // invoked with stdin piped, so data *will* come — but the writer may be
+  // delayed (e.g. the parent process hasn't flushed yet). 5 s is generous
+  // enough to cover slow writers; the hook's own kill-timeout is the
+  // ultimate backstop if stdin never materialises.
+  const PRE_DATA_TIMEOUT_MS = 5_000;
   const canAtomicsSleep =
     typeof SharedArrayBuffer === "function" &&
     typeof Atomics !== "undefined" &&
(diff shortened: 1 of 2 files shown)


tmchow:
@dkundel-openai the Codex review bot keeps flagging the EAGAIN timeout as too short (was 500ms, bumped to 2s). The concern is that if Claude Code takes >2s to pipe hook input, the hook silently returns empty. Any guidance on what timeout is reasonable here? Happy to adjust further if you have a sense of typical stdin delivery latency for hooks.

PR #125, opened 2026-04-03
fix: fall back to tmpdir state when CLAUDE_PLUGIN_DATA state is empty

### LINKED ISSUE #59: Review gate setup writes to temp dir, but Stop hook reads from persistent dir

tmchow:
## Summary

`/codex:setup --enable-review-gate` writes `stopReviewGate: true` to a temp directory because `CLAUDE_PLUGIN_DATA` isn't set when commands run via Bash. The Stop hook then reads from the persistent plugin data directory (where `CLAUDE_PLUGIN_DATA` points) and never finds the config.

## Changes

`resolveStateDir()` in `plugins/codex/scripts/lib/state.mjs` now checks both locations when `CLAUDE_PLUGIN_DATA` is set. If no `state.json` exists at the plugin data path but one exists in the tmpdir fallback, it returns the tmpdir path. New state still defaults to the plugin data directory.

Also fixes a pre-existing test failure in `state.test.mjs` where the tmpdir assertion failed because the test didn't isolate `CLAUDE_PLUGIN_DATA`. Added a new test covering the fallback behavior.

## Testing

- `npm test`: 62 pass, 3 fail (pre-existing in commands.test.mjs, unrelated to this change. Was 4 failures on main, this PR fixes 1.)
- `npm run build`: passes
- New test: verifies state written without `CLAUDE_PLUGIN_DATA` is found when `CLAUDE_PLUGIN_DATA` is later set

Fixes #59

### DIFF
--- a/plugins/codex/scripts/lib/state.mjs
+++ b/plugins/codex/scripts/lib/state.mjs
@@ -49,24 +49,32 @@ export function resolveStateDir(cwd) {
     // future reads/writes use the plugin data dir and state survives tmp cleanup.
     if (!fs.existsSync(path.join(primaryDir, STATE_FILE_NAME)) && fs.existsSync(path.join(fallbackDir, STATE_FILE_NAME))) {
       fs.cpSync(fallbackDir, primaryDir, { recursive: true });
-      // Rewrite logFile paths in migrated state.json so they point to the
-      // new persistent location instead of the old tmpdir.  Replace both
-      // raw paths and JSON-escaped paths (backslashes doubled on Windows).
-      const migratedStateFile = path.join(primaryDir, STATE_FILE_NAME);
-      try {
-        let updated = fs.readFileSync(migratedStateFile, "utf8");
-        const raw = updated;
-        updated = updated.replaceAll(fallbackDir, primaryDir);
-        const escapedFallback = fallbackDir.replaceAll("\\", "\\\\");
-        const escapedPrimary = primaryDir.replaceAll("\\", "\\\\");
-        if (escapedFallback !== fallbackDir) {
-          updated = updated.replaceAll(escapedFallback, escapedPrimary);
+      // Rewrite paths in all migrated JSON files (state.json + jobs/*.json)
+      // so logFile and other references point to the persistent location.
+      // Replace both raw paths and JSON-escaped paths (backslashes on Windows).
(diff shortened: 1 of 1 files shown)


tmchow:
> 🦀 Assigned! I will submit a PR for this issue. Looking at the fix now.

@D2758695161 you're clearly an automated bot who's logic is gone wrong 😑.  Don't propose PRs to issues fixed by the PRs you're reading.

PR #126, opened 2026-04-03
fix: gracefully handle unsupported thread/name/set on older Codex CLI

### LINKED ISSUE #119: companion script fails with 'unknown variant thread/name/set' on Codex CLI 0.118.0
## Description

The `codex-companion.mjs` script fails when used with Codex CLI v0.118.0 (installed via `brew install --cask codex`).

The companion script sends a `thread/name/set` JSON-RPC method that is not recognized by the CLI.

## Error

```
[codex] Starting Codex task thread.
Invalid request: unknown variant `thread/name/set`, expected one of `initialize`, `thread/start`, `thread/resume`, ...
```

## Environment

- **Codex CLI**: v0.118.0 (brew cask)
- **codex-plugin-cc**: v1.0.2 (commit 8e403f9)
- **Node.js**: v25.5.0
- **OS**: macOS (Apple Silicon)

## Steps to Reproduce

1. Install Codex CLI v0.118.0 via `brew install --cask codex`
2. Install codex-plugin-cc via `/plugin marketplace add openai/codex-plugin-cc`
3. Run `/codex:rescue` or any task command
4. Observe the `thread/name/set` error

## Expected Behavior

The companion script should use the supported API methods for the current Codex CLI version.

## Workaround

Using the Codex CLI directly from the terminal works fine. Only the plugin integration is affected.

tmchow:
## Summary

`startThread()` in `codex.mjs` calls `thread/name/set` unconditionally, which throws on Codex CLI v0.118.0 because that version doesn't support the method. This wraps the call in a try/catch so thread creation succeeds even when naming is unsupported.

## Changes

One file: `plugins/codex/scripts/lib/codex.mjs`, function `startThread()` (line 639-645).

Thread naming is cosmetic (used for job log labels) and should not prevent thread creation. The try/catch allows the plugin to work with both older and newer Codex CLI versions.

## Testing

- `npm test`: same results as main (pre-existing failures in commands.test.mjs, unrelated)
- `npm run build`: passes

Fixes #119

### DIFF
--- a/plugins/codex/scripts/lib/codex.mjs
+++ b/plugins/codex/scripts/lib/codex.mjs
@@ -641,8 +641,13 @@ async function startThread(client, cwd, options = {}) {
   if (options.threadName) {
     try {
       await client.request("thread/name/set", { threadId, name: options.threadName });
-    } catch {
-      // thread/name/set not supported on this Codex CLI version
+    } catch (err) {
+      // Only suppress "unknown variant/method" errors from older CLI versions
+      // that don't support thread/name/set. Rethrow auth, network, or server errors.
+      const msg = String(err?.message ?? err ?? "");
+      if (!msg.includes("unknown variant") && !msg.includes("unknown method")) {
+        throw err;
+      }
     }
   }
   return response;


D2758695161:
🦀 Assigned! I will submit a PR for this issue. Looking at the fix now.

PR #133, opened 2026-04-03
test: isolate saveState pruning test from shared git-repo state dir

JiayuuWang:
## Summary

- Fixes a flaky/failing test: `saveState prunes dropped job artifacts when indexed jobs exceed the cap`
- The test was failing because `resolveWorkspaceRoot()` calls `ensureGitRepository()`, which climbs to the nearest git root. When tests run inside this plugin's own repository, every fresh `makeTempDir()` workspace resolves to the **same** repo root — collapsing all tests onto one shared state directory.
- Artifact files written by `runtime.test.mjs` (e.g. `review-live.log`, `task-live.json`) bleed into the pruning test's jobs directory, causing the directory-contents `deepEqual` assertion to fail with unexpected extra files.

**Fix:** Set `CLAUDE_PLUGIN_DATA` to the isolated temp workspace for the duration of the test (with env-var cleanup in a `finally` block), so `resolveStateDir` roots state under the per-test temp dir. This is the same pattern already used by the `resolveStateDir uses CLAUDE_PLUGIN_DATA when it is provided` test in the same file.

## Test plan

- [ ] `node --test tests/state.test.mjs` passes (all 3 tests green)
- [ ] `node --test tests/state.test.mjs tests/render.test.mjs tests/git.test.mjs` passes
- [ ] No other test files are affected

### DIFF
--- a/tests/state.test.mjs
+++ b/tests/state.test.mjs
@@ -41,65 +41,80 @@ test("resolveStateDir uses CLAUDE_PLUGIN_DATA when it is provided", () => {
 });
 
 test("saveState prunes dropped job artifacts when indexed jobs exceed the cap", () => {
+  // Use CLAUDE_PLUGIN_DATA to ensure the state directory is rooted inside the
+  // isolated temp workspace rather than in the shared git-repo-derived path.
+  // Without this, resolveWorkspaceRoot() climbs to the git repo root and all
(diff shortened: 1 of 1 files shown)


dkundel-openai:
Duplicate of #456.

PR #162, opened 2026-04-06
fix: prevent Windows ENOENT for codex availability checks

### LINKED ISSUE #161: Windows: codex-companion.mjs fails with ENOENT because spawnSync can't resolve 'codex' in PATH
**Describe the bug**
On Windows, running `/codex:review` via Claude Code fails immediately with:

```
Codex CLI is not installed or is missing required runtime support. Install it with \`npm install -g @openai/codex\`, then rerun \`/codex:setup\`.
```

The Codex CLI IS installed and `codex --version` works in a terminal, but the companion script's `runCommand()` uses `spawnSync` without `shell: true`, which cannot resolve `codex` on Windows because Node's `spawnSync` without shell does not expand PATH on Windows.

**To Reproduce**
1. Have Codex CLI installed (`npm install -g @openai/codex`)
2. Run `codex --version` in PowerShell — confirms it works
3. Run `/codex:review` in Claude Code on Windows
4. Observe the "not installed" error

**Root Cause**
In `scripts/lib/process.mjs`, the `runCommand` function uses `spawnSync` without `shell: true`:

```js
export function runCommand(command, args = [], options = {}) {
  const result = spawnSync(command, args, {
    cwd: options.cwd,
    env: options.env,
    encoding: "utf8",
    input: options.input,
    stdio: options.stdio ?? "pipe"
  });
```

On Windows, `spawnSync` without `shell: true` cannot resolve commands in the user PATH (like `codex` installed via npm global). The shell is required to expand PATH.

**Evidence**
```js
// Without shell — ENOENT on Windows
spawnSync('codex', ['--version'], { stdio: 'pipe' })
// { error: { code: 'ENOENT', syscall: 'spawnSync codex', path: 'codex' } }

// With shell — works fine
spawnSync('codex', ['--version'], { shell: true, stdio: 'pipe' })
// { status: 0, stdout: 'codex-cli 0.118.0' }
```

**Suggested Fix**
In `scripts/lib/process.mjs`, add `shell: true` for Windows:

```js
export function runCommand(command, args = [], options = {}) {
  const result = spawnSync(command, args, {
    cwd: options.cwd,
    env: options.env,
    encoding: "utf8",
    input: options.input,
    stdio: options.stdio ?? "pipe",
    shell: process.platform === 'win32'  // <-- fix
  });
```

**Environment**
- OS: Windows 11
- Node.js: v24.12.0
- Codex CLI: 0.118.0
- Plugin: openai-codex via Claude Code plugin

**Reported by:** Thepushpendra68

Abdooo2235:
## Summary
Fixes a Windows command resolution failure where Codex availability checks could fail with `ENOENT` even when `codex` was installed globally and available in PATH.

On Windows, command resolution can require shell execution for PATH-resolved shims. This change keeps Windows shell behavior and adds explicit regression coverage for it.

Closes #161.

## Changes
- Updated command runner to make platform/spawn implementation injectable for tests while preserving behavior:
  - `shell` now resolves from `options.shell ?? platform === "win32"`
  - added `platform` and `spawnSyncImpl` test hooks
- Added regression tests to ensure:
  - Windows uses `shell: true` for PATH-resolved commands
  - non-Windows does not force shell mode

## Validation
- `npm test` (pass)
- includes new tests in `tests/process.test.mjs` for Windows/non-Windows shell behavior

## Risk
Low. Change is scoped to process invocation wiring and covered by targeted and full-suite tests.

### DIFF
--- a/plugins/codex/scripts/lib/process.mjs
+++ b/plugins/codex/scripts/lib/process.mjs
@@ -2,13 +2,16 @@ import { spawnSync } from "node:child_process";
 import process from "node:process";
 
 export function runCommand(command, args = [], options = {}) {
-  const result = spawnSync(command, args, {
+  const platform = options.platform ?? process.platform;
+  const spawnSyncImpl = options.spawnSyncImpl ?? spawnSync;
+
+  const result = spawnSyncImpl(command, args, {
     cwd: options.cwd,
     env: options.env,
     encoding: "utf8",
     input: options.input,
     stdio: options.stdio ?? "pipe",
-    shell: process.platform === "win32",
+    shell: options.shell ?? platform === "win32",
     windowsHide: true
   });
 
--- a/tests/process.test.mjs
+++ b/tests/process.test.mjs
@@ -1,7 +1,52 @@
 import test from "node:test";
 import assert from "node:assert/strict";
 
-import { terminateProcessTree } from "../plugins/codex/scripts/lib/process.mjs";
+import { runCommand, terminateProcessTree } from "../plugins/codex/scripts/lib/process.mjs";
+
+test("runCommand enables shell on Windows for PATH-resolved commands", () => {
+  let captured = null;
+
+  const result = runCommand("codex", ["--version"], {
+    platform: "win32",
+    spawnSyncImpl(command, args, options) {
+      captured = { command, args, options };
+      return {
+        status: 0,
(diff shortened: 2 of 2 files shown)


dkundel-openai:
Thanks this should have been fixed in #55

PR #176, opened 2026-04-07
Fix stuck running jobs by detecting dead review/task PIDs

Abdooo2235:
## Summary
Fixes stuck Codex jobs that stayed in `queued`/`running` after the tracked process died unexpectedly.

Closes #164.

## Problem
When a review/task process exited unexpectedly, job state could remain `running` indefinitely.
That caused `status --wait` to keep polling until timeout with no actionable failure reason.

## What Changed
- Added PID liveness detection for tracked jobs.
- Added reconciliation that converts stale active jobs with dead PIDs to `failed`.
- Persisted reconciliation to both the state index and per-job file.
- Stored a clear failure message and completion timestamp when dead PID is detected.
- Surfaced `errorMessage` in status output for easier diagnosis.
- Included wait-timeout context in single-job status rendering.
- Added a regression test for dead-PID plus `status --wait` behavior.

## Behavior After Fix
- Dead worker/reviewer PID is detected promptly.
- Job transitions to `failed` instead of remaining `running`.
- `status --wait` returns quickly with failure state rather than timing out.
- Diagnostic reason is visible in status output and persisted job artifacts.

## Test Plan
- Ran full suite: `npm test`
- Result: 67 passed, 0 failed.
- Includes regression coverage for dead-PID reconciliation in `status --wait`.

Issue #164, opened 2026-04-07
Review jobs stuck in 'running' after process death — no dead-PID detection


peterdrier:
## Problem

When a Codex review process crashes or times out, the companion job state file stays in `"status": "running"` indefinitely. The `--wait` polling loop continues checking for completion against a dead PID, eventually timing out with no useful error message.

## Reproduction

1. Launch a large review: `/codex:review --wait --base upstream/main` (148 files, ~19.5k lines)
2. The review session starts, logs "Reviewer started", then the process dies silently
3. Job state file still says `"status": "running"`
4. The `--wait` poll loop spins for 10+ minutes before the caller gives up
5. No error is recorded — the job just looks "stuck"

## Evidence from job log

```
[2026-04-07T02:21:02.393Z] Starting Codex Review.
[2026-04-07T02:21:03.165Z] Starting Codex review thread.
[2026-04-07T02:21:04.768Z] Thread ready (019d65be-512f-7cc0-85f2-a5ddeb9bce70).
[2026-04-07T02:21:04.825Z] Reviewer started: changes against 'upstream/main'
(nothing further — process PID 2568362 died)
```

## Expected behavior

1. The broker should detect when the child Codex process exits unexpectedly and mark the job as `"failed"` with an error message
2. The `--wait` poll should check if the PID is still alive and fail fast if it isn't
3. On timeout, the error message should indicate the process died rather than just "still running"

## Impact

- Wastes 10+ minutes of wall time before the caller realizes the review failed
- No diagnostic information about *why* it failed (OOM? API timeout? context overflow?)
- Makes large reviews unreliable — the only recovery is manual cancel + retry

Brooooooklyn:
Same here

dkundel-openai:
Duplicate of #222.

PR #184, opened 2026-04-08
fix(runtime): prevent tracked jobs hanging forever on broker disconnect

jonny981:
## Summary

A Codex task can become stuck in `status: running` indefinitely with no way to recover short of manually wiping the job-store and broker session. The user-visible symptom is that `/codex:status` keeps reporting a job as running for many minutes after the underlying codex CLI worker has died, and the next `/codex:cancel` reports `ECONNREFUSED` on the broker socket. This PR fixes four independent root causes that combine to produce the wedge.

Closes #183. Refs #176, #164.

## Reproduction

Run any moderately heavy task that exercises the broker for ~3-5 minutes. In my case it was a Codex review of a ~4,500-line markdown corpus. After the codex CLI processed several `nl`/`jq`/`rg` shell commands and started reasoning about the result, the worker fell silent. The job-store kept reporting `phase: running` for 9+ minutes with no log progress. `/codex:cancel` returned `ECONNREFUSED /var/folders/.../cxc-XXXXXX/broker.sock`. Both the companion PID and the broker PID were dead by the time I cancelled.

## Root causes

1. **`captureTurn` hangs forever on disconnection.** `lib/codex.mjs` awaits `state.completion` with no race against the client's `exitPromise`. When the underlying codex app-server (or the broker socket) dies mid-turn, no terminal event is delivered, so the await never resolves. Pending RPC requests are rejected by `handleExit` but the turn-completion promise is unrelated to those, so it sits forever.

2. **`runTrackedJob` has no timeout fallback.** `lib/tracked-jobs.mjs` awaits the runner without a timeout. When `captureTurn` hangs, the entire companion process hangs and the per-job state file is never transitioned to a terminal status.

3. **The broker zombifies on app-server death.** `app-server-broker.mjs` never listens for the underlying app-server client's exit. When the codex CLI process dies, the broker keeps running with a dead client, accepting new connections but unable to serve them. The next companion sees a stale `broker.json` that points at a half-dead broker.

4. **Status reads do not probe PID liveness.** `buildStatusSnapshot` and `buildSingleJobSnapshot` never check whether tracked PIDs are still alive. Once a job is wedged, every `/codex:status` query reports it as running.

## Fixes

- **`captureTurn` races against client exit (lib/codex.mjs).** A new `exitWatch` promise rejects `state.completion` with the recorded exit error when the client closes before a terminal event arrives. Progress emits a `failed` notification so the user sees what happened.

- **`runTrackedJob` enforces a hard timeout (lib/tracked-jobs.mjs).** Default 30 minutes, override via `CODEX_COMPANION_JOB_TIMEOUT_MS` env var or the `timeoutMs` option. On timeout the job transitions to `failed` with a clear error message instead of leaving the companion stuck.

- **Broker subscribes to appClient exit (app-server-broker.mjs).** When the underlying codex CLI exits unexpectedly the broker logs the reason, fans out a `notifications/broker/shuttingDown` event to any connected socket, tears down the unix socket and pid file, then exits with status 1. The next companion will detect a dead endpoint via `ensureBrokerSession` and respawn cleanly.

- **`isProcessAlive` helper (lib/process.mjs).** Uses `kill(pid, 0)` to probe pid liveness without affecting the process.

- **`markDeadPidJobFailed` + `reconcileIfDead` (lib/job-control.mjs).** Reconcile any active job whose tracked PID is gone. Reconciliation runs on every status read path now, not only the `--wait` polling loop, so a single `/codex:status` call surfaces dead workers immediately. This is broader than #176, which only reconciled inside `waitForSingleJobSnapshot`.

## Relationship to existing PRs/issues

- **#183** (runTrackedJob/captureTurn hang in finalizing) — fully addressed by the captureTurn race + the hard timeout.
- **#176** (Fix stuck running jobs by detecting dead review/task PIDs) — this PR includes equivalent dead-PID detection but also reconciles on every status read, not only `status --wait`. If #176 lands first, the overlap is small (the helper functions and a couple of import lines).
- **#164** (Review jobs stuck in 'running' after process death) — also addressed.

## Test plan

- New unit tests (11 cases, all pass):
  - `tests/process.test.mjs` — 3 new cases for `isProcessAlive` (invalid input, current pid, exited child).
  - `tests/dead-pid-reconcile.test.mjs` — 6 new cases covering `markDeadPidJobFailed` guards and reconcile-on-status for both single-job and full snapshots.
  - `tests/tracked-jobs-timeout.test.mjs` — 2 new cases proving the hard timeout transitions a hung job to `failed` and that fast runners still complete normally.
- Full suite: `node --test tests/*.test.mjs` → 91 pass, 5 fail. The 5 failures (`collectReviewContext skips untracked directories`, three `runtime.test.mjs` snapshot tests, and `resolveStateDir uses a temp-backed per-workspace directory`) are **pre-existing on `main` and unrelated to this patch** — I confirmed by running the same suite against pristine `main`.

## Test plan (manual)

- [x] Reproduced the wedge with a long-running Codex task on macOS.
- [x] Verified `/codex:status` now reconciles a dead job to `failed` on a single call.
- [x] Verified the new hard timeout fires and writes a `failed` status.
- [ ] Smoke-test on Windows (I do not have a Windows host; the changes use only `process.kill(pid, 0)` on Unix paths and the platform branch in `terminateProcessTree` is unchanged).

### DIFF
--- a/plugins/codex/scripts/app-server-broker.mjs
+++ b/plugins/codex/scripts/app-server-broker.mjs
@@ -70,6 +70,7 @@ async function main() {
   let activeStreamSocket = null;
   let activeStreamThreadIds = null;
   const sockets = new Set();
+  let shuttingDown = false;
 
   function clearSocketOwnership(socket) {
     if (activeRequestSocket === socket) {
@@ -100,7 +101,25 @@ async function main() {
   }
 
   async function shutdown(server) {
+    if (shuttingDown) {
+      return;
+    }
+    shuttingDown = true;
+    // Notify any client whose request is still in flight that the broker is
+    // going down. Without this they would hang on a half-open socket waiting
+    // for a response that will never arrive.
+    const fanoutMessage = {
+      method: "notifications/broker/shuttingDown",
+      params: {
+        reason: appClient.exitError ? String(appClient.exitError.message ?? appClient.exitError) : "broker shutdown"
+      }
+    };
     for (const socket of sockets) {
+      try {
+        send(socket, fanoutMessage);
+      } catch {
+        // Best-effort fanout. Continue tearing down.
+      }
       socket.end();
     }
     await appClient.close().catch(() => {});
@@ -114,6 +133,21 @@ async function main() {
   }
 
   appClient.setNotificationHandler(routeNotification);
+
+  // If the underlying codex CLI app-server dies (crashes, OOM, exits), tear
+  // down the broker rather than zombify with a dead client. The next companion
+  // process will detect a dead endpoint via ensureBrokerSession and respawn.
+  // See: openai/codex-plugin-cc#183.
+  appClient.exitPromise.then(() => {
+    if (shuttingDown) {
+      return;
+    }
+    const detail = appClient.exitError instanceof Error ? appClient.exitError.message : String(appClient.exitError ?? "unknown");
+    process.stderr.write(`[broker] underlying codex app-server exited: ${detail}\n`);
+    shutdown(server)
+      .catch(() => {})
+      .finally(() => process.exit(1));
+  });
 
   const server = net.createServer((socket) => {
     sockets.add(socket);
--- a/plugins/codex/scripts/lib/codex.mjs
+++ b/plugins/codex/scripts/lib/codex.mjs
@@ -597,7 +597,37 @@ async function captureTurn(client, threadId, startRequest, options = {}) {
       completeTurn(state, response.turn);
     }
 
-    return await state.completion;
+    // Race the turn completion against the client's exit promise so that if
+    // the underlying codex app-server (or broker socket) dies mid-turn, we
+    // reject instead of hanging forever waiting for a terminal event that
+    // will never arrive. See: openai/codex-plugin-cc#183.
+    const exitWatch = client.exitPromise.then(() => {
+      if (state.completed) {
+        return;
+      }
+      const reason =
+        client.exitError instanceof Error
+          ? client.exitError
+          : new Error(
+              client.exitError
+                ? String(client.exitError)
+                : "codex app-server connection closed before the turn completed."
+            );
+      emitProgress(
+        state.onProgress,
+        `Codex app-server disconnected mid-turn: ${reason.message}`,
+        "failed"
+      );
+      state.rejectCompletion(reason);
+    });
+
+    try {
+      return await state.completion;
+    } finally {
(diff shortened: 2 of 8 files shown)


steve-ham:
Hit this exactly as described in #183 today on plugin `1.0.3` / `codex-cli 0.121.0` / macOS 15.4 arm64, with one additional angle worth flagging for the fix:

**The hang reproduces deterministically when a `/codex:review` makes any MCP tool call to the `xcode` server (`xcrun mcpbridge`).** Same request to the same bridge responds in ~2.5s from a direct stdio probe and from `codex exec`, but the review path's call to `xcode/XcodeListWindows` sits forever — no progress events, no terminal event, `captureTurn` never resolves. Companion log ends with `Calling xcode/XcodeListWindows.` and then silence for 29+ minutes until manual cancel.

Two data points that might be useful for #184:

1. **`tool_timeout_sec = 30` on the MCP never fires in this path.** Verified it *does* fire cleanly on an equivalent `codex exec -c 'mcp_servers.hang={command="/tmp/hang.sh", tool_timeout_sec=5}' …` run (returns `tool call failed for hang/hang_forever … timed out awaiting tools/call after 5s` at exactly 5s). So the timeout machinery exists, it just isn't reached when the wedge is at the job/broker level.

2. **Killing the MCP subprocess mid-hang doesn't recover.** `kill -TERM` on both `xcrun mcpbridge` children of the codex app-server leaves the review stuck on the same call — no error, no retry, no timeout. Supports the PR's root cause #1 (captureTurn with no race against exit) and #3 (broker not listening for appClient exit).

After cancel, `/codex:setup` reported `ECONNREFUSED /var/folders/.../cxc-<id>/broker.sock` — the broker process had died but `broker.json` still pointed at it, and `ensureBrokerSession`'s socket-connect-only probe passed on the bound-but-dead socket. Manually removing `broker.json` + the `cxc-<id>` tmp dir unblocked the next review. Your `isProcessAlive` + broker-death propagation would catch this automatically.

Workaround for anyone else landing here: open the project in Xcode, disable the `build-ios-apps@openai-curated` plugin (its `xcodebuildmcp/build_run_sim` separately wedges Xcode's MCP tool handler after the sim build completes), and comment out `[mcp_servers.xcode]` in `~/.codex/config.toml` until #184 lands. `codex exec` with a review-style prompt works fine with Xcode MCP enabled.

Thanks for the thorough PR — the four root causes you identified map 1:1 to what I observed.

PR #190, opened 2026-04-09
[codex] sanitize codex child process env

seungpyoson:
## Summary

- add a shared `sanitizeChildEnv` helper in `plugins/codex/scripts/lib/process.mjs`
- route plugin child-process launches through that sanitizer in the direct app-server, broker, detached task worker, and stop-review gate paths
- add regression coverage for the routing-env contract and align the deterministic command-doc test with the current quoted `"$ARGUMENTS"` command form

## Why

This was surfaced by `seungpyoson/claude-config#525`.

The Codex marketplace plugin was forwarding ambient Rust verification routing variables across plugin process boundaries. The plugin did not expose a shared sanitizer, and downstream boundary tests were correctly identifying that drift.

## Impact

- strips routing-intent and cargo target override env vars from plugin child environments
- makes the process-boundary contract explicit in plugin code and tests
- keeps existing launch behavior unchanged apart from environment scrubbing

## Validation

- `node --test tests/process.test.mjs`
- `npm test`

### DIFF
--- a/plugins/codex/scripts/codex-companion.mjs
+++ b/plugins/codex/scripts/codex-companion.mjs
@@ -22,7 +22,7 @@ import {
   } from "./lib/codex.mjs";
 import { readStdinIfPiped } from "./lib/fs.mjs";
 import { collectReviewContext, ensureGitRepository, resolveReviewTarget } from "./lib/git.mjs";
-import { binaryAvailable, terminateProcessTree } from "./lib/process.mjs";
+import { binaryAvailable, sanitizeChildEnv, terminateProcessTree } from "./lib/process.mjs";
 import { loadPromptTemplate, interpolateTemplate } from "./lib/prompts.mjs";
 import {
   generateJobId,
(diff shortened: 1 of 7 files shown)


seungpyoson:
Requesting adversarial review.

Focus areas:
- child-process env sanitization boundaries
- whether `sanitizeChildEnv` is applied at every relevant launch surface
- regressions from scrubbing `RUST_VERIFICATION_*`, `BOLT_RUST_VERIFICATION_ROOT`, and `CARGO_*TARGET_DIR`
- whether the new tests actually prove the contract
- whether aligning `tests/commands.test.mjs` with the current quoted `"$ARGUMENTS"` command form is correct and minimal

Validation already run on this branch:
- `node --test tests/process.test.mjs`
- `npm test`

PR: https://github.com/openai/codex-plugin-cc/pull/190

seungpyoson:
Follow-up pushed in response to review:

- harden `sanitizeChildEnv(...)` to scrub keys case-insensitively so mixed-case Windows env keys cannot bypass the filter
- treat `null` like inherited env input instead of producing an empty env object
- extend tests to cover:
  - `CARGO_TARGET_DIR`
  - preserved `CARGO_HOME`
  - case-insensitive key scrubbing
  - inherited-env sanitization when `runCommand(...)` is called without an explicit `env`

Latest validation on branch:
- `node --test tests/process.test.mjs`
- `npm test`

Both are passing on the updated branch.

PR #225, opened 2026-04-14
[codex] Reconcile stale running jobs when pid exits

LawrenceChiu95:
## Summary
- auto-reconcile stale `running` jobs when the tracked pid no longer exists
- persist the reconciliation to both state index (`state.json`) and per-job file (`jobs/<id>.json`)
- append a log line so users can see why a job switched to `failed`

## Root Cause
`/codex:status` trusted persisted state for `running` jobs without validating whether the worker process was still alive. If a worker died unexpectedly, the job could remain `running` indefinitely and appear stuck.

## Changes
- `plugins/codex/scripts/lib/state.mjs`
  - add pid liveness check (`process.kill(pid, 0)`)
  - add reconciliation path in `listJobs()` for stale `running` jobs with dead pids
  - mark stale jobs as `failed` with `errorMessage`, `completedAt`, and `pid: null`
  - mirror updates into per-job json and append a log annotation
- `tests/state.test.mjs`
  - add regression test: `listJobs auto-reconciles stale running jobs when pid is no longer alive`

## Validation
- `node --test tests/state.test.mjs`
- `node --test tests/runtime.test.mjs --test-name-pattern "status shows phases|status preserves adversarial review kind labels|status --wait times out cleanly|cancel without a job id ignores active jobs from other Claude sessions|cancel with a job id can still target an active job from another Claude session|stop hook logs running tasks to stderr without blocking when the review gate is disabled"`

### DIFF
--- a/plugins/codex/scripts/lib/state.mjs
+++ b/plugins/codex/scripts/lib/state.mjs
@@ -2,6 +2,7 @@ import { createHash } from "node:crypto";
 import fs from "node:fs";
 import os from "node:os";
 import path from "node:path";
+import process from "node:process";
 
 import { resolveWorkspaceRoot } from "./workspace.mjs";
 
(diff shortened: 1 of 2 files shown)


dkundel-openai:
Duplicate of #425.

PR #255, opened 2026-04-21
fix: namespace app-server client info

### LINKED ISSUE #199: clientInfo.name should use codex-namespaced identifier instead of host application name
## Problem

The plugin currently sets `clientInfo.name` to `"Claude Code"` in `DEFAULT_CLIENT_INFO` ([app-server.mjs, line 28](https://github.com/openai/codex-plugin-cc/blob/main/plugins/codex/scripts/lib/app-server.mjs#L28)). This causes the Codex app-server to present `Claude Code/x.x.x` as the user-agent to upstream services, which creates several issues:

### 1. User-agent collision with the host application

Claude Code (Anthropic's CLI) already uses `Claude Code/x.x.x` as its own user-agent for Anthropic API requests. When the Codex plugin running *inside* Claude Code also presents the same UA string, upstream services cannot distinguish between:

- Native Claude Code requests (Anthropic Messages API)
- Codex requests routed through Claude Code (OpenAI Responses API)

This breaks any service that uses UA-based routing, rate limiting, or analytics.

### 2. Naming convention violation

Every other Codex surface uses a `codex_*` or `Codex *` prefix:

| Surface | `clientInfo.name` |
|---------|------------------|
| CLI (Rust) | `codex_cli_rs` |
| TUI | `codex-tui` |
| VS Code extension | `codex_vscode` |
| Desktop app | `Codex Desktop` |
| **Claude Code plugin** | **`Claude Code`** ← only outlier |

The Claude Code plugin is the only surface that uses the host application's name instead of following the established `codex_` + host platform pattern.

### 3. Compliance log misattribution

The [app-server documentation](https://developers.openai.com/codex/app-server) states that `clientInfo.name` is used to identify the client for the **OpenAI Compliance Logs Platform**. Using `"Claude Code"` incorrectly attributes Codex plugin usage to a different vendor's product in compliance records.

## Proposed fix

Rename `clientInfo.name` from `"Claude Code"` to `"codex_claude_code"`, consistent with the `codex_vscode` naming pattern:

```js
const DEFAULT_CLIENT_INFO = {
  title: "Codex Plugin",
  name: "codex_claude_code",
  version: PLUGIN_MANIFEST.version ?? "0.0.0"
};
```

### LINKED ISSUE #276: app-server clientInfo.name "Claude Code" causes 400 invalid_request_error for gpt-5.5
## Summary

The Codex Claude Code plugin's app-server identifies itself as `clientInfo.name = "Claude Code"` during the `initialize` handshake. The OpenAI Codex backend appears to gate access to newer models (currently `gpt-5.5`) by client name, and rejects requests from this identifier with a misleading version-related error:

```
{"type":"error","status":400,"error":{"type":"invalid_request_error",
 "message":"The 'gpt-5.5' model requires a newer version of Codex. Please upgrade to the latest app or CLI and try again."}}
```

The codex CLI itself (`codex exec -m gpt-5.5`) works fine on the same machine — it identifies as `codex_exec`. So this is not a CLI version issue; the user's CLI is current.

## Reproduction

- codex-cli: `0.126.0-alpha.10`
- codex plugin: `1.0.4`
- auth: ChatGPT Pro

```bash
# Works:
codex exec -m gpt-5.5 "say hi"

# Fails with the error above:
node ~/.claude/plugins/cache/openai-codex/codex/1.0.4/scripts/codex-companion.mjs task --model gpt-5.5 "say hi"
```

\`/codex:rescue\`, \`/codex:review\`, and \`/codex:adversarial-review\` all fail the same way when invoked with \`gpt-5.5\`, since they share the app-server path.

## Root cause

[\`plugins/codex/scripts/lib/app-server.mjs\`](https://github.com/openai/codex-plugin-cc/blob/main/plugins/codex/scripts/lib/app-server.mjs) hardcodes:

\`\`\`js
const DEFAULT_CLIENT_INFO = {
  title: "Codex Plugin",
  name: "Claude Code",
  version: PLUGIN_MANIFEST.version ?? "0.0.0"
};
\`\`\`

The codex tracing logs confirm requests go out as \`app_server.client_name="Claude Code"\` from the plugin vs \`app_server.client_name="codex_exec"\` from \`codex exec\`. Patching \`name\` to \`"codex_cli"\` locally makes the same \`gpt-5.5\` request succeed end-to-end (turn/completed → assistant reply).

## Suggested fix

Change \`DEFAULT_CLIENT_INFO.name\` to a \`codex_*\`-prefixed identifier (e.g. \`"codex_claude_code_plugin"\` or \`"codex_cli"\`) so the plugin inherits the same model-gating allowlist as the CLI. Alternatively, OpenAI could whitelist \`"Claude Code"\` server-side.

Happy to send a PR if useful.

d3v07:
Closes #199
Fixes #276

## Summary
- rename the default app-server client name to the namespaced plugin identifier
- capture the app-server initialize payload in the fake fixture
- add a regression test that verifies the initialize payload uses the namespaced client name

## Verification
- `npm test`
- `npm run build`

### DIFF
--- a/plugins/codex/scripts/lib/app-server.mjs
+++ b/plugins/codex/scripts/lib/app-server.mjs
@@ -25,7 +25,7 @@ export const BROKER_BUSY_RPC_CODE = -32001;
 /** @type {ClientInfo} */
 const DEFAULT_CLIENT_INFO = {
   title: "Codex Plugin",
-  name: "Claude Code",
+  name: "codex_claude_code",
   version: PLUGIN_MANIFEST.version ?? "0.0.0"
 };
 
--- a/tests/fake-codex-fixture.mjs
+++ b/tests/fake-codex-fixture.mjs
@@ -18,7 +18,7 @@ const readline = require("node:readline");
 
 	function loadState() {
 	  if (!fs.existsSync(STATE_PATH)) {
-	    return { nextThreadId: 1, nextTurnId: 1, appServerStarts: 0, threads: [], capabilities: null, lastInterrupt: null };
+	    return { nextThreadId: 1, nextTurnId: 1, appServerStarts: 0, threads: [], capabilities: null, lastInitialize: null, lastInterrupt: null };
 	  }
 	  return JSON.parse(fs.readFileSync(STATE_PATH, "utf8"));
 	}
@@ -271,6 +271,7 @@ rl.on("line", (line) => {
     switch (message.method) {
       case "initialize":
         state.capabilities = message.params.capabilities || null;
+        state.lastInitialize = message.params || null;
         saveState(state);
         send({ id: message.id, result: { userAgent: "fake-codex-app-server" } });
         break;
--- a/tests/runtime.test.mjs
+++ b/tests/runtime.test.mjs
@@ -44,6 +44,22 @@ test("setup reports ready when fake codex is installed and authenticated", () =>
   assert.equal(payload.sessionRuntime.mode, "direct");
 });
 
+test("setup initializes the app server with a codex-namespaced client name", () => {
+  const binDir = makeTempDir();
+  const statePath = path.join(binDir, "fake-codex-state.json");
+  installFakeCodex(binDir);
+
+  const result = run("node", [SCRIPT, "setup", "--json"], {
+    cwd: ROOT,
+    env: buildEnv(binDir)
+  });
+
+  assert.equal(result.status, 0, result.stderr);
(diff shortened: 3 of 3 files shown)


disi-omni:
This also resolves #276 — same root cause with a concrete user-impact repro: requests for `gpt-5.5` from the plugin path fail with a misleading `400 invalid_request_error: "The 'gpt-5.5' model requires a newer version of Codex"`, while `codex exec -m gpt-5.5` succeeds on the same machine. The model gate is on `clientInfo.name`, not on CLI version.

PR #262, opened 2026-04-23
fix(broker-lifecycle): reject crashed brokers via PID-alive probe

Dimaris-nsk:
## Summary

`ensureBrokerSession()` trusts any existing broker whose endpoint passes a 150ms socket ping. When the broker's underlying `codex app-server` subprocess is in a bad state but the socket accept-loop still works, every downstream task disconnects mid-turn — yet on each call the helper reuses the same broker instead of spawning a fresh one.

Observed in the wild: a broker process that had been up 3 days 11 hours caused 100% task failure across many invocations. Deleting `broker.json` (forcing a fresh broker on the next task) immediately restored reliability.

## Fix

Add `isSessionStale()` with two complementary checks, consulted before trusting the socket ping:

1. **PID-alive probe** (`process.kill(pid, 0)`) — catches the acute case where the broker has crashed but the socket file lingers.
2. **Age-based rotation** — tracks `session.startedAt` (new field) and forces a fresh broker past a threshold. Default 6h, overridable via `CODEX_COMPANION_BROKER_MAX_AGE_HOURS` env var. Catches the slow-degradation case neither PID nor socket ping surface.

If `isSessionStale()` returns true, the existing session flows through the same teardown path already used for missing/unreachable brokers — a fresh broker spawns.

## Test plan

- [x] `node --check` on modified file
- [x] Verified behavioral fix: after delete-and-respawn the broker via this logic, previously-failing tasks complete cleanly (3 consecutive green tasks on the same workspace).
- [ ] Regression: fresh broker session (age 0, PID alive) still reused — straight read of the new code confirms; will also appear in CI if upstream has broker fixtures.

## Notes on the default

6 hours is intuition-based, not data-driven — we didn't have a reliable timeline on when the codex app-server starts degrading. Env override is deliberate so operators can tune without a code change. Happy to switch to a more data-backed default if maintainers have telemetry.

## Related PRs (independent)

- `fix(captureTurn): race state.completion with exitPromise to surface transport close` — surfaces the failure mode when broker proxy drops
- `fix(plugin): upgrade --write sandbox to danger-full-access so MCP tool calls work` — unrelated sandbox/MCP issue

### DIFF
--- a/plugins/codex/scripts/lib/broker-lifecycle.mjs
+++ b/plugins/codex/scripts/lib/broker-lifecycle.mjs
@@ -3,7 +3,7 @@ import net from "node:net";
 import os from "node:os";
 import path from "node:path";
 import process from "node:process";
-import { spawn } from "node:child_process";
+import { spawn, execFileSync } from "node:child_process";
 import { fileURLToPath } from "node:url";
 import { createBrokerEndpoint, parseBrokerEndpoint } from "./broker-endpoint.mjs";
 import { resolveStateDir } from "./state.mjs";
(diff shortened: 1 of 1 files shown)


Dimaris-nsk:
Thanks @chatgpt-codex-connector — both findings applied in `6ce0540`:

**1. Legacy sessions (P1 — missing startedAt):**
Added explicit branch — if `startedAt` is absent, `isSessionStale()` returns `true` so pre-upgrade `broker.json` entries go through one-time rotation on first upgraded run. After that rotation the freshly-spawned broker has `startedAt` set, and the normal age path takes over.

**2. Orphan broker leak on age rotation (P1 — killProcess default):**
Added `defaultKillProcess()` (SIGTERM, best-effort, ignores missing-pid errors) and wired it into the rotation path: `killProcess: options.killProcess ?? defaultKillProcess`. Now age-based rotation actually terminates the previous detached broker instead of only removing its socket/pidfile.

Caller opt-out still works — any caller that explicitly passes `killProcess: null` (or a custom fn) preserves current behavior.

No test harness exists for broker lifecycle in this repo that I found, so behavioral verification was a manual repro against the v1.0.2 plugin cache (spawn broker → age past threshold → next ensureBrokerSession → observe PID terminates via SIGTERM + no orphaned `codex app-server` child in `ps`).

Dimaris-nsk:
Good catch — fixed in the same amended commit.

Added `verifyBrokerPid()` which cross-checks `session.pid` against the on-disk pid-file the broker wrote at spawn time (`app-server-broker.mjs` `writePidFile`). Rotation only passes `defaultKillProcess` to `teardownBrokerSession` when:

- `session.pidFile` path is set (we know where to look)
- The file still exists on disk
- The file's contents parse to the same PID as `session.pid`

If any check fails (file removed, different PID, unreadable) we fall through to `killProcess: null` — same as trunk before this PR — so a potentially-recycled PID is never signaled. Socket/pidfile/sessionDir cleanup still runs.

Invariant: the broker writes its pid-file before becoming discoverable, and on normal exit the file is removed by `teardownBrokerSession`. An abrupt exit leaves a stale pidfile, but then the `isPidAlive` gate in `isSessionStale` catches it first (returns stale → different branch that doesn't signal).

Dimaris-nsk:
Fair — pid-file alone is not enough because `app-server-broker.mjs` only removes it in the clean `shutdown()` path. An ungraceful crash leaves a stale pid-file that a recycled PID can match.

Strengthened in amended commit: now `verifyBrokerPid()` also inspects `ps -p <pid> -o command=` and requires the command line to contain `app-server-broker.mjs` before considering SIGTERM safe. Conditions for passing:

1. `session.pid` is a finite number
2. `session.pidFile` set
3. Platform is POSIX (Windows skipped — `ps` unavailable → returns false → fallthrough to `null` killProcess, matches trunk)
4. pid-file exists and contents parse to `session.pid`
5. `ps` lookup succeeds within 1s
6. `ps` output contains `app-server-broker.mjs`

Any failure → return `false` → rotation uses `killProcess: null` → no signal, only socket/pidfile cleanup. Safe default unchanged.

This closes the stale-pid-file + PID-recycle hole. If maintainers prefer a different identity source (e.g. read `/proc/<pid>/comm` on Linux, or require an explicit broker-side heartbeat file), happy to switch — let me know.

Dimaris-nsk:
Added Windows branch so age-based rotation doesn't silently leak brokers there.

New Windows path uses `tasklist /FI "PID eq <pid>" /FO CSV /NH`:
1. pid-file exists and contents match `session.pid`
2. `tasklist` returns a row (PID is live)
3. The row's image name is `node.exe`

Then the existing `defaultKillProcess` (`process.kill(pid, "SIGTERM")`) on Windows dispatches to `TerminateProcess` via libuv — same mechanism the plugin already uses elsewhere for subprocess cleanup.

This is **weaker** than the POSIX `ps`-based check (`tasklist` doesn't expose full command line, so we can't verify the `app-server-broker.mjs` arg string — just that PID is alive AND is `node.exe`). But it closes the common case the bot raised (recycled PID ends up on some foreign process) for the vast majority of rotations: node.exe + matching pid-file content is reasonably specific.

If a stronger Windows identity check is required, the repo would need a new per-broker token file (e.g. a random UUID broker writes on spawn that survives crashes unchanged). I left that out of this PR to keep scope tight; happy to follow up with a separate PR if maintainers want that level.

Fallback on any `tasklist` failure → return false → `killProcess: null` → trunk behavior (no signal), so Windows users can't be worse off than before.

Dimaris-nsk:
Dropped both problematic parts of the previous revision.

**Age-based rotation removed.** You were right — rotating a healthy-but-aged broker while a concurrent client is mid-turn interrupts that turn because `app-server-broker.mjs` exits on SIGTERM. Proper fix would need a live health probe (RPC round-trip with timeout) or a graceful drain protocol, neither of which belongs in this narrow PR. Filed as a follow-up for a separate change.

**Windows kill attempt removed.** `tasklist`'s image-name match alone (`node.exe`) is too weak to rule out recycled-PID foreign processes — you're right that without full command line we can't prove ownership. On Windows `verifyBrokerPid()` now returns `false` early; the rotation path there cleans socket/pidfile only, no SIGTERM. The detached old broker eventually exits since new clients will reach the new broker via its fresh pipe path.

**What remains in this PR (the narrow, defensible core):**

1. `isPidAlive()` — catches the crashed-broker case where the socket file lingers but the process is gone. Without this, `isBrokerEndpointReady` happily reuses a session whose underlying process is no longer servicing it.
2. `verifyBrokerPid()` — pid-file content match + POSIX `ps` command-line check before any SIGTERM. If any check fails → `killProcess: null` → socket/pidfile cleanup only, no signal, trunk behavior.
3. `defaultKillProcess()` — SIGTERM with swallowed errors, only wired into rotation when `verifyBrokerPid()` confirms ownership.

Commit message rewritten to match the scope.

Dimaris-nsk:
Good catch @chatgpt-codex-connector — valid P1, the script-name substring alone was too broad and would cross-kill a foreign broker after PID recycle.

Tightened in amended commit (`856fbd8`): `verifyBrokerPid()` now also requires the session's unique `--pid-file` and `--endpoint` paths to appear in the live `ps` command line.

```js
if (!cmd.includes("app-server-broker.mjs")) return false;
if (!cmd.includes(`--pid-file ${session.pidFile}`)) return false;
if (session.endpoint && !cmd.includes(`--endpoint ${session.endpoint}`)) return false;
return true;
```

Why this is instance-specific:

- `spawnBrokerProcess` spawns with `serve --endpoint <endpoint> --cwd <cwd> --pid-file <pidFile>` (broker-lifecycle.mjs L61).
- Both `endpoint` and `pidFile` live inside `sessionDir = createBrokerSessionDir()`, which is an `os.tmpdir()` `mkdtempSync` — unique per spawn.
- A recycled PID belonging to a foreign broker (different workspace, different \$CODEX_HOME) will therefore carry a different `--pid-file` / `--endpoint` arg value, so at least one of the two matches fails → `verifyBrokerPid` returns false → rotation falls through to `killProcess: null` (trunk behavior: socket/pidfile cleanup only, no SIGTERM).

`session.endpoint` is gated so legacy sessions written before this PR (no `endpoint` field saved) still validate on `pid-file` alone rather than failing open — same safe default on mismatch.

If maintainers prefer something stronger (e.g. a broker-written identity token file separate from `ps` scraping), happy to switch — let me know.

Issue #234, opened 2026-04-16
Skill(codex:rescue) infinite-recurses when invoked programmatically from main agent


lwbvv:
## Summary

Calling `Skill(codex:rescue)` programmatically from the main Claude Code
agent (not via the user-typed `/codex:rescue` slash command) triggers an
infinite recursion that hangs the session until manual cancellation. The
Codex CLI is never invoked.

Confirmed on plugin versions **1.0.2** and **1.0.3**.

## Repro

From the main agent (not the user's slash-command path):

```
Skill(codex:rescue, args: "<any rescue task>")
```

## Observed behavior

- UI appears frozen; no Codex job is created.
- No entry under `~/.claude/plugins/data/codex-openai-codex/state/<project>/state.json`.
- No log file under `~/.claude/plugins/data/codex-openai-codex/state/<project>/jobs/`.
- After ~3+ minutes, user must cancel manually.

## Root cause (from subagent fork log)

Timeline from `~/.claude/projects/<project>/<session>/subagents/agent-*.jsonl`
(`agentType: "general-purpose"`):

```
T+0s    fork subagent starts, loads rescue.md instructions + raw user request
T+2s    "I'll check for a resumable rescue thread first"
T+3s    Bash: codex-companion.mjs task-resume-candidate -> available: false ✓
T+15s   ❌ Skill(codex:codex-rescue) attempted -> "Unknown skill: codex:codex-rescue"
T+20s   ❌ recovery attempt: Skill(codex:rescue) called again (recurses into self)
T+217s  hang; user cancels
```

Why:

1. `commands/rescue.md` frontmatter sets `context: fork`, so invocation
   forks a `general-purpose` subagent and loads `rescue.md` body as its
   instructions.
2. The body instructs: *"Route this request to the `codex:codex-rescue`
   subagent."*
3. `codex:codex-rescue` is an **agent** (`Agent` tool's `subagent_type`),
   not a **skill** (`Skill` tool).
4. The forked general-purpose agent resolves the ambiguous wording by
   trying `Skill(codex:codex-rescue)` → `Unknown skill` error.
5. As a fallback, it re-invokes `Skill(codex:rescue)` — which forks
   itself again → infinite recursion → hang.

## Why the slash command works but programmatic call doesn't

When the user types `/codex:rescue`, Claude Code's slash-command router
handles dispatch and does not fall into the "route to subagent" prose
resolution path. When the main agent calls `Skill(codex:rescue)`, the
[...]

Issue #122, opened 2026-04-03
codex:codex-rescue agent times out before Codex task completes on large diffs


Anfed-2030:
## Description

When using `codex:codex-rescue` subagents to dispatch review tasks via the `codex-companion.mjs task` command, the agent's Bash call times out (300s) before Codex finishes its analysis on large diffs. The Codex task continues running in the background, but the agent returns empty-handed and the results are never delivered to the parent conversation.

## Reproduction

1. Dispatch a review task via `codex:codex-rescue` on a diff with 30+ files / ~2000 lines
2. Codex starts a thorough multi-pass analysis (44+ file reads observed)
3. The `codex-companion.mjs task` command blocks waiting for Codex to finish
4. After 300 seconds, the Bash timeout in the Claude Code agent fires
5. The agent returns with "task dispatched, running in background" — no results
6. Codex eventually finishes, but the verdict sits in the job log with no way to surface it back

## Observed behavior

The Codex task log shows active work throughout:
```
[codex] Assistant message captured: I'm separating "new bug introduced by the branch" from "pre-existing path...
[codex] Command completed: git show '...:lib/training/workout-lifecycle.ts' (exit 0)
[codex] Command completed: git grep -n "deleted_at" ... (exit 0)
... (44+ commands executed)
```

But the agent that dispatched it has already returned empty.

## Expected behavior

One of:
1. **Streaming partial results** — the companion script could stream findings as they're discovered, so the agent can return intermediate results before timeout
2. **Async result retrieval** — the companion script could return a job ID immediately, and provide a `codex-companion.mjs result <job-id>` command that the agent can poll
3. **Configurable timeout** — allow the agent or companion script to set a longer timeout for large diffs
4. **Automatic result callback** — when the task completes, write results to a known location that the parent session can detect

## Impact

For small diffs (1-5 files), Codex completes within the timeout and results are returned normally. For larger diffs (30+ files), Codex reviews are effectively lost — the analysis runs but nobody sees the output. This makes Codex unreliable as a review gate for substantial PRs, which are exactly the ones that benefit most from a second opinion.

## Environment

- macOS Darwin 25.3.0
- Node.js v20.19.6
- Plugin version: 1.0.2 (commit 8e403f9)
- Claude Code v2.1.90
- Bash timeout: 300,000ms (5 minutes)

PR #296, opened 2026-05-05
Harden rescue task handoff failures

pengyou200902:
## Issue

Follow-up to #234 and #235. PR #235 fixed the original `/codex:rescue` recursion by routing the command through the `Agent` tool instead of re-entering `Skill(codex:rescue)`. That fixed the missing-Agent/fork routing bug, but it left adjacent duplicate-handoff paths outside its scope.

Those remaining paths happen when the rescue forwarder's first `task` attempt fails before returning a useful terminal result. In particular, shell-quoted prompt forwarding is fragile for multiline or shell-sensitive task text, ambiguous model handling can pass an unintended model name, and the forwarder contract previously said to return nothing on failure, which gives the caller no clear terminal result.

## Fix

- Add explicit `task --stdin` support so the companion script reads forwarded task text directly instead of receiving it as a shell-quoted CLI argument.
- Reject mixed prompt sources with `--stdin` (`--prompt-file` or positional prompt text) so prompt transport is deterministic.
- Allow empty `--stdin` only for resume runs, preserving resume-only fallback to the default continue prompt.
- Tighten rescue model forwarding: pass only explicit user-requested model strings, with the documented `spark` -> `gpt-5.3-codex-spark` alias exception.
- Replace the empty failure response contract with a non-retry contract: return failure output exactly once and stop.
- Drop the stale `experimentalRawEvents` app-server request field so the build matches the current generated Codex protocol types.

## Tests

- `npm run build`
- `node --test tests/commands.test.mjs`
- `node --test tests/runtime.test.mjs -t "task --stdin"`
- `npm test`
- PR CI passes on latest commit

### DIFF
--- a/plugins/codex/agents/codex-rescue.md
+++ b/plugins/codex/agents/codex-rescue.md
@@ -30,6 +30,7 @@ Forwarding rules:
 - Leave model unset by default. Only add `--model` when the user explicitly asks for a specific model.
 - If the user asks for `spark`, map that to `--model gpt-5.3-codex-spark`.
 - If the user asks for a concrete model name such as `gpt-5.4-mini`, pass it through with `--model`.
+- Copy the user's requested model string into `--model` character-for-character, except for the documented `spark` alias. Do not substitute, shorten, normalize, or guess model names. Never emit a model name the user did not literally type or whose alias they did not literally type.
 - Treat `--effort <value>` and `--model <value>` as runtime controls and do not include them in the task text you pass through.
 - Default to a write-capable Codex run by adding `--write` unless the user explicitly asks for read-only behavior or only wants review, diagnosis, or research without edits.
 - Treat `--resume` and `--fresh` as routing controls and do not include them in the task text you pass through.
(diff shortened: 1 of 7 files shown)


lzrong0203:
Independent confirmation of the failure mode this PR targets, from a field incident we then verified forensically and reproduced mechanically. Sharing because it directly exercises both halves of this fix (the non-retry contract and `--stdin`), and it isolates which part of the problem remains out of scope.

## Field incident (v1.0.4; the rescue instruction files and broker/tracked-jobs code are byte-identical through 1.0.6 and current `main`)

One rescue request → two concurrent server-side jobs:

| UTC | Event |
|---|---|
| 07:57:08 | Forwarder runs foreground `task "$(cat promptfile)"` (multiline prompt forced the temp-file detour this PR's `--stdin` eliminates) |
| 07:57:10 | Server-side job A created (`task-mrona8sx-…`) |
| 08:07:10 | Host Bash tool kills the CLI at its 600 s ceiling (exit 143, "Command timed out after 10m 0s") — the timeout output contains **no job id** |
| 08:08:01 | Forwarder retries with `task --background` → job B (`task-mrono8y5-…`), same prompt |
| 08:12:47 | Manual `cancel` of job A: interrupt reports "thread not found" — the orphaned turn had already completed unattended; only the manual cancel transitioned its record out of `running` |

Why the forwarder retried instead of stopping: the only failure rule it has is "If the Bash call fails or Codex cannot be invoked, return nothing" (agents/codex-rescue.md:42), and it is explicitly forbidden to call `cancel`/`status`/`result` (codex-rescue.md:27–28). A timeout-after-job-creation therefore leaves it with no cleanup path and no terminal result to return — exactly the gap the "return failure output exactly once and stop" contract closes.

## Mechanical reproduction (no model in the loop)

```
timeout 20 node scripts/codex-companion.mjs task "<any prompt that runs >20s>" ; echo $?   # → 124
```

The killed client's job survives (still `running` ~44 s later, ≈3× the task's natural duration), the server-side turn completes unattended, and the record leaves `running` only via manual cancel ("no active turn to interrupt"). So the orphan requires nothing model-driven — any client death does it.

## The residual half, for scoping

Code-wise the freeze happens because `app-server-broker.mjs:225–233` socket close/error handlers clear ownership but never interrupt the turn, and `lib/tracked-jobs.mjs:142–204` transitions a record only from the client's own try/catch — there is no pid-liveness or heartbeat check. This PR prevents the *duplicate submission*; a dead client still wedges its job record at `running` (the #122 / #372 / #432 / #486 family). The two fixes compose well — just noting the boundary so this PR isn't expected to solve that half.

Full forensic timeline (transcript + job JSONs) available if useful.

PR #171, opened 2026-04-07
fix: strip ANSI escape sequences from JSONL parsing

deepak-imh:
## Summary

- Adds `stripAnsi()` utility (`plugins/codex/scripts/lib/strings.mjs`) that removes CSI and OSC terminal escape sequences
- Applies stripping in both JSONL parsing paths before `JSON.parse()`:
  - `app-server.mjs` — `handleLine()` (readline-based client)
  - `app-server-broker.mjs` — socket `data` handler (broker multiplexer)
- Adds 7 unit tests covering bracketed paste mode, SGR codes, OSC sequences, clean passthrough, and edge cases

## Problem

Shells like zsh emit `\e[?2004h` (bracketed paste mode) during initialization. When Codex CLI is spawned as a subprocess, these escape sequences leak into stdout. The JSONL protocol reader then fails:

```
Failed to parse codex app-server JSONL: Unexpected token '', "[?2004h" is not valid JSON
```

This makes `/codex:review` and `/codex:adversarial-review` unusable for users with default zsh/bash configurations.

## Test plan

- [x] All 7 new `strings.test.mjs` tests pass
- [x] All 66 existing tests still pass (73 total, 0 failures)
- [ ] Manual verification: run `/codex:review` in Claude Code with zsh default shell

Fixes #23

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #24, opened 2026-03-31
fix: strip ANSI escape sequences before JSONL parsing

Mingsicogi:
## Summary

- Strip ANSI escape sequences (e.g. bracketed paste mode `\e[?2004h`) from JSONL lines before `JSON.parse()` in both `app-server.mjs` and `app-server-broker.mjs`
- Prevents `Failed to parse codex app-server JSONL: Unexpected token` errors when the shell environment leaks terminal control characters into the Codex CLI output stream

## Problem

When `/codex:review` runs in environments where the shell (zsh/bash) emits ANSI escape sequences during startup, these control characters appear in the JSONL stream. `JSON.parse()` then fails because `[?2004h` is not valid JSON.

```
Failed to parse codex app-server JSONL: Unexpected token '', "[?2004h" is not valid JSON
```

This affects both foreground and background execution via Claude Code's Bash tool.

## Fix

Added a regex-based ANSI escape sequence stripper before JSON parsing in the two JSONL consumers:
- `plugins/codex/scripts/lib/app-server.mjs` — `handleLine()` method
- `plugins/codex/scripts/app-server-broker.mjs` — socket data handler

Lines that become empty after stripping are silently skipped (same as the existing blank-line check).

## Test plan

- [x] Verified the regex correctly strips `\e[?2004h`, `\e[0m`, `\e]0;title\x07` and similar sequences
- [x] Verified valid JSON lines pass through unchanged
- [x] Existing test suite passes (1 pre-existing failure unrelated to this change)

Fixes #23

PR #311, opened 2026-05-09
fix: drop non-JSONL garbage on codex app-server stdout

### LINKED ISSUE #23: Review fails with JSONL parse error: bracketed paste mode escape sequence in output ([?2004h)
## Bug

`/codex:review` (both foreground and background) fails with a JSONL parse error when the Codex CLI outputs terminal escape sequences (bracketed paste mode) that corrupt the JSON stream.

## Error

```
Failed to parse codex app-server JSONL: Unexpected token '', "[?2004h" is not valid JSON
```

## Reproduction

1. Authenticate with `codex login`
2. Have a working tree with staged/unstaged changes
3. Run `/codex:review` from Claude Code (either foreground or background mode)
4. The companion script (`codex-companion.mjs`) attempts to parse the Codex CLI stdout as JSONL
5. Terminal escape sequence `[?2004h` (bracketed paste mode) is included in the output, breaking JSON parsing

Both foreground and background execution produce the same error:
```bash
# Foreground
node "codex-companion.mjs" review
# → Failed to parse codex app-server JSONL: Unexpected token '', "[?2004h" is not valid JSON

# Background (via Claude Code Bash tool with run_in_background: true)
node "codex-companion.mjs" review
# → Same error
```

## Root Cause

`[?2004h` is the ANSI escape sequence for enabling **bracketed paste mode**, typically emitted by shells (bash/zsh) during initialization. When the Codex CLI spawns a subprocess or the shell environment leaks these escape codes into stdout, the JSONL parser in the companion script receives non-JSON data and fails.

This likely happens because:
- The user's shell profile (`.zshrc` / `.bashrc`) emits escape sequences on startup
- The Codex CLI subprocess inherits the terminal settings and outputs control sequences
- The companion script does not strip ANSI escape sequences before parsing JSONL lines

## Suggested Fix

Strip ANSI escape sequences from the Codex CLI output before attempting JSONL parsing. For example:

```javascript
// Before parsing each line as JSON
const ansiRegex = /\x1b\[[0-9;]*[a-zA-Z]|\x1b\][^\x07]*\x07/g;
const cleanLine = rawLine.replace(ansiRegex, '').trim();
if (!cleanLine) continue; // skip empty lines after stripping
const parsed = JSON.parse(cleanLine);
```

## Environment

- macOS 15.5 (Darwin 24.6.0)
- Claude Code (CLI)
- Node.js (via Claude Code Bash tool)
- Shell: zsh
- codex-plugin-cc: latest
- Codex CLI: authenticated and working (`codex login` successful)

## Additional Context

- Other review-related issues (#3 sandbox variant, #14 EISDIR, #18 bwrap) are unrelated — this is a pure output parsing issue.
- The review job log shows `Starting Codex Review.` and then immediately fails, suggesting the error occurs during the initial JSONL handshake with the Codex app-server.

### LINKED ISSUE #310: Windows zh-TW: codex app-server JSONL parser crashes on Big5-encoded taskkill stdout leak
## Bug

On Windows systems with a non-English (specifically zh-TW / CP-950 / Big5) system locale, `/codex:review`, `/codex:task`, and the stop-time review gate fail with:

```
Failed to parse codex app-server JSONL: Unexpected token '�', "���\: PID "... is not valid JSON
```

The companion process exits with status 1 and the connection is torn down. Subsequent retries are racy: they sometimes succeed (when an existing broker happens to absorb the request), but a fresh broker spawn reliably reproduces the failure.

## Reproduction

1. Windows + zh-TW (Traditional Chinese / Taiwan) system locale (`chcp` returns `950`).
2. Have any project that starts the codex CLI with at least one MCP server that fails to start (e.g. an offline server, a misconfigured one, or one with handshake errors). On my machine `unityMCP`, `cheatengine`, and `ida-pro-mcp` all fail; any one of them is enough.
3. Tear down any running broker (`%LOCALAPPDATA%\Temp\cxc-*\broker.pid`) so the next companion invocation has to spawn a fresh one.
4. From the project directory:
   ```
   node "<plugin-cache>/scripts/codex-companion.mjs" task --json "anything"
   ```

The companion fails immediately with the parse error above. Re-running the same command after a successful broker is up succeeds — broker mode does not exercise the leaky path.

## Root cause

The JSONL parser in `plugins/codex/scripts/lib/app-server.mjs` `handleLine()` calls `JSON.parse()` on every line received from `codex.exe`'s stdout pipe and, on failure, calls `handleExit()` which rejects every in-flight request and tears down the connection.

Capturing raw bytes from inside `app-server-broker.mjs`'s internal `SpawnedCodexAppServerClient` (full hex dump below), the offending line is **not** corrupted JSON — it is a complete localized Windows OS message that has nothing to do with the protocol:

```
hex   = a6 a8 a5 5c 3a 20 50 49 44 20 ac b0 20 31 32 33 34 ...
cp950 = 成功: PID 為 1234 (PID 為 5678 的子處理程序) 的處理程序已終止。
utf-8 = ���\: PID �� 1234 (PID �� 5678 ���l�B�z�{��) ...
```

That is the success message printed by `taskkill /T /F /PID xxxx` — Windows' built-in process-tree killer — under the system locale (CP-950 / Big5). The companion process itself does not call `taskkill` on this code path, so the bytes have to be coming from inside `codex.exe` itself.

Walking the `app-server-broker.mjs` log on the failing run confirms it (timestamps abridged):

```
SPAWN proc.pid=77412 transport=direct
STDOUT len=209  "{\"id\":1,\"result\":{\"userAgent\":\"...\",\"codexHome\":...}}"      ← initialize ok
STDOUT len=94   "{\"method\":\"remoteControl/status/changed\",...}"                    ← ok
STDOUT len=652  "{\"method\":\"mcpServer/startupStatus/updated\",...starting...}"      ← ok
STDERR len=53   "���~: ��..."  // CP-950 "錯誤: 找不到..." (taskkill ENOENT path)
STDOUT len=246  "{\"method\":\"mcpServer/startupStatus/updated\",...failed...}"        ← ok
...
STDOUT len=111  "���\: PID �� 70260 (PID �� 1396 ��...) ..."                          ← LEAK
BAD LINE transport=direct  ← parser tears the connection down
```

The leak appears immediately after a `mcpServer/startupStatus/updated`-`failed` notification. The most plausible explanation is that the upstream codex CLI invokes `taskkill` to clean up an MCP child whose handshake failed, **and incorrectly leaves taskkill's stdout connected to its own stdout** (instead of capturing it or redirecting to `NUL`). On en-US Windows the leaked line happens to be ASCII (`SUCCESS: ...`) but still not JSON, so issue #23-style failures (different first byte) are the same bug class.

This is therefore primarily an upstream codex-CLI bug. I'll file it at `openai/codex` separately, link from here.

## Why this is distinct from #23 / #24 / #97 / #171

[#23][i23] reports the same outer symptom (`Failed to parse codex app-server JSONL`) but the leak source there is shell-init noise (`\x1b[?2004h`, bracketed paste mode). The proposed fixes ([#24][p24], [#97][p97], [#171][p171]) all add `stripAnsi()` and re-parse. That regex would not match the bytes above — they are not ANSI escapes, they are localized OS text — so JSONL parsing still fails after stripping.

Both bug classes share the same correct shape of fix at the plugin layer, though: **lines that cannot possibly be JSONL records should be dropped, not treated as protocol violations that tear the connection down.** A JSONL record always begins with `{` or `[`. Any line whose first non-whitespace, post-`stripAnsi` character is something else is definitively garbage from outside the protocol.

A PR that implements that one-line guard (and subsumes #23) is incoming.

## Why it's racy

Once a broker is alive, all subsequent connections from the companion go over the named pipe and never spawn a direct codex.exe in the parent process — so the leaky stdout never flows through the JSONL parser of the parent. The leak is still happening *inside* the broker process, but is parsed there and fails the same way; the broker just dies more quietly.

This also explains why running the same command twice in a row often shows different outcomes (first call: broker bootstrap fails on the leak; second call: existing broker handles it).

## Environment

| | |
|--|--|
| OS | Windows 11 Pro 26200, zh-TW system locale (CP-950) |
| Shell | PowerShell 7 + Git Bash (both reproduce) |
| Node | v22.16.0 |
| codex CLI | 0.130.0 |
| codex-plugin-cc | 1.0.4 |
| Reproducer | runs `taskkill` ⇒ Big5 success line bytes `a6 a8 a5 5c 3a 20 50 49 44 ...` |

## Workaround

Disable the stop-time review gate (`/codex:setup --disable-review-gate`) until a fix lands, OR keep a long-lived broker alive (do not let `%LOCALAPPDATA%\Temp\cxc-*` go stale).

[i23]: https://github.com/openai/codex-plugin-cc/issues/23
[p24]: https://github.com/openai/codex-plugin-cc/pull/24
[p97]: https://github.com/openai/codex-plugin-cc/pull/97
[p171]: https://github.com/openai/codex-plugin-cc/pull/171

lask3802:
## Summary

- Adds `lib/jsonl.mjs#cleanProtocolLine(rawLine)`: strips ANSI/OSC escape sequences, trims whitespace, and returns the cleaned line **only if** it starts with a JSON delimiter (`{` or `[`). Otherwise returns `null`.
- Rewires `lib/app-server.mjs` `handleLine()` to skip on `null` instead of treating non-JSON lines as protocol violations that tear the connection down.
- Adds `tests/jsonl.test.mjs` (11 cases) covering bracketed-paste prefix, OSC window-title, SGR colours, the exact CP-950 mojibake from #310, plain JSON passthrough, malformed-but-JSON-shaped fall-through to the parser, and non-string input.

## Problem

`handleLine()` calls `JSON.parse()` on every line read from the codex app-server stdout pipe and, on failure, calls `handleExit()` — rejecting every in-flight request and closing the connection. In practice the codex CLI occasionally emits bytes onto stdout that have nothing to do with the protocol. Two known sources:

1. **Terminal / shell init noise** (issue #23). zsh writes the bracketed-paste marker `\x1b[?2004h` when a subprocess inherits the parent's TTY. The user-visible error:
   ```
   Failed to parse codex app-server JSONL: Unexpected token '', "[?2004h" is not valid JSON
   ```
2. **Localized OS messages on Windows non-English locales** (issue #310 ⇄ root cause filed upstream as [openai/codex#21957](https://github.com/openai/codex/issues/21957)). On zh-TW (CP-950 / Big5), when `codex.exe` runs `taskkill /T /F` to clean up a failed MCP child and inadvertently routes taskkill's stdout to its own stdout, the JSONL stream receives:
   ```
   raw   : a6 a8 a5 5c 3a 20 50 49 44 20 ac b0 20 31 32 33 34 ...
   cp950 : 成功: PID 為 1234 ...
   utf-8 : ���\: PID �� 1234 ...
   ```
   which is the exact `Unexpected token '�', "���\: PID "...` reported in #310.

Both bug classes share the same correct fix at the plugin layer: **lines that cannot possibly be JSONL records should be dropped, not treated as protocol violations**. A JSONL record always begins with `{` or `[`. Any line whose first non-whitespace, post-strip-ANSI character is something else is definitively garbage from outside the protocol.

## Why this approach over the existing PRs (#24, #97, #171)

[#171][p171] (the most recent of three open PRs) introduces a `stripAnsi()` helper and applies it before `JSON.parse()`. That correctly fixes #23 — but does not fix #310, because the leaked bytes are not ANSI escapes. They are localized OS text. After `stripAnsi()` the line is still not JSON, the parser still throws, the connection still dies.

The `cleanProtocolLine()` guard in this PR is a strict superset:
- It still strips ANSI/OSC sequences (so #23 is covered identically to #171).
- It additionally drops anything whose post-strip first character is not `{` or `[` (so #310 is covered).

If maintainers prefer to land #171 first I'm happy to rebase this onto it (the regex code can move to `lib/strings.mjs#stripAnsi` and `cleanProtocolLine` becomes a one-liner that imports it). Either order works; the unit tests stay the same.

## What this PR explicitly does NOT do

- It does **not** patch the broker side (`app-server-broker.mjs`'s socket data handler). That handler reads JSONRPC requests *from the companion*, which sends well-formed JSON; there is no observed corruption source on that path. #171 does patch it for symmetry — happy to add the same here if you'd like.
- It does **not** fix the upstream stdio leak in `codex.exe`. That is filed as [openai/codex#21957](https://github.com/openai/codex/issues/21957) and is out of scope for this repo. This PR is defense-in-depth for any embedder of `codex app-server`.

## Test plan

- [x] `node --test tests/jsonl.test.mjs` — 11/11 pass.
- [x] `node --test tests/*.test.mjs` — all failures present on this branch (`state.test.mjs`, `runtime.test.mjs`, `process.test.mjs` — all Windows-only `os.tmpdir()` / MSYS path translation issues) reproduce against `main` from the same checkout, i.e. no regressions introduced by this PR.
- [x] End-to-end: stop-time review gate fired during a Claude Code session running on Windows zh-TW reliably reproduced the original failure pre-patch; with this patch applied to the plugin cache, three back-to-back `codex-companion.mjs task` invocations (each preceded by killing the broker to force the leaky direct-spawn path) now all return `{"status":0,...}`.
- [ ] Manual: `/codex:review` from a zsh-default Claude Code session (issue #23 case) — I do not have a zsh box handy; logically equivalent to PR #171's verification.

## Closes / refs

- Closes #310
- Refs #23 (also addressed by this PR — happy to mark Closes if you'd prefer to consolidate)
- Refs #24, #97, #171 (alternative approaches to the same shape of bug)
- Refs root-cause issue: [openai/codex#21957](https://github.com/openai/codex/issues/21957)

[p171]: https://github.com/openai/codex-plugin-cc/pull/171

### DIFF
--- a/plugins/codex/scripts/lib/app-server.mjs
+++ b/plugins/codex/scripts/lib/app-server.mjs
@@ -14,6 +14,7 @@ import { spawn } from "node:child_process";
 import readline from "node:readline";
 import { parseBrokerEndpoint } from "./broker-endpoint.mjs";
 import { ensureBrokerSession, loadBrokerSession } from "./broker-lifecycle.mjs";
+import { cleanProtocolLine } from "./jsonl.mjs";
 import { terminateProcessTree } from "./process.mjs";
 
 const PLUGIN_MANIFEST_URL = new URL("../../.claude-plugin/plugin.json", import.meta.url);
@@ -114,16 +115,20 @@ class AppServerClientBase {
     }
   }
 
-  handleLine(line) {
-    if (!line.trim()) {
+  handleLine(rawLine) {
+    const candidate = cleanProtocolLine(rawLine);
+    if (candidate === null) {
+      // Empty line, terminal noise (issue #23), or localized OS messages
+      // (e.g. CP-950 mojibaked taskkill SUCCESS on Windows zh-TW). Drop
+      // it instead of tearing the connection down — see lib/jsonl.mjs.
       return;
     }
 
     let message;
     try {
-      message = JSON.parse(line);
+      message = JSON.parse(candidate);
     } catch (error) {
-      this.handleExit(createProtocolError(`Failed to parse codex app-server JSONL: ${error.message}`, { line }));
+      this.handleExit(createProtocolError(`Failed to parse codex app-server JSONL: ${error.message}`, { line: rawLine }));
       return;
     }
 
--- a/plugins/codex/scripts/lib/jsonl.mjs
+++ b/plugins/codex/scripts/lib/jsonl.mjs
@@ -1,0 +1,60 @@
+/**
+ * Helpers for reading the Codex app-server JSONL protocol.
+ *
+ * The codex CLI's app-server occasionally emits non-JSON bytes onto stdout
+ * that have nothing to do with the protocol. Two known sources:
+ *
+ *   1. Terminal/shell init noise — e.g. zsh writes the bracketed-paste
+ *      marker `\x1b[?2004h` when a subprocess inherits the parent's TTY
+ *      (issue #23).
+ *
+ *   2. Localized OS messages on Windows non-English locales. On zh-TW
+ *      (CP-950 / Big5), when codex.exe runs `taskkill /T /F` to clean up
+ *      a failed MCP child and inadvertently routes taskkill's stdout to
+ *      its own stdout, we see the bytes `A6 A8 A5 5C 3A 20 50 49 44 ...`
+ *      ("成功: PID 為 xxxx ...") which decode under UTF-8 as
+ *      `���\\: PID ...`.
+ *
+ * Both cases historically tore the broker connection down with
+ * `Failed to parse codex app-server JSONL: Unexpected token …`. The
+ * client never recovers because `handleExit` rejects every in-flight
+ * request before any subsequent valid record arrives.
+ *
+ * `cleanProtocolLine` is the conservative guard:
+ *
+ *   - Strip CSI / OSC ANSI escape sequences from the raw line.
+ *   - Trim whitespace.
+ *   - Require the first remaining character to be `{` or `[`. JSONL
+ *     records cannot start with anything else, so any other prefix is
+ *     definitively garbage and is dropped.
+ *
+ * Lines that pass these checks are still parsed with `JSON.parse`; if
+ * that fails the caller surfaces a real protocol error as before.
+ */
+
+// CSI:  ESC [ … final-byte                  (e.g. \x1b[?2004h, \x1b[31m)
+// OSC:  ESC ] … (BEL | ESC \)               (e.g. window-title sets)
+// eslint-disable-next-line no-control-regex
+const ANSI_ESCAPE_RE = /\x1b\[[0-9;?]*[a-zA-Z]|\x1b\][^\x07]*(?:\x07|\x1b\\)/g;
+
+/**
+ * Returns a JSON-shaped candidate string for the given raw line, or
+ * `null` if the line should be skipped because it cannot be valid JSONL.
+ *
+ * @param {string} rawLine
+ * @returns {string | null}
+ */
+export function cleanProtocolLine(rawLine) {
+  if (typeof rawLine !== "string") {
+    return null;
+  }
+  const cleaned = rawLine.replace(ANSI_ESCAPE_RE, "").trim();
+  if (!cleaned) {
+    return null;
+  }
+  const firstChar = cleaned.charCodeAt(0);
+  if (firstChar !== 0x7B /* { */ && firstChar !== 0x5B /* [ */) {
+    return null;
+  }
+  return cleaned;
+}
--- a/plugins/codex/scripts/lib/jsonl.mjs
+++ b/plugins/codex/scripts/lib/jsonl.mjs
@@ -32,16 +32,25 @@
  * that fails the caller surfaces a real protocol error as before.
  */
 
-// CSI:  ESC [ … final-byte                  (e.g. \x1b[?2004h, \x1b[31m)
-// OSC:  ESC ] … (BEL | ESC \)               (e.g. window-title sets)
+// CSI: ESC [ <params> <intermediates> <final>
+//      params       = 0x30..0x3F (digits, ?, etc.)
+//      intermediates= 0x20..0x2F (space, !, ", $, %, &, ', (, ), *, +, ,, -, ., /)
+//      final        = 0x40..0x7E (any byte in @, A..Z, [, \, ], ^, _, `, a..z, {, |, }, ~)
+// This grammar matches the original ECMA-48 spec and covers cases like
+// bracketed-paste *content* markers `\x1b[200~` / `\x1b[201~` (final
+// byte `~`) which a letter-only `[a-zA-Z]` final misses.
+// OSC: ESC ] <text> (BEL | ESC \)            e.g. window-title sets
 // eslint-disable-next-line no-control-regex
-const ANSI_ESCAPE_RE = /\x1b\[[0-9;?]*[a-zA-Z]|\x1b\][^\x07]*(?:\x07|\x1b\\)/g;
+const ANSI_ESCAPE_RE = /\x1b\[[\x30-\x3F]*[\x20-\x2F]*[\x40-\x7E]|\x1b\][^\x07]*(?:\x07|\x1b\\)/g;
 
 /**
  * Returns a JSON-shaped candidate string for the given raw line, or
  * `null` if the line should be skipped because it cannot be valid JSONL.
  *
- * @param {string} rawLine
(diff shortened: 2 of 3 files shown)


lask3802:
Thanks @copilot for the review — both points addressed in 2dc20bf:

1. **CSI final-byte range (jsonl.mjs:38).** Good catch on `\x1b[200~` / `\x1b[201~` — bracketed-paste *content* markers (final byte `~`) sit between the on/off toggles, so a letter-only final would have left them in place and then the `{` / `[` guard would have *dropped* a perfectly valid JSON record sandwiched between them. Regex broadened to the full ECMA-48 CSI grammar (params `0x30..0x3F`, intermediates `0x20..0x2F`, final `0x40..0x7E`). New regression test covers `\x1b[200~{...}\x1b[201~` plus the boundary finals `@` (0x40) and `` ` `` (0x60).

2. **JSDoc accuracy (jsonl.mjs:49).** Updated `@param` to `unknown` to match the runtime contract (non-string short-circuits to `null`, asserted by a unit test). Note added in the prose too so editors don't mislead callers into pre-narrowing.

Full test run: `node --test tests/jsonl.test.mjs` → 12/12 pass.

PR #335, opened 2026-05-19
feat: per-job event stream + watchdog + compact recovery (1.1.0)

suminerProxy:
## Why

When Claude Code's main loop delegates a coding task to Codex via
`/codex:rescue`, the subagent makes a single synchronous `Bash` call into
the companion `task` command and waits for it to return. If Codex stalls
(prompt too long, model hang, tool deadlock, API rate limit), the main
loop stalls too — there is no out-of-band signal it can poll to detect
the freeze, and `/codex:status` only exposes coarse job-level phase.

This PR adds the missing channel: a per-job NDJSON event stream the main
Claude loop can poll with `/codex:events`, plus a stall watchdog that
emits a `phase:"stuck"` event when codex goes quiet, and a
protocol-native `/codex:compact` recovery path for context overflow.
`/codex:rescue` now defaults to `--background` so the main loop never
blocks waiting on the synchronous Bash call.

## What ships in 1.1.0

- `/codex:events <job-id>` — incremental polling of normalized codex
  notifications. Supports `--since` / `--after-seq` / `--limit` / `--json`.
- `/codex:compact <thread-id>` — wraps `thread/compact/start` for the
  cancel → compact → resume-with-amended-prompt flow.
- `/codex:rescue` default flipped to `--background`. The main loop polls
  `/codex:events` instead of blocking.
- 60s per-job stall watchdog (override via `CODEX_COMPANION_STALL_SECONDS`)
  emits `{type:"watchdog", phase:"stuck"}`. Does not cancel.
- New `{type:"job/exited"}` terminal event — single source of truth for
  end-of-job. Fixes a correctness bug where codex turn failures were
  reported as `phase:"completed"` because `runTrackedJob` resolves with
  `exitStatus != 0` rather than throwing.
- `runAppServerTurn` surfaces `usage` at top level; real-time updates
  flow via `thread/tokenUsage/updated` events (`phase:"metering"`).
- codex CLI 0.131 notification coverage extended: `thread/status/changed`,
  `warning`, `thread/tokenUsage/updated`, plus item types `userMessage`,
  `assistantMessage` / `agentMessage`, `reasoning`. E2E against real
  codex 0.131 shows `unknown` phase count goes from 5/9 (56%) to 0/13.

## Architecture

The event capture point is `captureTurn`'s notification handler, NOT the
broker. The broker is a single-flight, single-stream-owner router
(`app-server-broker.mjs:84-90`) — it cannot do per-job fan-out without a
rewrite. Instead, `captureTurn` accepts an `onNotification` callback that
runs after `applyTurnNotification`, normalizes the message, and the
task-worker appends to `{stateDir}/jobs/{jobId}.events.ndjson`. The
watchdog `setInterval` shares a closure with the same handler in the
worker process so cross-process state synchronization is unneeded.

Single-job-at-a-time is preserved (matches broker reality). Concurrent
multi-job and broker fan-out are deferred to a future PR.

## Test plan

- [x] `npm test`: 128 / 128 pass. Adds 42 new tests (10 events API, 25
      normalize, 7 events command integration).
- [x] Test isolation: separate prior PR (#334 if accepted) unsets
      `CLAUDE_PLUGIN_DATA` and `CODEX_COMPANION_SESSION_ID` in
      `tests/helpers.mjs` so `npm test` works inside a plugin host
      runtime. This branch includes that commit.
- [x] No breaking changes — `/codex:review`, `/codex:adversarial-review`,
      `/codex:result`, `/codex:cancel`, `/codex:status` all keep the same
      signatures and output.

## Verification status

Verified end-to-end against codex CLI 0.131.0-alpha.9:

- ✅ **Success path**: 13 events, `job/exited phase:"completed"
  exitCode:0`, 0 unknown-phase events. Token usage event + idle event +
  reply preview ("Codex replied: pong") all surface correctly.
- ✅ **Failed path**: dispatch with `--effort minimal "..."` triggers
  codex's `invalid_request_error` (web_search incompatible with
  minimal effort). Result: `job/exited phase:"failed" exitCode:1
  errorMessage:"Task did not complete successfully (see prior events
  for details)."` Stored `state.json` shows `status:"failed"
  phase:"failed"` — the verdict matches what runs through the events
  stream. This is the bug fix in `3a0ae2d`: before this PR, `job/exited`
  reported `phase:"completed" exitCode:0` because the worker keyed off
  thrown exceptions rather than `execution.exitStatus`.
- ✅ **Stall watchdog**: `CODEX_COMPANION_STALL_SECONDS=5` + `--effort
  xhigh` reasoning task. After 8.54s of silence inside a Reasoning item
  (longer than the 5s threshold), the worker emitted
  `{type:"watchdog", phase:"stuck", stallMs:8540, since:"<ts of last
  event>"}`. Codex resumed naturally a moment later; the watchdog did
  NOT cancel the job and did NOT re-fire spuriously when subsequent
  short quiet periods (~2s) occurred. The `since` timestamp aligns
  exactly with the previous notification's `ts`.
- ✅ **Compact rejection path**: smoke test `/codex:compact <bogus-id>`
  returns `attempted:true, compacted:false, transport:"direct",
  detail:"invalid thread id: ..."` — the codex-side error is preserved.

## Open schema unknowns (does not block this PR)

- `thread/compact/start` **success** payload shape is undocumented for
  codex CLI 0.131. The wrapper preserves it verbatim under `.result`
  for forward-compat. The rejection path is structured; the success
  path shape is the open question for the next docs revision.
- `thread/tokenUsage/updated` payload field names not confirmed against
  codex CLI 0.131 docs. The normalize function tries `inputTokens` /
  `outputTokens` / `cachedInputTokens` and a few aliases, falls back to
  a stable "Token usage updated." label. The event itself still fires
  correctly (`phase:"metering"`); only the human-readable token count
  in `.message` requires knowing the exact field names.

Both are tracked at the relevant code paths.

## Build status

CI on this branch will need to handle the existing upstream
incompatibility with codex CLI 0.131 (`requestAttestation` /
`experimentalRawEvents` schema drift in
`plugins/codex/scripts/lib/app-server.mjs` and `lib/codex.mjs:64`).
That is a pre-existing condition on `main` and is intentionally not
addressed in this PR to keep the diff focused on observability.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -5,13 +5,13 @@
   },
   "metadata": {
     "description": "Codex plugins to use in Claude Code for delegation and code review.",
-    "version": "1.0.4"
+    "version": "1.1.0"
   },
   "plugins": [
     {
       "name": "codex",
       "description": "Use Codex from Claude Code to review code or delegate tasks.",
-      "version": "1.0.4",
+      "version": "1.1.0",
       "author": {
         "name": "OpenAI"
       },
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@openai/codex-plugin-cc",
-  "version": "1.0.4",
+  "version": "1.1.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@openai/codex-plugin-cc",
-      "version": "1.0.4",
+      "version": "1.1.0",
       "license": "Apache-2.0",
       "devDependencies": {
         "@types/node": "^25.5.0",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@openai/codex-plugin-cc",
-  "version": "1.0.4",
+  "version": "1.1.0",
   "private": true,
   "type": "module",
   "description": "Use Codex from Claude Code to review code or delegate tasks.",
--- a/plugins/codex/.claude-plugin/plugin.json
+++ b/plugins/codex/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "codex",
-  "version": "1.0.4",
+  "version": "1.1.0",
   "description": "Use Codex from Claude Code to review code or delegate tasks.",
   "author": {
     "name": "OpenAI"
--- a/plugins/codex/CHANGELOG.md
+++ b/plugins/codex/CHANGELOG.md
@@ -1,5 +1,47 @@
 # Changelog
 
+## 1.1.0
+
+Observability rework for the main-loop orchestration case: the consumer of
+"what is codex doing right now" is the calling Claude session, not a human
+dashboard. Adds a per-job NDJSON event stream the main loop can poll, plus
+a protocol-native recovery path for context overflow.
+
+- `/codex:events <job-id>`: new slash command. Streams normalized codex
+  notifications from `{stateDir}/jobs/{jobId}.events.ndjson`. Supports
+  `--since <iso>` / `--after-seq <n>` / `--limit <n>` / `--json` for
+  incremental polling. Each event carries `seq`, `ts`, `method`, `phase`,
+  `itemType`, `message`, and the raw payload.
+- `/codex:compact <thread-id>`: new slash command wrapping codex
+  app-server's `thread/compact/start`. Protocol-native recovery for
+  "prompt too long" — typical flow is cancel → compact → resume with an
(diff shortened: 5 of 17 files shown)


suminerProxy:
Closing — switching to a fork-internal PR (suminerProxy/codex-plugin-cc) so the owner can self-merge. Re-opening if upstream eventually wants this contribution.

Pelton:
Was the watchdog / compact-recovery / event-stream work here folded into another effort — PR #343 looks adjacent on broker cleanup, or is the watchdog path being addressed elsewhere? Tracking the broker-leak issue and want to make sure I'm following the active thread.

Issue #248, opened 2026-04-19
Stop review gate blocks Claude on infrastructure errors, causing rewake loops


syy674998887:
## Summary

`scripts/stop-review-gate-hook.mjs` conflates two distinct outcomes into the same `{ decision: "block" }` hook
output:

1. **Genuine BLOCK** — Codex task returned `BLOCK: <reason>` (a real policy decision)
2. **Infrastructure failure** — the Codex task could not produce a valid ALLOW/BLOCK decision at all (timeout,
   non-zero exit, empty output, invalid JSON)

Combined with Claude Code's rewake-on-block semantic, any transient upstream failure (ChatGPT usage limit, auth
expiry, network blip, Codex app-server crash) creates a sustained rewake loop. Claude replies → hook fires → Codex
still failing → block → Claude rewakes → … The loop burns Claude tokens continuously and produces zero useful review,
and the user has no way out other than `/codex:setup --disable-review-gate`.

## Reproduction

1. Enable the gate: `/codex:setup --enable-review-gate`
2. Induce any transient Codex-side failure. Easiest repro: exhaust your ChatGPT / Codex usage quota.
3. Finish any Claude turn (even a trivial one).
4. The hook's spawned `codex task` exits with `status: 1` and empty output. The hook emits `{ decision: "block" }`.
5. Claude Code rewakes the model; Claude replies; the hook fires again with the same failure; loop continues until
   the user manually disables the gate or the quota resets.

In a real session I observed ~10 consecutive failed jobs in
`~/.claude/plugins/data/codex-openai-codex/state/<workspace>-<hash>/state.json`, all with:

```
"summary": "You've hit your usage limit. To get more access now, send a request to your admin or try again at 7:32
PM."
"status": "failed"
```

Each one produced a BLOCK decision to Claude Code, despite no review having actually run.

## Root cause

In `scripts/stop-review-gate-hook.mjs`:

- `parseStopReviewOutput()` returns `{ ok: false, reason }` for **both** `BLOCK: ...` responses and any
  unexpected/empty output.
- `runStopReview()` returns `{ ok: false, reason }` for timeouts, non-zero exits, and JSON parse failures.
- `main()` treats every non-`ok` result identically:

  ```js
  if (!review.ok) {
    emitDecision({ decision: "block", reason: ... });
    return;
  }
  ```

There is no code path for "the review didn't run — don't block, just warn". All infrastructure failures are mapped to
the same blocking decision as a real Codex BLOCK.

[...]

Issue #191, opened 2026-04-09
Stop hook hangs on Windows when review gate is disabled (stdin blocking)


emamimohsen:
On Windows with Git Bash as the configured shell, the Stop hook opens a new terminal window and hangs indefinitely (up to
  the 900s timeout) after every Claude response, even when the review gate is disabled.

  stop-review-gate-hook.mjs calls fs.readFileSync(0, "utf8") (blocking stdin read) at the top of main(), before checking
  whether config.stopReviewGate is enabled. On Windows with Git Bash, the hook runner does not properly pipe/close stdin for
  the spawned Node.js process, so the read blocks forever.

  The gate check at line 154 (if (!config.stopReviewGate) { return; }) never runs because execution is stuck at line 143
  (readHookInput()).

  Environment:
  - OS: Windows 11 Enterprise
  - Shell: Git Bash (C:/Users/.../Git/usr/bin/bash.exe)
  - Claude Code with Codex plugin v1.0.3
  - Review gate: disabled (stopReviewGate: false)

lucasmiachon-blip:
+1 Confirming this on Windows 11 + Git Bash + plugin 1.0.3 (verified the v1.0.4 manifest is byte-identical, so still affected).

## Empirical evidence

Direct manual run with `echo '{}' |` to provide stdin returns cleanly:

```
$ time (echo '{}' | node "stop-review-gate-hook.mjs")
real  0m0.644s
exit  0
```

Under the actual Claude Code harness (no piped stdin), the same hook hangs until the manifest's 900ms timeout, after which the harness emits:

```
Stop hook error: Failed with non-blocking status code: No stderr output
```

Signature is consistent with `fs.readFileSync(0)` blocking on a stdin handle the harness leaves open without writing — exactly as the original report describes.

## Confirmed config

- `state.json`: `config.stopReviewGate: false`
- Plugin: codex 1.0.3, also tested against the v1.0.4 manifest — identical `5 / 5 / 900` timeouts in `hooks/hooks.json`.

## Suggested fix directions

1. **Primary (matches root cause):** move `readHookInput()` to *after* the `config.stopReviewGate` check at line 154. When the gate is off, the script can short-circuit without reading stdin at all. This eliminates the hang in the common (default) case.

2. **Defensive alternative:** non-blocking stdin read with a short fallback — e.g. `Promise.race` between an event-based stdin reader and `setTimeout(50, () => ({}))`. Useful if any future code path needs the input even when the gate is off.

[...]

PR #352, opened 2026-05-27
fix(stop-review-gate-hook): always emit valid JSON on stdout

AliceLJY:
## Summary

`scripts/stop-review-gate-hook.mjs` returns silently in three no-decision paths in `main()`:

1. `if (!config.stopReviewGate)` — the default, hit on every Stop event when the user has not opted into the review gate
2. `if (setupNote)` — Codex not yet set up
3. The trailing path where the review came back `ALLOW`

In all three, `logNote()` writes to stderr but stdout is never touched. Newer hook executors (observed in current Claude Code / Codex CLI builds) strict-parse hook stdout as JSON. Empty stdout is no longer treated as "no decision" — it surfaces as a visible error after every Claude turn:

```
Stop hook (failed)
error: hook returned invalid stop hook JSON output
```

Older Claude Code tolerated empty stdout, which is why this only became visible recently.

## Fix

Emit `{}` (a no-decision JSON object) before each silent `return`. Three-line change, no behavior change other than satisfying the strict parser. The block path is untouched — it already emits a `{decision: "block", reason}` object.

## Test plan

- [x] `echo '{"hook_event_name":"Stop","session_id":"test","stop_hook_active":false}' | node scripts/stop-review-gate-hook.mjs` → stdout is `{}`, exit 0
- [x] `JSON.parse(stdout)` succeeds
- [x] stderr `logNote` messages preserved (no regression in user-facing diagnostics)
- [x] `decision: "block"` path unchanged

## Related issues / context

I searched open issues and PRs for `stop-review-gate`, `empty stdout`, `hook returned invalid`, `JSON output` — none describe this specific defect. The closest are:

- #191 (Stop hook hangs on Windows when review gate disabled) — different root cause (stdin blocking, not stdout shape)
- #248 (Stop review gate blocks Claude on infrastructure errors) — touches `parseStopReviewOutput`, but only the gate-enabled paths

Both are orthogonal to this fix.

### DIFF
--- a/plugins/codex/scripts/stop-review-gate-hook.mjs
+++ b/plugins/codex/scripts/stop-review-gate-hook.mjs
@@ -153,6 +153,7 @@ function main() {
 
   if (!config.stopReviewGate) {
     logNote(runningTaskNote);
+    emitDecision({});
     return;
   }
 
@@ -160,6 +161,7 @@ function main() {
   if (setupNote) {
     logNote(setupNote);
     logNote(runningTaskNote);
+    emitDecision({});
     return;
   }
 
@@ -173,6 +175,7 @@ function main() {
   }
 
   logNote(runningTaskNote);
+  emitDecision({});
 }
 
 try {
--- a/tests/runtime.test.mjs
+++ b/tests/runtime.test.mjs
@@ -1894,7 +1894,7 @@ test("stop hook logs running tasks to stderr without blocking when the review ga
   });
 
   assert.equal(blocked.status, 0, blocked.stderr);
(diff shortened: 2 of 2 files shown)


AliceLJY:
Addressed the P1 review note from @chatgpt-codex-connector in commit a1020ad — updated the three `stdout.trim() === ""` assertions in `tests/runtime.test.mjs` (lines 1897, 1925, 1950) to expect `"{}"` so they match the new no-decision contract.

Verified locally: all 5 stop-hook tests now pass.

Issue #213, opened 2026-04-13
feat: support user-level defaults for companion config (e.g. stopReviewGate)


bk-squared:
## Summary
Currently `codex-companion` stores config per git-repo root, with no fallback source. Options like `stopReviewGate` have to be toggled individually for every repo. It would be great to support a user-level defaults file that workspace state can override.

## Related
- #44 — complementary, not duplicate. That issue proposes persisting `--model` / `--effort` defaults in the existing per-workspace state. This issue proposes a **lower-priority, user-level baseline** that applies across all workspaces. If both land, #44 gives you per-workspace defaults and this gives you cross-workspace ones; they compose cleanly.
- #59 — adjacent state-resolution bug (setup writes to one state dir, Stop hook reads from another). Not the same problem, but anyone touching `resolveStateDir` / `getConfig` will want to read both together.

## Scope (explicit)
Per-workspace state always wins. The user-level file is a **baseline only** — never overrides an explicit per-repo setting. No existing state schema changes.

## Current behavior
`scripts/lib/workspace.mjs` resolves the workspace as the nearest git repo root, and `scripts/lib/state.mjs` keys state under `\$CLAUDE_PLUGIN_DATA/state/<slug>-<sha256(realpath)[:16]>/state.json`. `getConfig()` only reads that per-repo file — there is no lookup of a shared/user-level default. So enabling the review gate across N sibling repos under a shared workspace requires N separate \`codex-companion setup --enable-review-gate\` invocations.

## Use case
I maintain a workspace with many sibling git repos (research / infra / teaching subtrees). I want the stop-time review gate enabled by default everywhere, while still being able to disable it in specific repos that don't need it. Running the setup command repeatedly across every new clone is friction, and easy to forget.

This is the same ergonomic pattern that git (\`~/.gitconfig\` → repo \`.git/config\`), npm (\`~/.npmrc\` → project), and Claude Code itself (\`~/.claude/settings.json\` → project \`.claude/settings.json\`) already provide.

## Proposed behavior
Introduce a lower-priority defaults file, e.g. \`\$CLAUDE_PLUGIN_DATA/defaults.json\` (or \`~/.config/codex-companion/defaults.json\`). \`getConfig()\` would shallow-merge in this order:

\`\`\`
built-in defaults  <  user defaults  <  per-workspace state
\`\`\`

The setup command could gain \`--global\` (or \`--user\`) to write the defaults file instead of the workspace state:

\`\`\`
[...]

PR #353, opened 2026-05-27
feat(stop-review-gate): add monorepo mode for sibling git repos

PAzter1101:
## Summary

The stop-time review gate currently resolves a single workspace via `git rev-parse --show-toplevel` from the session cwd. In layouts where several independent git repos sit under one parent directory (a common monorepo pattern — sibling clones managed by convention rather than git submodules), changes Claude makes in a *sibling* repo are silently skipped: the hook runs review only in the primary repo, sees no diff there, returns `ALLOW`, the session ends without the review actually covering the work.

This PR adds an opt-in `monorepoMode` config flag (default `false`, fully backwards-compatible). When enabled on the primary repo, the `Stop` hook scans the parent directory for sibling git repos and runs an independent stop-time review for each one that

- has its own `stopReviewGate` enabled, and
- has uncommitted changes.

Empty-diff siblings are skipped (Codex would just reply `ALLOW` and waste tokens). If any review blocks, the combined reason is returned and the gate blocks the Stop.

When `monorepoMode` is off (the default), behavior is identical to today: only the primary workspace is reviewed.

## Configuration

```bash
/codex:setup --enable-monorepo
/codex:setup --disable-monorepo
```

Setup must still be run separately in each sibling repo to enable the gate for it (`/codex:setup --enable-review-gate`). The hook respects each sibling's own state — a repo whose gate is off is skipped.

## Why opt-in

- Auto-scanning the parent dir without an explicit opt-in could surprise users with single-repo workflows where the parent happens to contain other git checkouts (e.g. `~/code/` with dozens of unrelated clones).
- Keeps single-repo semantics deterministic and free of any FS walk.

## Test plan

- `npm test` — 86 → 88 tests (two new state tests for the flag); only pre-existing failures remain (#133-style flake on `resolveStateDir` and the `status`/`result` formatter tests). No regressions from this change.
- End-to-end stop hook tests in `tests/runtime.test.mjs` (which set up a git repo + a committed change) continue to pass — primary workspace review path is unchanged.

## Out of scope

- Recursive scanning (depth > 1) — kept fixed at immediate children for predictability.
- Cross-platform path quirks (e.g. case-insensitive FS) — uses existing `fs.realpath` resolution already in `state.mjs`.
- User-level defaults — that's #213, complementary, not duplicated here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -221,6 +221,19 @@ When the review gate is enabled, the plugin uses a `Stop` hook to run a targeted
 > [!WARNING]
 > The review gate can create a long-running Claude/Codex loop and may drain usage limits quickly. Only enable it when you plan to actively monitor the session.
 
+#### Monorepo mode (multiple sibling git repos)
+
+By default the review gate only reviews the workspace identified by `git rev-parse --show-toplevel` from the session's cwd. In a layout where several independent git repositories live under one parent directory (a common pattern for monorepos managed with submodules-like checkouts, or simply by convention), changes Claude makes in a *sibling* repo are never seen by the gate.
+
+Enable monorepo mode on the primary repo to make the `Stop` hook also scan the parent directory for sibling git repos and run a review in each one that has the gate enabled and has uncommitted changes:
+
+```bash
+/codex:setup --enable-monorepo
+/codex:setup --disable-monorepo
+```
+
+Setup must still be run separately in each sibling repo to enable the gate for it (`/codex:setup --enable-review-gate`). The hook respects each sibling's own state — a repo whose gate is off is skipped.
+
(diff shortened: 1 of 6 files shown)


PAzter1101:
Gentle ping on this one — it's been open ~7 weeks with no review yet, and it looks like the PR CI hasn't been triggered (fork PRs need a maintainer to approve the workflow run).

@dkundel-openai could you take a look, or point me to whoever owns the review-gate area? Happy to rebase, split it up, or add more test coverage for the sibling-discovery path if that helps get it into review. It's fully opt-in and backwards-compatible, so it should be low-risk to land.

Thanks!

PR #357, opened 2026-06-03
Add managed image generation (/codex:imagegen)

### LINKED ISSUE #356: task/rescue hangs at "starting" for image-generation prompts
The `task` runtime behind `/codex:rescue` and `codex-companion.mjs task` drives a `codex app-server` turn. When the prompt asks Codex to generate an image, the job stays in the `starting` phase and never finishes. I ran one with `--write` for over seven minutes: no file, no error, and `/codex:status` reported `running` / `starting` the whole time. The same prompt through `codex exec` writes the PNG in under a minute.

## Repro

1. `node scripts/codex-companion.mjs task --write "Use your built-in image generation to create a 512x512 cream square with a small icon centered. Save the PNG to /tmp/codex-img-test.png. Do not write code."` (or the matching `/codex:rescue`).
2. `node scripts/codex-companion.mjs status <job-id>` keeps returning `running`, phase `starting`, for as long as you wait. `cancel` is the only way out.
3. For comparison, `codex exec --skip-git-repo-check --sandbox workspace-write -C /tmp "Use your built-in image generation ... Save the PNG to /tmp/codex-img-test2.png. Do not write code." < /dev/null` saves the PNG in about 30 to 60 seconds.

Versions: codex-plugin-cc 1.0.4, codex-cli 0.136.0, Node v22.16.0, macOS, ChatGPT subscription auth.

## What looks like the cause

`lib/codex.mjs` sends a text-only `UserInput` (`[{ type: "text", text: prompt }]`) into a `codex app-server` turn. `codex app-server` is marked experimental, and the built-in image generation that `codex exec` exposes does not appear to be reachable inside the app-server turn. A prompt that asks for an image then has no tool to call, so the turn never progresses instead of erroring out.

## What I'd like to see

1. A turn that cannot proceed should time out or return a clear error rather than sitting in `starting` with no output. As it stands the job hangs silently and the only trace is candidate files left in `~/.codex/generated_images/`.
2. A managed path for image generation. Today the only working route is raw `codex exec`. Running a few of those at once contends for the Codex websocket and returns 403/429 while still spending quota, which is the exact failure the companion's serialized job runtime is built to avoid. Routing image-style tasks through `codex exec` under the same job tracking, or enabling whatever app-server feature backs imagegen, would cover it.

I'm willing to do the work and send a PR for either part, the graceful-failure fix or wiring image tasks through `codex exec` behind a flag. I'd rather hear which approach you'd accept before I start, so I build the right one.

lifeodyssey:
Closes #356.

## Why

The companion runs code tasks through one serialized codex app-server connection. Image generation never had a managed path. The only working route was a bare parallel Codex CLI, which contends for the same websocket and returns 403/429 while still spending quota. Sending an image prompt through the app-server task runtime instead just sat in the "starting" phase, which is the hang reported in #356.

## What it actually was

Not a missing capability. `modelProvider/capabilities/read` reports `imageGeneration: true`, `image_generation` is a stable default-on feature, and a turn does emit an `imageGeneration` item carrying the PNG (base64 `result` plus a `savedPath`). The image arrives early. The reason the turn looks hung is that after producing the image the model often keeps going, running shell commands to verify or convert the file, so the turn runs for minutes while the companion waits for `turn/completed`.

## The change

A managed image path that captures the `imageGeneration` item the moment it arrives, writes the bytes to `--out`, then interrupts the turn so the post-image tail never runs. It reuses the single serialized app-server connection, so image jobs never run concurrently, which is the contention the bare-CLI route caused. End to end against a real account this returns in about a minute instead of hanging.

- `/codex:imagegen` and an `imagegen` companion subcommand
- `--out`, `--image ref[,ref...]` (reference images for editing), `--background`, `--force`, `--model`
- a read-only thread: the managed path decodes and writes the bytes itself, so the model needs no workspace write access
- a startup timeout (`CODEX_PLUGIN_IMAGE_TIMEOUT_MS`, default 180s) and fast failure on an empty turn or an error notification

## Tests

`node --test`. The fake codex grows an `imagegen` behavior; coverage includes the happy path (asserting the written bytes equal the decoded base64 `result`, not a copy of `savedPath`), the `savedPath` fallback, no-image, error, timeout, reference inputs, the `--json` payload, and overwrite refusal.

Happy to adjust the command surface or split this up if you would prefer a different shape.

### DIFF
--- a/plugins/codex/.claude-plugin/plugin.json
+++ b/plugins/codex/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "codex",
-  "version": "1.0.4",
+  "version": "1.0.5",
   "description": "Use Codex from Claude Code to review code or delegate tasks.",
   "author": {
     "name": "OpenAI"
--- a/plugins/codex/CHANGELOG.md
+++ b/plugins/codex/CHANGELOG.md
@@ -1,5 +1,16 @@
 # Changelog
 
+## 1.0.5
+
+- Add managed image generation: `/codex:imagegen` plus an `imagegen` companion
+  subcommand, run over the same single serialized app-server connection the
+  plugin already uses for code tasks. Flags: `--out`, `--image <ref[,ref...]>`
+  (edit from reference images), `--background`, `--force`, `--model`. The path
+  captures the `imageGeneration` item the moment it arrives, writes the bytes to
+  `--out`, then interrupts the turn so the model's post-image shell tail never
+  runs. Reusing the one serialized connection also avoids the 403/429 contention
+  that bare parallel Codex CLI calls caused. Fixes #356.
+
 ## 1.0.0
 
 - Initial version of the Codex plugin for Claude Code
--- a/plugins/codex/commands/imagegen.md
+++ b/plugins/codex/commands/imagegen.md
@@ -1,0 +1,30 @@
+---
+description: Generate an image with Codex (managed and serialized) and save it to a file
+argument-hint: '[--out <path>] [--force] [--image <ref[,ref...]>] [--background] [--model <model>] [what the image should be]'
+allowed-tools: Bash(node:*)
+---
+
+Generate an image through the Codex companion's managed, serialized image path and return its output verbatim.
+
+Raw user request:
+$ARGUMENTS
+
+Run one `Bash` call:
+
+```bash
+node "${CLAUDE_PLUGIN_ROOT}/scripts/codex-companion.mjs" imagegen "$ARGUMENTS"
+```
+
+How it works:
+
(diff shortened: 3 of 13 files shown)


lifeodyssey:
Thanks for the review — both P2s are addressed in 4d64c02:

- **Propagate image write failures** (`lib/codex.mjs`): the notification handler's blanket `catch` now calls `settleErr(...)` with the real error instead of swallowing it, so a failed `--out` write rejects immediately rather than hanging until the 180s timeout while reporting "produced no image". Added a test that points `--out` at an unwritable path and asserts it fails via the filesystem error, not the timeout backstop.
- **Avoid blocking coding prompts that mention imagegen** (`imagegen-route-hook.mjs`): the bare-CLI block now matches only strong "generate an image" phrases, so coding tasks that merely mention `imagegen`/`gpt-image` (e.g. a file or test name) are no longer rejected. Added a regression test.

@codex review

PR #373, opened 2026-06-14
fix: reconcile dead-pid jobs so status/result stop reporting a zombie 'running' state

njrun1804:
## Problem

A `task`/`review` worker that dies before recording a terminal state — crash, SIGKILL, OOM, broker-socket drop, or a SessionEnd teardown in another session — leaves its job stuck `status: "running"` forever. `status` then shows a **growing `elapsed` over a frozen `updatedAt`** (elapsed is computed live from `startedAt` while the stored status never advances), and `result` throws *"still running"* / *"No job found"*. The status output actively misreports liveness, because `buildStatusSnapshot`/`resolveResultJob` partition jobs purely on the stored `status` string and never probe the worker pid.

## Fix

- **`lib/job-control.mjs`** — every job-read path (`buildStatusSnapshot`, `buildSingleJobSnapshot`, `resolveResultJob`, `resolveCancelableJob`) routes through a `reconcileJob`: for a `queued`/`running` job with a finite `pid`, `process.kill(pid, 0)` → `ESRCH` means the worker is gone, so the job is persisted as `failed:orphaned`. Idempotent (a reconciled job is terminal, so later reads skip it); a live pid (deliverable or `EPERM`) or the current process's own pid is never touched.
- **`lib/tracked-jobs.mjs`** — `runTrackedJob` installs `exit`/`SIGTERM`/`SIGINT`/`SIGHUP` finalizers that synchronously flush a terminal `failed:interrupted` record if the worker dies while still `running` (SIGKILL/OOM can't be trapped — those are caught by the reconcile above). Disarmed on normal completion. Defense in depth.
- **`lib/render.mjs`** — `renderStoredJobResult` falls back to the job's log tail when no result payload was captured, so an orphaned job still yields its partial output instead of an empty "No result" message.

## Tests

`tests/job-control.test.mjs` (new): dead-pid → `failed:orphaned`, live-pid untouched, `result` resolves a reconciled orphan. Full suite **86 → 89, all green** (`node --test tests/*.test.mjs`).


dkundel-openai:
Duplicate of #425.

PR #374, opened 2026-06-14
feat: add /codex:transfer for Claude Code session handoff

stefanstokic-oai:
## Description

Adds `/codex:transfer` to import the current Claude Code session into a persistent Codex thread and print the corresponding `codex resume <thread-id>` command.

The command reuses Codex's native external-agent session importer so transferred conversations follow the same conversion rules and produce the same visible turn history as Codex App imports. The existing `SessionStart` hook captures Claude's transcript path automatically, with `--source <claude-jsonl>` available as a manual override.

Includes compatibility handling for unsupported Codex versions, clear import-failure errors, documentation, and test coverage. Validated with `npm test` (90 tests) and `npm run build`.

Issue #28, opened 2026-03-31
Buffered  notifications can drop subagent names and log raw thread IDs


VOIDXAI:
## Bug

When a task spawns a subagent early enough, the job log can lose the subagent name and fall back to raw thread IDs like `thr_2`.

## Reproduction

1. Run a task that quickly spawns a subagent / collaboration tool turn
2. Inspect the resulting job log
3. Observe entries like:
   - `Starting subagent thr_2 via collaboration tool: wait.`
   - `Subagent thr_2 reasoning: ...`

Instead of the expected nickname / label (for example `design-challenger`).

## Root Cause

`captureTurn()` buffers notifications that arrive before the main turn ID is known. During replay, buffered notifications are filtered through `belongsToTurn(...)`.

That works for turn-scoped items, but it is too strict for thread lifecycle notifications like:

- `thread/started`
- `thread/name/updated`

If one of those arrives early for a subagent thread, it can be dropped before label registration happens. Later log rendering then has no thread label and falls back to the raw thread ID.

## Impact

- Logs and progress output become harder to understand.
- Subagent labeling is unstable and depends on event timing.
- Any future behavior that depends on stored thread labels can inherit the same race.

## Suggested Fix

Replay buffered thread lifecycle notifications unconditionally before applying turn-scoped filtering to the remaining buffered messages.

PR #376, opened 2026-06-16
Bound the Codex turn await: fix unbounded hang on stalled/dead app-server + make turn budget configurable

russjhammond:
## Problem

A foreground Codex task can hang until the caller's external process timeout kills it with no output. Two distinct defects in `scripts/lib/codex.mjs`:

**1. Unbounded, un-rejectable completion await.** `captureTurn` ends in `return await state.completion` — a Promise resolved only by `completeTurn()` (on a `turn/completed` notification, or a 250ms inferred fallback). Its `rejectCompletion` counterpart is wired into the turn-capture state but has **zero call sites** — dead code. So once `startRequest()` resolves with `inProgress`, if the app-server dies or simply stalls, *nothing* rejects the await. It hangs until the caller (e.g. a host harness with a ~2-minute shell ceiling) SIGKILLs the process — surfacing as "hangs every time, returns nothing."

**2. Turn budget read at import time.** `const TURN_TIMEOUT_MS = Number(process.env.CODEX_TURN_TIMEOUT_MS) || 600000` is evaluated when the module is imported. The companion sets the env var *after* import (e.g. a foreground budget below the caller's ceiling, or via a flag), so the override never takes effect — the budget is frozen at the default and effectively inert.

## Fix

- Wire `rejectCompletion` to `client.exitPromise` so a mid-turn process exit rejects the await immediately with a structured error instead of hanging.
- Race `state.completion` against a hard deadline timer (cleared in a `finally`).
- Resolve the budget at **call time** — `resolveTurnTimeoutMs(options)`: explicit option → `CODEX_TURN_TIMEOUT_MS` env → default — so runtime overrides actually apply.
- `codex-companion.mjs`: set a foreground turn budget just below the caller's external ceiling and add a `--turn-timeout-ms` flag. Background/detached jobs inherit the full default.

## Verification

- `node --check` passes on both files.
- `task --turn-timeout-ms 3000` on a long-generation prompt now aborts at ~3s with `codex turn exceeded the 3000ms turn budget` (previously ran unbounded — the flag was inert).
- A normal foreground turn still completes cleanly (exit 0).

### DIFF
--- a/plugins/codex/scripts/codex-companion.mjs
+++ b/plugins/codex/scripts/codex-companion.mjs
@@ -719,9 +719,25 @@ function enqueueBackgroundTask(cwd, job, request) {
   };
 }
 
+// Set the per-turn budget for a FOREGROUND command (codex.mjs reads
+// CODEX_TURN_TIMEOUT_MS at call time). Precedence: explicit --turn-timeout-ms
+// flag > a pre-set CODEX_TURN_TIMEOUT_MS env (e.g. settings.json) > the
+// foreground default just under the host Bash ceiling, so a stalled turn
+// returns a structured timeout instead of being SIGKILLed with no output.
+// Only call this on a foreground path: a detached background worker inherits
+// the parent env, so capping it here would shrink the background budget too.
+function applyForegroundTurnBudget(options) {
+  const explicit = Number(options["turn-timeout-ms"]);
+  if (Number.isFinite(explicit) && explicit > 0) {
+    process.env.CODEX_TURN_TIMEOUT_MS = String(explicit);
+  } else if (!process.env.CODEX_TURN_TIMEOUT_MS) {
+    process.env.CODEX_TURN_TIMEOUT_MS = String(FOREGROUND_TURN_TIMEOUT_MS);
+  }
+}
+
 async function handleReviewCommand(argv, config) {
   const { options, positionals } = parseCommandInput(argv, {
-    valueOptions: ["base", "scope", "model", "cwd"],
+    valueOptions: ["base", "scope", "model", "cwd", "turn-timeout-ms"],
     booleanOptions: ["json", "background", "wait"],
     aliasMap: {
       m: "model"
@@ -746,6 +762,12 @@ async function handleReviewCommand(argv, config) {
     jobClass: "review",
     summary: metadata.summary
   });
+  // Review turns run foreground (--wait) through the same path as tasks; give
+  // them the same foreground budget so a stall returns a structured timeout
+  // instead of hitting the host Bash ceiling with an empty result.
+  if (!options.background) {
+    applyForegroundTurnBudget(options);
+  }
   await runForegroundCommand(
     job,
     (progress) =>
@@ -814,18 +836,10 @@ async function handleTask(argv) {
     return;
   }
 
-  // Foreground turn budget. codex.mjs reads CODEX_TURN_TIMEOUT_MS. Precedence:
-  // explicit --turn-timeout-ms flag, then a pre-set CODEX_TURN_TIMEOUT_MS env
-  // (e.g. from settings.json), else the foreground default just under the Bash
-  // ceiling so a stall returns a structured timeout instead of a SIGKILL.
-  // This runs only on the foreground path; the background branch returned above
-  // and its detached worker inherits the parent env unchanged (full budget).
-  const explicitTurnTimeout = Number(options["turn-timeout-ms"]);
-  if (Number.isFinite(explicitTurnTimeout) && explicitTurnTimeout > 0) {
-    process.env.CODEX_TURN_TIMEOUT_MS = String(explicitTurnTimeout);
-  } else if (!process.env.CODEX_TURN_TIMEOUT_MS) {
-    process.env.CODEX_TURN_TIMEOUT_MS = String(FOREGROUND_TURN_TIMEOUT_MS);
-  }
+  // Foreground turn budget (see applyForegroundTurnBudget). Runs only on the
+  // foreground path; the background branch returned above and its detached
+  // worker inherits the parent env unchanged (full default budget).
+  applyForegroundTurnBudget(options);
 
   const job = buildTaskJob(workspaceRoot, taskMetadata, write);
   await runForegroundCommand(
--- a/plugins/codex/scripts/codex-companion.mjs
+++ b/plugins/codex/scripts/codex-companion.mjs
@@ -68,6 +68,16 @@ const ROOT_DIR = path.resolve(fileURLToPath(new URL("..", import.meta.url)));
 const REVIEW_SCHEMA = path.join(ROOT_DIR, "schemas", "review-output.schema.json");
 const DEFAULT_STATUS_WAIT_TIMEOUT_MS = 240000;
 const DEFAULT_STATUS_POLL_INTERVAL_MS = 2000;
+// Full per-turn budget for background/detached jobs (no external Bash ceiling
+// to collide with). codex.mjs reads CODEX_TURN_TIMEOUT_MS; this is the value we
+// thread down for background runs when the user hasn't overridden it.
+const DEFAULT_TURN_TIMEOUT_MS = 600000;
+// Foreground runs are invoked by Claude Code's Bash tool, which SIGKILLs node
+// at its own timeout (default 120000ms) and returns nothing. Set the runtime
+// turn budget just below that so the foreground turn fails fast with a
+// structured "turn timed out — re-run with --background" message that carries
+// any partial output, instead of being killed with an empty result.
+const FOREGROUND_TURN_TIMEOUT_MS = 110000;
 const VALID_REASONING_EFFORTS = new Set(["none", "minimal", "low", "medium", "high", "xhigh"]);
 const MODEL_ALIASES = new Map([["spark", "gpt-5.3-codex-spark"]]);
 const STOP_REVIEW_TASK_MARKER = "Run a stop-gate review of the previous Claude turn.";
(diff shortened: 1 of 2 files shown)


russjhammond:
Thanks for the review — both P2s addressed in `688192a`:

1. **Unobserved completion rejection** — moved the `client.exitPromise → rejectCompletion` wiring to *after* `startRequest()` resolves. An early `startRequest()` rejection (broker-busy, app-server exit while the request is pending) now propagates directly and `state.completion` is never observed, so it can't surface as an unhandled rejection. Mid-turn-death coverage is unchanged (handler is in place before the completion is awaited).

2. **Review turns** — extracted `applyForegroundTurnBudget(options)` and now call it from both `handleTask` and `handleReviewCommand`, so `--wait` review / adversarial-review turns get the foreground budget instead of falling back to 600000ms. Also accept `--turn-timeout-ms` on the review command; guarded with `!options.background` so detached workers keep the full default.

Verified: `node --check` on both files; `--turn-timeout-ms 3000` aborts a long turn at ~3s with the structured error; a normal turn still completes (exit 0).

russjhammond:
Fair point — a scoping note on this PR vs. a follow-up.

The primary goal here is eliminating the unbounded **hang**: a stalled or dead app-server previously blocked on `await state.completion` until the host's external timeout SIGKILLed the process with no signal. After this change a timed-out turn rejects at the budget and exits non-zero with `codex turn exceeded the <n>ms turn budget` on stderr — already a strict improvement over a silent kill.

Rendering the timeout (plus any partial turn output) to **stdout** as a structured result is a reasonable next step, but it touches the foreground render path (`runForegroundCommand` / result rendering) and the right shape depends on this repo's output conventions, which you'd know better than a drive-by patch would. Happy to extend this PR (or open a follow-up) to render a structured timeout result if that's the direction you'd prefer — I just didn't want to guess at the rendering contract here.

axisrow:
@russjhammond — thank you for this PR. I forked openai/codex-plugin-cc and ported your fix into my fork (axisrow/codex-plugin-cc#26), so this comment is sharing back what I found while adapting it.

Your diagnosis was spot-on — `rejectCompletion` as dead code with zero call sites was the exact root cause. The port went through two adversarial review cycles (Codex + Claude subagent) and your code held up well. Three things came out of the review that I think are worth sharing:

1. **The foreground/background env-inheritance hazard is real and worth knowing about.** Your `applyForegroundTurnBudget` mutates `process.env.CODEX_TURN_TIMEOUT_MS`, and the detached background worker inherits `env: process.env`. In my fork this was safe because `handleTask` returns early on the background branch before the call — but it's a fragile invariant. If anyone later moves the background-return or reorders the handler tail, the bare call would leak the 110s foreground budget into background workers. A cleaner path that eliminates this entirely: thread `turnTimeoutMs` through the options object that already flows from `handleTask` → `executeTaskRun` → `runAppServerTurn` → `captureTurn`. Your `resolveTurnTimeoutMs` already checks `options.turnTimeoutMs` first — the precedence chain is ready, the callers just never feed it. The env read stays as a fallback for operator-configured overrides.

2. **`exitRaceSettled` is redundant.** I dropped it in a cleanup pass. JS promises silently ignore any settle after the first, so `rejectCompletion` on an already-settled `state.completion` is already a no-op. `state.completed` is the load-bearing guard (it protects against a late-arriving `exitPromise.then` after normal completion). `exitRaceSettled` was the weaker half — same protection the settled-promise semantics give for free.

3. **Three Codex findings that are real but out of scope for this PR** (follow-up improvements, not bugs in your fix):
   - A timed-out broker turn keeps executing after the job is reported failed (the deadline rejects the await but doesn't send `turn/interrupt`).
   - The deadline is armed only after `turn/start` responds — a stalled `turn/start` request itself remains unbounded.
   - `--turn-timeout-ms` is silently ignored for background runs.

My fork PR with attribution to this PR: axisrow/codex-plugin-cc#26. Happy to talk through any of it — and equally happy to leave it here.

axisrow:
Follow-up on my earlier comment — the three items I flagged as "out of scope for this PR" are now closed in my fork, in case the notes are useful to anyone picking them up later.

Following the scoping Russ drew in [his comment above](https://github.com/openai/codex-plugin-cc/pull/376#issuecomment-4714055886) — this PR's job is eliminating the unbounded *hang*, and the three below were consciously deferred — I split them into a separate follow-up PR (axisrow/codex-plugin-cc#28) so #376's scope stays clean. No ask that you take any of it.

1. **Interrupt the turn on deadline.** When the deadline rejects the await, the broker keeps executing a write-capable turn after the job is already reported failed. Fix: wrap the `Promise.race` in try/catch; on the budget-error, best-effort `interruptAppServerTurn(cwd, { threadId, turnId })` (bounded 2s await, swallow failures) before re-throwing. Needs `cwd` threaded into `captureTurn` options.

2. **Arm the deadline before `turn/start`.** Your `exitPromise` wiring runs only after `startRequest()` resolves, so a stalled `turn/start` (app-server alive, never responds) stays unbounded. Fix: hoist the deadline + the `exitPromise → rejectCompletion` wiring above `startRequest()`, and race the entire lifecycle (`startRequest` + completion) against the one deadline. Trade-off: if the deadline fires during `startRequest` there's no `turnId` to interrupt yet — but the process no longer hangs.

3. **`--turn-timeout-ms` for background runs.** I replaced the `process.env` mutation (`applyForegroundTurnBudget`, foreground-only, gated by `if (!options.background)`) with plain options threading — `resolveTurnTimeoutMsFromOptions(options)` resolves flag → env → default and flows through `buildTaskRequest → handleTaskWorker → executeTaskRun → runAppServerTurn → captureTurn`. Background gets the full 600s default (no external ceiling to collide with). The env read stays as a fallback for operator-configured overrides. This also retires the env-inheritance hazard I worried about in point 1 of my first comment.

Verified TDD-first: a `stalled-turn-start` fake-codex behavior (never answers `turn/start`) timed out via `--turn-timeout-ms 3000` instead of hanging. 162/166 tests green (4 pre-existing failures unchanged).

Fork PR with attribution back to this one: axisrow/codex-plugin-cc#28. Pure sharing — take it, adapt it, or set it aside.

russjhammond:
@axisrow — this is the good kind of open-source citizenship, and I appreciate it. Forking, porting through two adversarial review cycles, and then bringing the findings *back here* instead of quietly keeping them in your fork is exactly how this is supposed to work.

Your three points all land:

1. **Options-threading over env mutation** — cleaner than my `process.env.CODEX_TURN_TIMEOUT_MS` write. `resolveTurnTimeoutMs` already checks `options.turnTimeoutMs` first, so the precedence chain was ready and the callers just never fed it. Threading it retires the background env-inheritance hazard entirely.
2. **`exitRaceSettled` is redundant** — agreed. `state.completed` is the load-bearing guard, and settled-promise semantics make a second reject a no-op.
3. **The three deferred items** (interrupt-on-deadline, arming the deadline *before* `turn/start`, and `--turn-timeout-ms` for background runs) are real and well-scoped as a follow-up.

I've picked this PR back up: rebased onto current `main` (trivial overlap with the #374 transfer-command constants, resolved by keeping both) and I'm folding your findings in with attribution to axisrow/codex-plugin-cc#26 and #28. Thanks again for the rigor. 🙏

axisrow:
Awesome — really glad this one's moving again, and thanks for the attribution. Happy it was useful.

axisrow:
@russjhammond — following up on your "folding your findings in with attribution" note, with one correction to my own earlier point and one larger suggestion.

**First, a correction: point 1 of my first comment doesn't describe a bug in this PR.** I said the env mutation was "safe in my fork because `handleTask` returns early on the background branch" and framed it as a fragile invariant. Re-reading your diff at the current head, `applyForegroundTurnBudget(options)` is called from **both** `handleTask` and `handleReviewCommand`, and `resolveTurnTimeoutMs` reads `process.env.CODEX_TURN_TIMEOUT_MS` globally — so both branches are covered, implicitly but genuinely. There's no asymmetry here.

The asymmetry I later hit was a side effect of **my** move to options-threading (fork #28): once the value travels through an options object instead of the environment, every call site has to be fed explicitly, and I missed one — the adversarial-review path dropped `turnTimeoutMs` and silently fell back to the 600s default. That's a bug I introduced by taking your env write out, not one you shipped. Fixed in fork axisrow/codex-plugin-cc#44. If you adopt options-threading, that call site is the one to check; if you keep the env write, the issue doesn't arise. Sorry for the mis-attribution.

**The larger suggestion: the wall-clock budget is measuring the wrong thing.**

A fixed per-turn budget conflates two very different states — "the app-server is dead" and "the agent is doing slow, real work". `gpt-5.6-sol` at `effort=xhigh` on a large adversarial review routinely runs well past 10 minutes while streaming events the whole time. It isn't stalled; it's slow. That forces an impossible tuning choice: set the budget low enough to catch a dead app-server quickly and you kill legitimate long reviews mid-flight; set it high enough to tolerate them and a genuinely wedged app-server hangs just as long before surfacing.

The metric that separates the two cleanly is **idle time**. Concretely:

- Replace the one-shot deadline `setTimeout` with a resettable closure stored on the turn-capture state.
- `applyTurnNotification` is the single choke point every turn notification already flows through — reset the deadline there on `turn/started`, `item/started`, `item/completed`, **and** on in-item progress/delta notifications. That last part matters: a single long-running command execution or MCP tool call emits no item boundary for minutes, so without delta resets it gets killed despite being demonstrably alive.
- Keep a much larger wall-clock ceiling as a second, coarse failsafe against a turn that resets its own timer forever (runaway tool-call loop).

A turn that keeps producing events can then run as long as it needs; only genuine silence trips the timeout. Your actual fix here — making the completion `await` reject-able at all (`rejectCompletion` was dead code) and resolving the budget at call time — is orthogonal and fully preserved; this only changes what the resulting timer *measures*. Implemented in fork axisrow/codex-plugin-cc#41.

Two connections worth knowing about:

- **#407** adds an opt-in structured-output retry loop (up to `structuredRetryAttempts`, default 3, additional `captureTurn` calls). Under a wall-clock budget each retry gets a *fresh* full budget, so the worst case is ~4x the configured timeout while every individual turn stays formally within its deadline. Under an idle budget that multiplication doesn't happen — total time is bounded by real work, not by attempts × worst-case budget.
- **#375** rewrites the same `handleReviewCommand` options object that this PR and #407 both touch, so some conflict is likely whichever lands first.

As always: take it, adapt it, or leave it. The fork is there if the diffs are useful.

Issue #63, opened 2026-03-31
`codex login status` panics on macOS inside Claude Code sandbox (SCDynamicStore NULL)


msadikyildiz:
## Description

On macOS, running `codex login status` inside Claude Code's sandbox causes a Rust panic in the `system-configuration` crate, which makes `getCodexLoginStatus()` report `loggedIn: false` even when the user has valid tokens in `~/.codex/auth.json`.

## Error

```
WARNING: proceeding, even though we could not update PATH: Operation not permitted (os error 1)

thread 'main' panicked at system-configuration-0.6.1/src/dynamic_store.rs:154:1:
Attempted to create a NULL object.
```

## Root cause

The Codex CLI binary accesses macOS's `SCDynamicStore` API on every invocation to update the system PATH. When Claude Code's sandbox blocks this access, `SCDynamicStoreCreate` returns NULL and the Rust wrapper panics instead of handling the error gracefully.

- `codex --version` shows the warning but still succeeds (the panic is non-fatal for this command)
- `codex login status` triggers a full panic and exits non-zero
- `codex whoami` also panics

Running the same commands outside the sandbox works correctly:
```
$ codex login status
Logged in using ChatGPT
```

## Impact

- `/codex:setup` reports `loggedIn: false` despite valid authentication
- `ensureCodexReady()` throws "Codex CLI is not authenticated" before any review or task
- All `/codex:review` and `/codex:task` commands are blocked

## Workaround

Read `~/.codex/auth.json` directly in `getCodexLoginStatus()` before falling back to the binary call. The file contains the same tokens that `codex login status` would check.

## Environment

- macOS (Darwin 25.3.0, Apple Silicon)
- Codex CLI: 0.117.0
- codex-plugin-cc: 1.0.1
- Claude Code sandbox enabled

## Related

- #21 (`codex login status` false-negative with Azure provider)
- #58 (same symptom with Azure OpenAI)
- #41 (401 after fresh login)
- #18 (sandbox restrictions on Linux/bwrap)

PR #379, opened 2026-06-17
fix: reconcile orphaned jobs whose launcher process has died

mroldrobot:
## Problem

A tracked job's completion is written by the launcher process that owns it
(`tracked-jobs.mjs::runTrackedJob`): it marks the job `running`, `await`s the
turn, then writes `completed`/`failed`. If that launcher dies before the turn
finishes — a cancelled background job, an ended Claude Code session, a crash, or
the machine sleeping — the completion write never runs and the job is frozen at
`running`/`queued` with a now-dead pid.

Nothing reconciles this: `buildStatusSnapshot` / `buildSingleJobSnapshot` filter
purely on the stored `status` string, and the only `ESRCH` handling in the
codebase is in the cancel path. So `/codex:status` reports the job `running`
forever and `/codex:result` refuses with "still running". In practice these
orphans accumulate (I had several stuck for days).

## Fix

`listJobs()` — which every status/result/cancel path flows through — now probes
the recorded launcher pid with `process.kill(pid, 0)`:

- `ESRCH` (process gone) → reconcile the job to `failed`, clear the pid, set
  `completedAt` and an explanatory `errorMessage`. The correction is persisted to
  both the state index and the per-job file.
- `EPERM` / unknown → treat as alive (fail safe).
- Jobs without a pid, or already in a terminal state, are left untouched.

A dead background turn now surfaces as `failed` (and is retryable) instead of
hanging `running` indefinitely.

## Tests

Adds coverage in `tests/state.test.mjs`: dead-pid running/queued reconciliation,
persistence across reads, and live-pid / pidless / finished jobs left untouched.
`node --test tests/state.test.mjs` → 7/7 green.

Note: 3 unrelated tests (#63/#65/#67) fail under the aggregate `npm test` due to
pre-existing cross-file state leakage; they pass in isolation and fail
identically on `main` without this change.

### DIFF
--- a/plugins/codex/scripts/lib/state.mjs
+++ b/plugins/codex/scripts/lib/state.mjs
@@ -146,8 +146,85 @@ export function upsertJob(cwd, jobPatch) {
   });
 }
 
+function isProcessAlive(pid) {
+  if (!Number.isInteger(pid) || pid <= 0) {
+    return true; // unknown pid -> never treat as dead
+  }
+  try {
+    process.kill(pid, 0);
+    return true;
+  } catch (error) {
+    if (error?.code === "ESRCH") {
+      return false; // no such process
+    }
+    return true; // EPERM (process exists, other owner) or anything else -> assume alive
+  }
+}
+
+// A tracked job's completion is written by the launcher process that owns it
+// (see tracked-jobs.mjs::runTrackedJob). If that launcher dies before the turn
+// finishes (cancelled background job, ended session, crash, sleep), the job is
+// frozen at "running"/"queued" with a stale pid forever. Reconcile such orphans
+// to "failed" at read time so /status, /result and /cancel reflect reality.
+function reconcileJobLiveness(jobs) {
+  const changedIds = [];
+  const now = nowIso();
+  const reconciled = jobs.map((job) => {
+    if (job.status !== "running" && job.status !== "queued") {
+      return job;
+    }
+    if (job.pid == null || isProcessAlive(job.pid)) {
+      return job;
+    }
+    changedIds.push(job.id);
+    return {
+      ...job,
+      status: "failed",
+      phase: "failed",
+      pid: null,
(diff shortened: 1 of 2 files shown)


xoonjaeho:
I applied this locally on top of `db52e28` (clean apply) and it fixes the zombie jobs on my host — `listJobs` reconciles, and `result` starts answering for a job whose launcher died instead of refusing forever.

One gap worth closing: **the fail-closed liveness default is the best thing about this PR and nothing tests it.**

```js
if (error?.code === "ESRCH") {
  return false; // no such process
}
return true; // EPERM (process exists, other owner) or anything else -> assume alive
```

That last `return true` is what makes this safer than the alternative in #225, which defaults to *dead* on any unexpected errno. Wrongly marking a live job as `failed` is much worse than leaving a zombie, since a codex turn routinely sits silent for many minutes.

But flipping it to `return false` passes all four tests here, and the rest of the suite. A test that catches it:

```js
test("listJobs keeps a running job when the liveness probe fails for a reason other than ESRCH", () => {
  const workspace = makeTempDir();
  upsertJob(workspace, { id: "eperm", status: "running", pid: DEAD_PID, workspaceRoot: workspace });

  const originalKill = process.kill;
  process.kill = () => {
    throw Object.assign(new Error("operation not permitted"), { code: "EPERM" });
  };
  let jobs;
  try {
    jobs = listJobs(workspace);
  } finally {
    process.kill = originalKill;
  }

  assert.equal(jobs.find((job) => job.id === "eperm").status, "running");
});
```

Mutation-checked both ways: it fails with the default flipped and passes with it restored. Feel free to take it as-is.

Environment: Windows 11, node v24.14.0. Suite went 91/79 pass → 96/84 pass with this PR plus that test, no new failures (the 12 pre-existing failures here are Windows-only — Unix-socket and temp-dir assertions that CI never sees).

dkundel-openai:
Duplicate of #425.

PR #383, opened 2026-06-18
[codex] fix hooks config schema

bjulian5:
## Summary

Removes the unsupported top-level `description` field from `plugins/codex/hooks/hooks.json`.

Codex's plugin hook config parser expects the top-level object to contain only `hooks`, so the released plugin currently fails to parse with:

`unknown field description, expected hooks`

The descriptive metadata already lives in the plugin manifest; the hook config only needs the hook declarations.

## Validation

- `jq 'keys' plugins/codex/hooks/hooks.json` returns only `["hooks"]`
- `npm test` passes (`86/86`)

### DIFF
--- a/plugins/codex/hooks/hooks.json
+++ b/plugins/codex/hooks/hooks.json
@@ -1,5 +1,4 @@
 {
-  "description": "Optional stop-time review gate for Codex Companion.",
   "hooks": {
     "SessionStart": [
       {


bjulian5:
@dkundel-openai This PR is ready from my side: it removes the unsupported top-level `description` key that makes the released plugin hook config fail to parse. Local validation: `jq` shows only the top-level `hooks` key, and `npm test` passes (`86/86`). GitHub reports the PR as mergeable, but I only have read permission on the upstream repo, so I cannot merge or enable auto-merge.

Wintersta7e:
Confirming the same fix independently. Worth noting the blast radius: the parser rejects the file outright, so this affects every hook the plugin registers, not only the stop gate.

One suggestion — the key has been there since the initial commit, and nothing currently stops it coming back. A one-line assertion in the existing `hooks keep session-end cleanup and stop gating enabled` test pins the shape:

```js
assert.deepEqual(Object.keys(JSON.parse(source)), ["hooks"]);
```

Happy for you to take that into this PR.

dkundel-openai:
Duplicate of #406.

PR #389, opened 2026-06-21
fix: align app-server capabilities and thread params with the current codex protocol

armanobosyan:
## Summary

- `npm run build` (the tsc check against the generated app-server types) currently fails on `main` with the latest codex CLI (0.141.0): `InitializeCapabilities` now requires `requestAttestation`, and `experimentalRawEvents` has been removed from `thread/start` params.
- Declares `requestAttestation: false` in the default capabilities and stops sending the removed `experimentalRawEvents` field, so the typecheck passes against the current protocol again.

## Testing

- `npm run build`
- `node --test tests/runtime.test.mjs`

### DIFF
--- a/plugins/codex/scripts/lib/app-server.mjs
+++ b/plugins/codex/scripts/lib/app-server.mjs
@@ -32,6 +32,7 @@ const DEFAULT_CLIENT_INFO = {
 /** @type {InitializeCapabilities} */
 const DEFAULT_CAPABILITIES = {
   experimentalApi: false,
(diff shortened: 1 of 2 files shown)


armanobosyan:
Closing as redundant: main (v1.0.5) already declares `requestAttestation` in the default capabilities and drops the removed `experimentalRawEvents` from thread/start params (via #374), so the typecheck passes. The stalled-turn fix that depended on this is in #390 (now rebased onto v1.0.5 and standalone).

Issue #222, opened 2026-04-14
No OS-level process liveness check — stale 'running' state after worker death (root cause for #164, #202)


bowenQT:
## Summary

When a Codex worker process dies unexpectedly (SIGKILL, OOM, broker socket disconnect, app-server crash), the companion job state file retains `status: "running"` indefinitely. **No code path in the companion checks whether the stored PID is actually alive.** This is the shared root cause behind #164 (review stuck in "running") and #202 (zombie job blocks subsequent tasks), plus a third unreported symptom: `result <job-id>` returning "No job found" while `status <job-id>` reports the full (stale) job object.

## Three Symptoms, One Root Cause

| Symptom | Affected command | Existing issue |
|---------|-----------------|----------------|
| `--wait` poll loop spins until timeout against a dead process | `status --wait` | #164 |
| "Task X is still running" blocks all subsequent task calls | `task` (via `resolveLatestTrackedTaskThread`) | #202 |
| `result <job-id>` says "No job found" while `status <job-id>` returns the job | `result` vs `status` | *(new)* |

### 1. Detached worker spawned with zero crash monitoring

`spawnDetachedTaskWorker()` in `codex-companion.mjs` (lines 613–624):

```javascript
function spawnDetachedTaskWorker(cwd, jobId) {
  const child = spawn(process.execPath, [...], {
    detached: true,      // own process group
    stdio: "ignore",     // parent never sees output
    windowsHide: true
  });
  child.unref();         // parent drops all references
  return child;
}
```

- No `child.on('exit', ...)` handler
- No `child.on('error', ...)` handler
- Parent has **no visibility** into worker lifecycle after spawn

### 2. `runTrackedJob()` catch block only handles JS errors

`tracked-jobs.mjs` (lines 142–204): The `try/catch` around `await runner()` handles JavaScript exceptions but **cannot fire** when the detached worker process is killed at the OS level (SIGKILL, OOM killer, etc.). The promise simply never settles, and the state file remains at `status: "running"`.

### 3. No PID liveness probe in any status/query path

The following functions read the state file and return `status` as-is, **without checking if `job.pid` is alive**:

| Function | File | Lines |
|----------|------|-------|
| `enrichJob()` | `job-control.mjs` | 161–181 |
| `buildSingleJobSnapshot()` | `job-control.mjs` | 242–254 |
| `waitForSingleJobSnapshot()` | `codex-companion.mjs` | 293–309 |
[...]

PR #392, opened 2026-06-21
fix: reap zombie 'running' jobs at the resume gate via pid-liveness check (#222, #202)

storyarcade:
### Problem
When a Codex task worker crashes or is killed externally, its job record stays `status: "running"` forever. The next `task`/resume call in that cwd then throws and is permanently blocked:

> `Task <id> is still running. Use /codex:status before continuing it.`

`resolveLatestTrackedTaskThread()` finds any `queued|running` job and throws **without checking whether the worker process is actually alive** — there is no pid-liveness check, so a dead worker's record is indelible. (Root cause: #222; symptoms: #202, #164.)

### Fix
Before the blocking throw, probe the worker pid:

1. `process.kill(pid, 0)` → `ESRCH` ⇒ dead, `EPERM` ⇒ alive-but-not-ours (treat as alive). Portable across macOS/Linux/Windows via libuv; no new dependency.
2. If the worker is **dead**, reconcile the job to `failed` — writing `state.json` (the store the gate reads) plus a best-effort per-job-file mirror for status/result readers — then mutate the in-memory snapshot and fall through to normal resolution.
3. Only throw the "still running" error if the worker is **actually alive** (live jobs are unaffected).

### Why minimal (vs #216)
This touches only the gate function: ~12 lines + two small helpers. It does **not** add a deadline/timeout, a background reaper, or a daemon — just "don't block on a provably-dead worker." Builds on the approach in the now-conflicting #216, scoped to the surgical, conflict-free subset.

### Known limitation (PID reuse)
A PID-based probe inherits the standard caveat: if the OS recycles the crashed worker's PID before the next `task` call and an unrelated process holds it, the probe briefly treats the zombie as alive. It self-corrects on a later call, and the dominant case (immediate crash) is handled correctly. A future hardening could store the job id in a pidfile and verify identity, not just existence — out of scope here.

### Test
- Spawn a task, `kill -9` the worker, run another task in the same cwd → previously throws forever; now the dead job is reconciled to `failed` and the new task proceeds.
- Live job unaffected: a running worker still correctly blocks with the original message.

`node --check` passes on the patched file.

### DIFF
--- a/plugins/codex/scripts/codex-companion.mjs
+++ b/plugins/codex/scripts/codex-companion.mjs
@@ -288,6 +288,41 @@ function isActiveJobStatus(status) {
   return status === "queued" || status === "running";
 }
 
+/** Portable liveness probe. ESRCH => dead; EPERM => alive (exists, not ours). */
+function isWorkerAlive(pid) {
+  if (!pid || pid <= 0) return false; // no pid recorded -> cannot be alive
+  try {
+    process.kill(pid, 0);
+    return true;
+  } catch (err) {
+    return Boolean(err && err.code === "EPERM"); // exists but not ours -> alive
+  }
+}
+
+/**
+ * Mark a dead-pid job failed: write state.json (what the gate reads) and mirror
+ * the per-job file so status/result readers stay consistent.
+ */
+function reconcileDeadJob(workspaceRoot, job) {
+  const patch = {
+    id: job.id,
+    status: "failed",
+    pid: null,
+    errorMessage: "Worker process not alive; reconciled by resume gate.",
+    completedAt: nowIso()
+  };
+  // upsertJob writes state.json, which is what the gate (and resolution) reads.
+  upsertJob(workspaceRoot, patch);
+  // Best-effort: keep the per-job file consistent so /codex:status and result
+  // readers don't show a stale "running". The state.json write above is the
+  // authoritative un-stick; a per-job write failure is non-fatal.
+  try {
+    writeJobFile(workspaceRoot, job.id, { ...job, ...patch });
+  } catch {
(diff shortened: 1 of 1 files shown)


ghbaud:
This is a clean, minimal fix for the blocking symptom in #202, and `process.kill(pid, 0)` is the right portable primitive for it.

One scope note, in case the intent is for this to also close the #222 root cause and not just #202: the reconcile happens at the resume gate via `resolveLatestTrackedTaskThread()`, which reads `state.json`. That covers every job that made it into `state.json`. It does **not** cover a worker that dies *before* it registers there — one that crashes after `writeJobFile()` but before `upsertJob()`. That job exists only as an orphaned per-job file with a stale `running`/`queued` status and a dead pid, and it never appears in `state.json` at all.

That variant doesn't trigger #202's blocking failure — the gate never finds the job, so nothing gets blocked — which is exactly why it's reasonable to leave it out of a #202-focused change. But it is still part of the #222 root cause: the dead job stays invisible to `/codex:status` and `/codex:result`, because both of those also start from `state.json`. So if this lands and #222 is closed along with it, that never-registered case would quietly remain.

I hit this exact variant on Windows 11 and wrote it up here: [#222 (comment)](https://github.com/openai/codex-plugin-cc/issues/222#issuecomment-4504061644). The patch I'm running locally adds a small sweep over the per-job files that finds entries missing from `state.json` with a dead pid and reconciles them to `failed`, run at the status/result read paths. It's complementary to this PR rather than competing, since it touches the read paths instead of the resume gate. Happy to open it as a follow-up on top of this if that's useful.

dkundel-openai:
Duplicate of #425.

PR #400, opened 2026-06-25
fix: idle self-shutdown for the app-server broker to stop stale brokers accumulating

storyarcade:
### Problem
The shared Codex app-server broker (`app-server-broker.mjs`) is spawned `detached`+`unref`'d and reused per-cwd via endpoint-readiness (`broker-lifecycle.mjs` `ensureBrokerSession`). When a session ends, crashes, or its host process is replaced, the broker keeps its `codex app-server` child (and everything that child fans out) alive with **no client connected**.

It is never torn down, because `ensureBrokerSession` only tears down the **current cwd's** stale entry — it never reaps brokers left behind by sessions in *other* cwds. Across many sessions in distinct working directories, idle brokers accumulate monotonically (observed: a dozen idle brokers, oldest ~15 days, each pinning an app-server subtree).

### Fix
Give the broker an **idle self-shutdown**. It already tracks connected `sockets` and already has a full `shutdown()`; this wires a timer that calls the existing `shutdown()` when no client has been connected for an idle window.

- Default **30 min**, configurable via `CODEX_BROKER_IDLE_SHUTDOWN_MS`, **`<= 0` disables**.
- The timer is **cleared the instant any client connects** and only **arms when `sockets.size === 0`**, so it can never race a live broker: a long turn holds its socket open for the whole turn, and the broker only sits clientless *between* turns (reused by endpoint-readiness, not a held socket). After idle-exit, `ensureBrokerSession` re-probes the now-dead endpoint and respawns cleanly.
- Also reorders `shutdown()` to stop accepting connections **before** the async `appClient.close()`, closing a small window where a connect during teardown could be dropped mid-handshake.

### Tests
`tests/broker-idle-shutdown.test.mjs` uses the existing `fake-codex` fixture (runs in CI without a real login): self-exit when never connected, connected-client cancels the timer, re-arm-and-exit after disconnect, disabled at `<= 0`, and unset-env falls back to the long default. `node --test` passes (a few pre-existing unrelated flaky tests aside).

### DIFF
--- a/plugins/codex/scripts/app-server-broker.mjs
+++ b/plugins/codex/scripts/app-server-broker.mjs
@@ -71,6 +71,48 @@ async function main() {
   let activeStreamThreadIds = null;
   const sockets = new Set();
 
+  // Idle self-shutdown: a broker is spawned detached+unref'd per session and is reused by
+  // endpoint-readiness, keyed by cwd. When a session ends (or crashes, or its host process is
+  // replaced) the broker keeps its app-server child alive with no client connected, and is never
+  // torn down because teardown only ever targets the *current* cwd's stale entry. Across sessions in
+  // distinct cwds these accumulate. Exit when no client has been connected for an idle window so a
+  // dead session's broker reaps itself. Disabled when the window is <= 0.
+  //
+  // Why this never races a live broker: a client connects per logical operation and disconnects when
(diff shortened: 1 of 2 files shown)


dkundel-openai:
Duplicate of #580.

PR #407, opened 2026-06-28
fix: recover structured output for /codex:adversarial-review when the turn uses tools

qinjie:
## Problem

`/codex:adversarial-review` fails with `Unexpected token … in JSON at position 0` on every non-trivial review. When Codex investigates the diff before answering, **codex-cli 0.142.3 ends the `outputSchema` turn without emitting the forced `final_answer`**, so the companion captures a `phase: "commentary"` preamble and `parseStructuredOutput`'s bare `JSON.parse` throws.

Plain `/codex:review` is unaffected — it uses the native reviewer (`exitedReviewMode` → `reviewText`), no `outputSchema`, no `JSON.parse`.

### What the user sees

```
# Codex Adversarial Review

Codex did not return valid structured JSON.

- Parse error: Unexpected token I in JSON at position 0

Raw final message:

I'll inspect the HEAD~2..HEAD diff directly and trace the changed runtime paths …
```

Turn duration is suspiciously short (~9–12s): Codex investigates, the turn completes, but no structured `final_answer` is ever emitted.

### Root cause

Driving `codex app-server` directly (initialize → `thread/start` → `turn/start` with an `outputSchema` that has a `verdict` property) shows:

- **No tool use** ("reply with approval") → `item/completed { agentMessage, phase:"final_answer", text:'{"verdict":"approve",…}' }` ✅ schema works.
- **With tool use** ("read file X, then return JSON") → `agentMessage phase=commentary "I'll read…"` → several `commandExecution` items → `turn/completed` with **no `final_answer`** ❌.

So this is a codex-cli behavior (forced-schema output dropped when one turn mixes tool calls + `outputSchema`), but the plugin makes it fatal because `recordItem()` sets `lastAgentMessage` for every `agentMessage` (including non-final `commentary`), and `parseStructuredOutput` hard-fails on prose.

## Fix

In `plugins/codex/scripts/lib/codex.mjs`:

1. **Two-turn elicitation in `runAppServerTurn`** — new opt-in `structuredRetryPrompt` option. If a schema was requested but the captured message isn't JSON once the turn completes, resume the *same thread* with a JSON-only prompt (retry budget `structuredRetryAttempts`, default 3). Same-thread resume retains the investigation context, so the model emits the verdict it skipped. Guarded by `!looksLikeJson(finalMessage)`, so it never fires once the CLI emits `final_answer` directly — a forward-compatible no-op once the upstream CLI bug is fixed.
2. **`parseStructuredOutput` tolerates a fenced ```json block** via a small `stripJsonFences` helper (also used by `looksLikeJson`).

In `plugins/codex/scripts/codex-companion.mjs`:

3. The adversarial-review call site passes `structuredRetryPrompt`. The `task` path and `/codex:review` are untouched.

> **Maintainer note:** a cleaner long-term variant of (1) is to prefer the last `agentMessage` with `phase === "final_answer"` when populating `lastAgentMessage`/`finalMessage`, so non-final `commentary` can never clobber the answer. That's included in spirit but is **necessary-but-not-sufficient** on 0.142.3 — no `final_answer` is emitted *at all* after tool use, so the resume turn is what actually produces the JSON. Happy to reshape toward whichever approach you prefer.

## Tests

- `tests/fake-codex-fixture.mjs`: new `adversarial-tooluse-then-json` behavior that, on the first `outputSchema` turn, emits a `commentary` preamble + a `commandExecution` and completes with **no** `final_answer`; the follow-up JSON-only turn emits the structured `final_answer`. This reproduces the 0.142.3 protocol exactly.
- `tests/runtime.test.mjs`: asserts `adversarial-review` exits 0 and renders `Verdict:` (not "did not return valid structured JSON") under that behavior.

Verified fail-before / pass-after: the new test fails on `main` with the exact parse error and passes with this change. The rest of the suite's pass/fail set is unchanged by this PR.

```
npm test
```

## Risk / scope

- Opt-in via `structuredRetryPrompt`; the `task` path and `/codex:review` are untouched.
- Forward-compatible: becomes a no-op if codex-cli later emits `final_answer` after tool use.
- Adds one extra LLM turn (up to N retries) only on the adversarial path and only when the first turn produced no JSON.

---

### DIFF
--- a/plugins/codex/scripts/codex-companion.mjs
+++ b/plugins/codex/scripts/codex-companion.mjs
@@ -413,6 +413,10 @@ async function executeReviewRun(request) {
     model: request.model,
     sandbox: "read-only",
     outputSchema: readOutputSchema(REVIEW_SCHEMA),
+    /* LOCAL PATCH (two-turn-review): if the investigation turn ends without the
+     * forced JSON, resume the thread once to elicit the verdict. */
+    structuredRetryPrompt:
+      "Output ONLY the JSON review object matching the required schema, based on the investigation you just completed. Do not run any commands or include any prose outside the JSON.",
     onProgress: request.onProgress
   });
   const parsed = parseStructuredOutput(result.finalMessage, {
--- a/plugins/codex/scripts/lib/codex.mjs
+++ b/plugins/codex/scripts/lib/codex.mjs
@@ -87,6 +87,34 @@ function buildTurnInput(prompt) {
   return [{ type: "text", text: prompt, text_elements: [] }];
 }
 
+/* LOCAL PATCH (two-turn-review): codex-cli omits the forced-schema final_answer
+ * when a turn also runs tools, leaving only a prose preamble. These helpers let
+ * runAppServerTurn detect that and parseStructuredOutput tolerate fenced JSON.
+ * Auto-re-applied by ~/Documents/claude-setup-review-loop-with-codex/hooks/commit-review-loop.mjs. */
+function stripJsonFences(text) {
+  const trimmed = String(text ?? "").trim();
+  if (!trimmed.startsWith("```")) {
+    return trimmed;
+  }
+  return trimmed
+    .replace(/^```(?:json)?\s*/i, "")
+    .replace(/\s*```$/, "")
+    .trim();
+}
+
+function looksLikeJson(text) {
+  const candidate = stripJsonFences(text);
+  if (!candidate || (candidate[0] !== "{" && candidate[0] !== "[")) {
+    return false;
+  }
+  try {
+    JSON.parse(candidate);
+    return true;
+  } catch {
+    return false;
+  }
+}
+
 function shorten(text, limit = 72) {
   const normalized = String(text ?? "").trim().replace(/\s+/g, " ");
   if (!normalized) {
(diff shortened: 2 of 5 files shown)


axisrow:
@qinjie — a note from a neighbouring PR, in case it's useful. Not a defect in this change; an interaction with the turn-budget work in #376.

Your retry loop issues up to `structuredRetryAttempts` (default 3) additional `captureTurn` calls when the captured message isn't valid JSON. Each `captureTurn` gets its **own** turn budget, so under any fixed wall-clock per-turn budget the total is up to ~4x the configured timeout (1 initial + 3 retries) while every individual turn stays formally inside its deadline. Nothing here exceeds its budget; the budget just isn't aggregate.

Where that bites is the foreground path. #376 sets a foreground budget of 110000ms specifically to stay under Claude Code's Bash-tool ceiling (~120s) so a timeout surfaces as a structured error instead of a SIGKILL. With retries, worst case is ~440s against a ~120s ceiling — the host kills the process first, and the user gets a silent kill with no output rather than the structured timeout #376 exists to provide. Rare (it needs repeated non-JSON output), but it's precisely the failure mode that work set out to eliminate.

Two ways out. A caller-level aggregate budget across the retry loop is the direct fix. The other is to change what the timer measures: an **idle** (inactivity) timeout, reset on every turn notification, removes the multiplication for free — total wall-clock becomes (attempts × real work time) rather than (attempts × worst-case budget), so a retry loop is bounded by how long the agent is actually silent, not by how many attempts it makes. I've implemented that in my fork (axisrow/codex-plugin-cc#41) and suggested it on #376.

Worth coordinating whichever lands first — #376 and this PR touch the same `handleReviewCommand` options object (as does #375). Purely informational; no ask on this PR.

PR #408, opened 2026-06-30
fix(app-server): pass `-c model="..."` to `codex app-server` so options.model takes effect

hiirott:
## Summary

`runAppServerTurn` / `runAppServerReview` accept a `model` option that is forwarded to the `thread/start` and `turn/start` requests. However, the spawn of `codex app-server` itself (in `lib/app-server.mjs`) never adds a corresponding `-c model=\"…\"` config override, so the codex CLI session boots with the **CLI's built-in default model**.

On a ChatGPT account where the default (`gpt-5.3-codex`) is unsupported, every task launched via:

```bash
node scripts/codex-companion.mjs task --background --model gpt-5.4 "<prompt>"
```

fails during the first `turn/start` with:

```
Codex error: 'gpt-5.3-codex' model is not supported when using Codex with a ChatGPT account.
```

even though the user explicitly passed `--model gpt-5.4`.

`codex exec --model gpt-5.4 "say hi"` works fine — the regression is specific to the app-server path used by the companion.

## Fix

Thread the `model` option through to the spawn site:

- **`lib/app-server.mjs`** — when `this.options.model` is set, add `-c model=${JSON.stringify(model)}` to the spawned argv. JSON quoting yields a valid TOML string literal the codex CLI's `-c` parser accepts unambiguously (e.g. `model=\"gpt-5.4\"`).
- **`lib/codex.mjs` `withAppServer`** — accept a third `clientOptions` argument and forward it to `CodexAppServerClient.connect` for both the primary and the direct-retry connection.
- **`lib/codex.mjs` `runAppServerReview` / `runAppServerTurn`** — pass `{ model: options.model }` through to `withAppServer`.

When the caller does not set `options.model`, behavior is unchanged (no `-c` flag is appended).

`interruptAppServerTurn` and `findLatestTaskThread` still create their own short-lived clients without a model override; they are not exercised by the failing `task --background --model` path and are left for a follow-up.

## Test plan

Local verification on macOS / codex-cli 0.130.0:

- [x] Before this patch: `node companion.mjs task --background --fresh --model gpt-5.4 \"say hi\"` → job fails with the `gpt-5.3-codex is not supported` error during `turn/start`.
- [x] After this patch: identical command → status=`completed`, ~7 s, assistant returns \"Hi\". Job id `task-mqzvgcqz-7o506t`, thread `019f15cb-b60a-7013-ace1-ed235e61de96`.
- [x] When `--model` is omitted, no `-c model=...` flag is appended (verified by inspecting the spawned argv) — behavior unchanged for callers that rely on the CLI default.

## Notes

- Originally surfaced while debugging tester orchestration in [hiirott/claude-config#2](https://github.com/hiirott/claude-config/issues/2).
- Broker mode is also covered: the broker daemon spawns its inner app-server via the same `CodexAppServerClient.connect(..., { disableBroker: true })` path in `app-server-broker.mjs`, so this fix is effective for both Spawned and Broker transports.

### DIFF
--- a/plugins/codex/scripts/app-server-broker.mjs
+++ b/plugins/codex/scripts/app-server-broker.mjs
@@ -52,7 +52,7 @@ async function main() {
   }
 
   const { options } = parseArgs(argv, {
-    valueOptions: ["cwd", "pid-file", "endpoint"]
+    valueOptions: ["cwd", "pid-file", "endpoint", "model"]
   });
 
   if (!options.endpoint) {
@@ -65,7 +65,7 @@ async function main() {
   const pidFile = options["pid-file"] ? path.resolve(options["pid-file"]) : null;
   writePidFile(pidFile);
 
-  const appClient = await CodexAppServerClient.connect(cwd, { disableBroker: true });
+  const appClient = await CodexAppServerClient.connect(cwd, { disableBroker: true, model: options.model });
   let activeRequestSocket = null;
   let activeStreamSocket = null;
   let activeStreamThreadIds = null;
--- a/plugins/codex/scripts/lib/app-server.mjs
+++ b/plugins/codex/scripts/lib/app-server.mjs
@@ -186,7 +186,11 @@ class SpawnedCodexAppServerClient extends AppServerClientBase {
   }
 
   async initialize() {
-    this.proc = spawn("codex", ["app-server"], {
+    const args = ["app-server"];
+    if (this.options.model) {
+      args.push("-c", `model=${JSON.stringify(String(this.options.model))}`);
+    }
+    this.proc = spawn("codex", args, {
       cwd: this.cwd,
       env: this.options.env,
       stdio: ["pipe", "pipe", "pipe"],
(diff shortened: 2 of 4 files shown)


axisrow:
Nice catch on the root cause — confirming it from a different angle in case it helps this land.

I independently hit the same wall while trying to forward `model`/`effort` to the review commands: the RPC params (`thread/start`, `turn/start`, `review/start`) carry a `model` field, but probing the running app-server shows it has no effect — the session boots with the CLI default. Your `-c model="…"`-at-spawn approach matches what I concluded is the only effective path, so this is the right fix direction.

Two things worth folding in / coordinating on:

1. **`effort` needs the same treatment.** `#476` reports `--effort` is silently ignored. Even where the RPC param exists (`turn/start` has `effort`), whether the server honors it is the same open question as `model` — and if the answer is "needs `-c model_reasoning_effort=…` at spawn" (parallel to your `model` override), the spawn-override site you're touching here is the natural place to handle both, rather than a separate effort-only PR. Worth checking whether codex accepts `model_reasoning_effort` via `-c` and unifying the override plumbing.

2. **Native `review` has no `effort` field in the protocol at all** (`ReviewStartParams = { threadId, target, delivery? }`; `tsc` rejects `model`/`effort` on it with `TS2353`), and the server drops unknown fields (serde without `deny_unknown_fields`). So for native review, `-c …` at spawn is the *only* lever for both model and effort — another reason a single spawn-time override mechanism (this PR, possibly generalized like the `CODEX_PLUGIN_CC_ARGS` idea in #418/#419) is cleaner than per-RPC-field plumbing.

Happy to test/help if you want to extend this to cover `effort`. Not trying to step on the PR — just want to avoid a second incompatible effort-fix landing separately.

Issue #409, opened 2026-06-30
[Windows] POSIX path conversion mangles slash-prefixed CLI args under Git Bash


mariohe0007:
## Summary

On Windows, when the codex-companion runtime is invoked from a Git Bash
(MSYS2) shell, arguments that look like POSIX absolute paths (e.g. a PID
passed as `/12345`, or flags like `/codex:cancel`) get auto-translated by
MSYS's path-conversion layer into `C:/Program Files/Git/12345`, which
breaks the call.

## Reproduction

Environment:
- Windows 10 / 11
- Git for Windows (MSYS2-based bash)
- `@openai/codex-plugin-cc` 1.0.5

Steps:
1. From Git Bash, start a background Codex job and obtain its job id.
2. Run `/codex:cancel <pid>` where `<pid>` is passed as `/12345`.
3. Observe that the call fails — the argument received by the runtime is
   `C:/Program Files/Git/12345`, not `/12345`.

## Workaround

Set `MSYS_NO_PATHCONV=1` in the user's shell profile (`~/.bashrc` for Git
Bash; `$PROFILE` for any PowerShell session that may spawn Git Bash
sub-shells). This globally disables MSYS path conversion for the session.

## Suggested fix

Either:
- Have the plugin's wrapper script set `MSYS_NO_PATHCONV=1` (or use the
  per-arg `MSYS2_ARG_CONV_EXCL=*` form) when spawning child processes on
  Windows, scoped narrowly so it does not pollute the parent shell, or
- Document the requirement explicitly in the README's Windows section.

Happy to send a PR if the maintainers prefer the wrapper-level fix.

Issue #331, opened 2026-05-19
codex-companion cancel taskkill fallback breaks under Git Bash — MSYS translates /PID arg to C:/Program Files/Git/PID


lukaskucinski:
## Environment

- codex plugin v1.0.4
- Windows 10 Pro 10.0.19045
- Git Bash (MSYS2) — the default Bash environment for Claude Code on Windows
- `node codex-companion.mjs cancel <job-id>` invoked when a job has stalled

## Symptom

When `codex-companion`'s soft-cancel IPC fails and the script falls back to `taskkill /PID <pid> /T /F`, Git Bash's MSYS layer translates the literal `/PID` argument into a Windows path, mangling the command:

```
taskkill /PID 34428 /T /F: exit=1: ERROR: Invalid argument/option - 'C:/Program Files/Git/PID'.
Type "TASKKILL /?" for usage.
```

MSYS sees `/PID` as a root-level Unix path and prefixes it with the Git for Windows install root. The actual `taskkill` receives `C:/Program Files/Git/PID 34428 /T /F` and rejects the bogus argument.

Result: the cancel fallback fails, the wedged codex process keeps running, and the user has to clean up manually via `Stop-Process -Id <pid> -Force` from PowerShell.

## Reproduction

1. Launch any codex adversarial-review job from Git Bash on Windows
2. Trigger a stall condition (e.g., the IPC deadlock described in #330; or kill the codex process out-of-band so the broker thinks it's running but it isn't)
3. Run `node codex-companion.mjs cancel <job-id>` from Git Bash
4. Observe the MSYS-mangled taskkill stderr

## Suggested fixes (any of these works)

1. Set `MSYS_NO_PATHCONV=1` in the env for the taskkill invocation
2. Pass `//PID` (double-slash escape recognized by MSYS as "no translation")
3. **Preferred:** on Windows, use `powershell.exe -NoProfile -Command "Stop-Process -Id <pid> -Force"` instead of `taskkill` — no MSYS translation, cleaner exit semantics, no dependency on which shell wraps the node invocation

Option 3 is what we recommend to our users in our orchestrator's recovery guidance, so confirmed working.

## Reference

Same orchestrator run as the IPC deadlock issue (#330). Job ID `review-mpbtwgbj-gswbux`, PID 34428.

ryantokam:
Confirming this on a second environment and adding the exact code location + a tested minimal patch.

[...]

PR #411, opened 2026-06-30
fix: preserve taskkill flags under Git Bash

Epochex:
## Summary
- scope MSYS path-conversion opt-out to the Windows `taskkill` process-tree cleanup call
- preserve `/PID`, `/T`, and `/F` when Claude Code is launched from Git Bash/MSYS
- cover the Windows helper path with a regression assertion for the child env

Addresses #331. Related to #409.

## Tests
- `node --test tests\process.test.mjs`
- `git diff --check`

Not run: full `npm.cmd test` timed out locally on Windows while runtime broker fixtures kept running; this change is covered by the focused process helper test.

### DIFF
--- a/plugins/codex/scripts/lib/process.mjs
+++ b/plugins/codex/scripts/lib/process.mjs
@@ -54,6 +54,14 @@ function looksLikeMissingProcessMessage(text) {
   return /not found|no running instance|cannot find|does not exist|no such process/i.test(text);
 }
 
+function envWithoutMsysPathConversion(env) {
+  return {
+    ...(env ?? process.env),
+    MSYS_NO_PATHCONV: "1",
+    MSYS2_ARG_CONV_EXCL: "*"
+  };
+}
+
 export function terminateProcessTree(pid, options = {}) {
   if (!Number.isFinite(pid)) {
     return { attempted: false, delivered: false, method: null };
@@ -66,7 +74,8 @@ export function terminateProcessTree(pid, options = {}) {
   if (platform === "win32") {
     const result = runCommandImpl("taskkill", ["/PID", String(pid), "/T", "/F"], {
       cwd: options.cwd,
-      env: options.env
+      // Keep taskkill's /PID and /T flags intact when Claude Code runs under Git Bash/MSYS.
+      env: envWithoutMsysPathConversion(options.env)
     });
 
     if (!result.error && result.status === 0) {
--- a/tests/process.test.mjs
+++ b/tests/process.test.mjs
@@ -7,8 +7,9 @@ test("terminateProcessTree uses taskkill on Windows", () => {
   let captured = null;
   const outcome = terminateProcessTree(1234, {
     platform: "win32",
-    runCommandImpl(command, args) {
-      captured = { command, args };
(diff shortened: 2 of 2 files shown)


rajpratham1 (approved):
This is a focused, low-risk fix for a real Windows/MSYS compatibility issue.

Why it looks correct
It introduces:
envWithoutMsysPathConversion()

which sets:

MSYS_NO_PATHCONV=1
MSYS2_ARG_CONV_EXCL=*

These are the standard environment variables used to disable Git Bash/MSYS argument rewriting.

Previously:
taskkill /PID 1234 /T /F

could be rewritten by Git Bash, causing /PID and /T to be interpreted as paths instead of switches.

Now only the taskkill invocation receives the modified environment:
env: envWithoutMsysPathConversion(options.env)

so the change is isolated and doesn't affect other commands.

Test quality

The updated test verifies:

Windows platform
Git Bash (SHELL=...bash.exe)
existing environment variables are preserved
MSYS_NO_PATHCONV
MSYS2_ARG_CONV_EXCL
correct taskkill arguments

This is exactly the regression that needed coverage.

Risk

Very low.

The environment override applies only to the Windows taskkill execution and preserves the caller's existing environment.

Minor suggestion

envWithoutMsysPathConversion() could avoid recreating the environment if the variables are already set, but this is purely an optimization.

BlueHarmel:
Independent reproduction, in case it helps this land.

**Environment:** Windows 11, Codex CLI 0.146.0, plugin 1.0.6, Git Bash (`SHELL=C:\Program Files\Git\bin\bash.exe`), non-English (ko-KR) locale.

`runCommand` defaults to `process.env.SHELL` on win32, so `terminateProcessTree` runs `taskkill` through bash and MSYS rewrites the first flag into a path. Measured directly:

```
shell=C:\Program Files\Git\bin\bash.exe | status=1   | ERROR: Invalid argument/option - 'C:/Program Files/Git/PID'
shell=false                             | status=128 | ERROR: The process "999999" not found.
```

`status=128` is the correct "no such process" answer — with bash in the way, `taskkill` never sees `/PID` at all.

**User-visible effect:** `/codex:cancel <job>` reports the taskkill failure but the job stays `running` in `/codex:status`. I hit this today cancelling two stray `task` runs; both remained listed as `running | starting` until the launcher processes were killed by other means.

I patched my local cache with `shell: false` on that one call rather than the env opt-out, and it fixes the same symptom:

```
{"attempted":true,"delivered":true,"method":"taskkill","status":0}   → process gone
```

Not suggesting a change to this PR's approach — `MSYS_NO_PATHCONV` keeps the shell behaviour from #178 intact, which is probably the safer scope. Just adding a second data point that the root cause is the shell, and that the failure is silent from the user's side.

dkundel-openai:
Duplicate of #577.

Epochex:
Understood. #577 carries the same scoped `MSYS_NO_PATHCONV` / `MSYS2_ARG_CONV_EXCL` fix and also handles the localized `taskkill` cases, so closing this one in favor of the active, broader fix makes sense. Thanks for pointing it to the replacement.

PR #422, opened 2026-07-02
fix(stop-review-gate): fail open on infra errors instead of blocking

### LINKED ISSUE #248: Stop review gate blocks Claude on infrastructure errors, causing rewake loops

### LINKED ISSUE #306: Reaching codex rate limit causes an infinite review cycle that consumes claude code tokens for no reason
The below conversations was observed when hitting the 5h rate limit on the ChatGPT subscription.

Ideally the codex hook would fail more silently in the event of a rate limit (maybe output a warning, but definitely not block the session from ending). If I wasn't at the computer, this would have likely continued all the way until claude code rate limits were hit.

Claude Code v2.1.133

```
⏺ Ran 1 stop hook (ctrl+o to expand)
  ⎿  Stop hook error: The stop-time Codex review task failed: {
    "status": 1,
    "threadId": "019e05fc-03b2-77b1-87a0-449bbedb2b44",
    "rawOutput": "",
    "touchedFiles": [],
    "reasoningSummary": []
  }

⏺ That's a tool-level failure, not code-level feedback. The stop-time review
  task itself errored:

  - status: 1 — non-zero exit
  - rawOutput: "" — empty (the review never produced output)
  - touchedFiles: [], reasoningSummary: [] — nothing inspected, nothing reasoned
   about

  This is different from the prior three stop-hook events on this PR sequence,
  all of which surfaced concrete findings (prod rollout path can fail…,
  get-function guard misclassifies…, cicd-strategy.md contradicts…). Empty
  output + status 1 means the Codex helper crashed or timed out before it could
  review anything; there's nothing in the message to act on.

  No code change indicated. PR #10 still contains the cicd-strategy
  contradiction fix from the previous (successful) review round. The branch is
  in the same state it was when I pushed 625206c — review just didn't run this
  time.

  A few options for how to proceed:

  1. Retry the stop-time review. If it was a transient failure (rate limit,
  network blip), a second attempt may succeed and produce real findings. I can't
   trigger this myself — it fires on session-end via your hook.
  2. Move on without retry. Phase 4 is a docs-only PR with two independent
  review cycles already done (the multi-round Codex feedback during
  implementation, plus the in-doc grep checks I ran). Risk of skipping a third
  review on a docs-only change is low.
  3. Diagnose the Codex helper. If you'd like, you can run /codex:setup (the
  codex plugin's setup skill is available in this session) to verify the CLI is
  working before retrying.

  Want me to do anything specific, or is this just informational and you'll
  decide on the merge?

⏺ Ran 1 stop hook (ctrl+o to expand)
  ⎿  Stop hook error: The stop-time Codex review task failed: {
    "status": 1,
    "threadId": "019e05fc-5973-7f62-a3ae-a96f273f611f",
    "rawOutput": "",
    "touchedFiles": [],
    "reasoningSummary": []
  }

⏺ Same failure shape, second time in a row, different threadId — so it's a fresh
   attempt that also crashed empty rather than a single retry hiccup. Two
  consecutive status: 1 / rawOutput: "" runs makes a transient network blip much
   less likely; this looks like a persistent issue with the Codex helper itself
  (auth, CLI binary, runtime config, or rate cap).

  Still no code change indicated — and I shouldn't invent one. PR #10 hasn't
  changed since 625206c.

  If you want to clear the gate, the most direct path is running /codex:setup to
   check whether the Codex CLI is reachable and authenticated. That's the skill
  that's purpose-built for diagnosing this exact case (per its description:
  "Check whether the local Codex CLI is ready and optionally toggle the 
  stop-time review gate"). Want me to invoke it, or would you rather drive that
  yourself?

  Alternative: the same skill mentions toggling the stop-time review gate. If
  you've already manually reviewed PR #10 and want to merge without further
  auto-review, that's the lever — but I'd rather you make that call than do it
  on my initiative.

⏺ Ran 1 stop hook (ctrl+o to expand)
  ⎿  Stop hook error: The stop-time Codex review task failed: {
    "status": 1,
    "threadId": "019e05fc-a019-7b60-b82a-ae1fabb3b2b9",
    "rawOutput": "",
    "touchedFiles": [],
    "reasoningSummary": []
  }

⏺ Third consecutive identical empty failure. The Codex helper is reliably broken
   in this session, not flaky. I have no findings to act on and I'm not going to
   invent any. PR #10 stays at 625206c.

  The gate is yours to handle — /codex:setup to diagnose, or toggle the
  stop-time gate off if you want to end the session without it. I'll wait
  quietly.

⏺ Ran 1 stop hook (ctrl+o to expand)
  ⎿  Stop hook error: The stop-time Codex review task failed: {
    "status": 1,
    "threadId": "019e05fc-d3ae-73e2-8fcc-e88024c99aec",
    "rawOutput": "",
    "touchedFiles": [],
    "reasoningSummary": []
  }

⏺ Fourth identical failure. No new information. Standing by.

⏺ Ran 1 stop hook (ctrl+o to expand)
  ⎿  Stop hook error: The stop-time Codex review task failed: {
    "status": 1,
    "threadId": "019e05fc-eebd-7333-9b77-d9082e1a5d7e",
    "rawOutput": "",
    "touchedFiles": [],
    "reasoningSummary": []
  }

⏺ Fifth.

```

fangkangmi:
Fixes #248
Fixes #306

## Problem

The Stop review gate maps **every** non-`ALLOW` outcome to a blocking decision: a genuine `BLOCK:` verdict, but also timeouts, non-zero companion exits (e.g. rate limits), invalid JSON, and empty/unexpected output all returned `{ok:false}`, and `main()` emitted `{"decision":"block"}` for any `!ok`.

When Codex has a transient infrastructure failure, this loops: the hook blocks → Claude re-wakes → the stop-time review re-runs → hits the same failure → blocks again — burning tokens indefinitely with no useful review until the user disables the gate. #306 is the same root cause reached via the rate-limit path.

## Fix

Split the review result into three outcomes — `allow` / `block` / `infra_error`:

- Only a genuine `BLOCK:` verdict emits a Stop `block` (reason text unchanged).
- Infrastructure failures (timeout, non-zero exit, invalid JSON, empty/unexpected output) now **fail open**: a warning is logged to stderr ("could not complete … Allowing the stop instead of blocking; run /codex:review --wait manually or bypass the gate") and no decision is emitted, so the rewake loop cannot form.
- `ALLOW:` behavior is unchanged.

Two hardenings on top:

1. **Verdict salvage:** the companion prints its JSON payload before setting a non-zero exit code, so a turn that fails *after* streaming its final verdict still carries `BLOCK: …` in `rawOutput`. The non-zero-exit path now parses the payload and honors a present `BLOCK` verdict instead of discarding it. True infra failures produce no verdict and still fail open, so the loop protection is unaffected.
2. Reason fragments spliced into the stderr warning are trimmed and stripped of a trailing period (period-terminated error details previously produced `".."`).

Complementary to #352 (always-valid stdout JSON) and #396 (opt-in round cap) — neither addresses the infra-vs-verdict conflation itself.

## Testing

- New fixture behaviors: `stop-gate-infra` (thread/start raises a 429) and `stop-gate-block-then-error` (verdict streamed, then the turn ends `"failed"`).
- New tests: the hook fails open on an infra error (no stdout decision, stderr warning, no `".."`), and still blocks when a failed turn carried a streamed `BLOCK` verdict.
- Full suite: **92/92 pass**.
- Verified end-to-end against the **real Codex CLI (0.138.0)**, no mocks:
  - Genuine auth infra error (`CODEX_HOME` → empty dir): pre-fix hook emitted `{"decision":"block", …}` (the loop trigger); this branch fails open with the stderr warning.
  - Real authenticated review turns: clean work → model returned `ALLOW:` → no block; deliberately broken work → model returned `BLOCK:` (it independently found the planted regression) → block emitted, as before.

### DIFF
--- a/plugins/codex/scripts/stop-review-gate-hook.mjs
+++ b/plugins/codex/scripts/stop-review-gate-hook.mjs
@@ -102,6 +102,15 @@ function parseStopReviewOutput(rawOutput) {
   };
 }
 
+function tryParseVerdict(stdout) {
+  try {
+    const payload = JSON.parse(String(stdout ?? ""));
+    return parseStopReviewOutput(payload?.rawOutput);
+  } catch {
+    return null;
+  }
+}
+
 function runStopReview(cwd, input = {}) {
   const scriptPath = path.join(SCRIPT_DIR, "codex-companion.mjs");
   const prompt = buildStopReviewPrompt(input);
@@ -124,6 +133,14 @@ function runStopReview(cwd, input = {}) {
   }
 
   if (result.status !== 0) {
+    // The companion prints its JSON payload before setting a non-zero exit
+    // code, so a turn that failed AFTER streaming a final verdict still
+    // carries it in rawOutput. Never discard a BLOCK we actually received;
+    // fail open only when no verdict is present.
+    const salvaged = tryParseVerdict(result.stdout);
+    if (salvaged?.outcome === REVIEW_BLOCK) {
+      return salvaged;
+    }
     const detail = String(result.stderr || result.stdout || "").trim();
     return {
       outcome: REVIEW_INFRA_ERROR,
@@ -180,8 +197,11 @@ function main() {
     // Fail open on infrastructure failures so a transient Codex problem cannot
     // turn into an unbounded Stop-hook rewake loop (#248, #306). We warn on
     // stderr but do NOT block, letting the session end normally.
+    // Embedded error details often end with a period; strip it so splicing
+    // into the sentence below does not produce "..".
+    const reason = String(review.reason ?? "").trim().replace(/\.$/, "");
     logNote(
-      `Codex stop-time review could not complete — ${review.reason}. Allowing the stop instead of blocking; ` +
+      `Codex stop-time review could not complete — ${reason}. Allowing the stop instead of blocking; ` +
         "run /codex:review --wait manually or bypass the gate."
     );
     logNote(runningTaskNote);
--- a/plugins/codex/scripts/stop-review-gate-hook.mjs
+++ b/plugins/codex/scripts/stop-review-gate-hook.mjs
@@ -18,6 +18,15 @@ const SCRIPT_DIR = path.dirname(fileURLToPath(import.meta.url));
 const ROOT_DIR = path.resolve(SCRIPT_DIR, "..");
 const STOP_REVIEW_TASK_MARKER = "Run a stop-gate review of the previous Claude turn.";
 
+// Stop-review outcomes. Only a genuine BLOCK verdict should block the Stop
+// event. Infrastructure failures (timeout, crash, rate limit, auth blip, empty
+// or garbled output) must fail OPEN, otherwise the Stop hook re-triggers, the
+// review re-runs, hits the same failure, and loops indefinitely — burning
+// tokens with no useful review (#248, #306).
+const REVIEW_ALLOW = "allow";
+const REVIEW_BLOCK = "block";
+const REVIEW_INFRA_ERROR = "infra_error";
+
 function readHookInput() {
   const raw = fs.readFileSync(0, "utf8").trim();
   if (!raw) {
@@ -70,28 +79,26 @@ function parseStopReviewOutput(rawOutput) {
   const text = String(rawOutput ?? "").trim();
   if (!text) {
     return {
-      ok: false,
-      reason:
-        "The stop-time Codex review task returned no final output. Run /codex:review --wait manually or bypass the gate."
+      outcome: REVIEW_INFRA_ERROR,
+      reason: "the review task returned no final output"
     };
   }
 
   const firstLine = text.split(/\r?\n/, 1)[0].trim();
   if (firstLine.startsWith("ALLOW:")) {
-    return { ok: true, reason: null };
+    return { outcome: REVIEW_ALLOW, reason: null };
   }
   if (firstLine.startsWith("BLOCK:")) {
     const reason = firstLine.slice("BLOCK:".length).trim() || text;
     return {
-      ok: false,
+      outcome: REVIEW_BLOCK,
       reason: `Codex stop-time review found issues that still need fixes before ending the session: ${reason}`
     };
   }
 
   return {
-    ok: false,
-    reason:
-      "The stop-time Codex review task returned an unexpected answer. Run /codex:review --wait manually or bypass the gate."
+    outcome: REVIEW_INFRA_ERROR,
+    reason: "the review task returned an unexpected answer (no ALLOW:/BLOCK: verdict)"
   };
 }
 
@@ -111,19 +118,16 @@ function runStopReview(cwd, input = {}) {
 
   if (result.error?.code === "ETIMEDOUT") {
     return {
-      ok: false,
-      reason:
-        "The stop-time Codex review task timed out after 15 minutes. Run /codex:review --wait manually or bypass the gate."
+      outcome: REVIEW_INFRA_ERROR,
+      reason: "the review task timed out after 15 minutes"
     };
   }
 
   if (result.status !== 0) {
     const detail = String(result.stderr || result.stdout || "").trim();
     return {
-      ok: false,
-      reason: detail
-        ? `The stop-time Codex review task failed: ${detail}`
-        : "The stop-time Codex review task failed. Run /codex:review --wait manually or bypass the gate."
+      outcome: REVIEW_INFRA_ERROR,
+      reason: detail ? `the review task failed: ${detail}` : "the review task failed"
     };
   }
 
@@ -132,9 +136,8 @@ function runStopReview(cwd, input = {}) {
     return parseStopReviewOutput(payload?.rawOutput);
   } catch {
     return {
-      ok: false,
-      reason:
-        "The stop-time Codex review task returned invalid JSON. Run /codex:review --wait manually or bypass the gate."
+      outcome: REVIEW_INFRA_ERROR,
+      reason: "the review task returned invalid JSON"
     };
   }
 }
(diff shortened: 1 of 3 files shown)


rajpratham1 (approved):
This is a solid reliability fix. It changes the stop-review gate to fail open on infrastructure failures while still honoring genuine BLOCK verdicts, preventing infinite stop-hook loops.

Strengths

Correctly distinguishes review outcomes

ALLOW
BLOCK
INFRA_ERROR

This is much clearer than the previous boolean ok/not ok behavior.

Prevents infinite retry loops

Timeouts
429 rate limits
authentication failures
invalid JSON
empty responses

now allow the stop instead of repeatedly re-running the review.

Doesn't weaken real review protection

The important logic is:

const salvaged = tryParseVerdict(result.stdout);
if (salvaged?.outcome === REVIEW_BLOCK) {
    return salvaged;
}

If the review already emitted a BLOCK before crashing, the session is still blocked.

That preserves safety while only failing open for genuine infrastructure failures.

User-facing messages are cleaner
strips trailing periods
avoids ".. Allowing..."
Test coverage

The new tests cover the important regressions:

✅ infrastructure failure → allow stop
✅ streamed BLOCK then process failure → still block
✅ 429/rate-limit behavior
✅ message formatting regression

These directly validate the new behavior.

Minor suggestion

Instead of several repeated constructions like

{
  outcome: REVIEW_INFRA_ERROR,
  reason: ...
}

a small helper (e.g. infraFailure(reason)) could reduce duplication, but this is only a maintainability improvement.

PR #424, opened 2026-07-03
fix: cancel dead pids on non-English Windows (#423)

### LINKED ISSUE #423: `/codex:cancel` throws and leaves the job stuck as "running" when the pid is already dead on non-English Windows
## Environment

* Windows 11 Pro, zh-TW locale, console codepage 950 (Big5)
* Plugin: `codex` 1.0.4 (marketplace `openai-codex`)
* Claude Code 2.1.193, Node v24.15.0

## Summary

On a non-English Windows system, cancelling a background task whose worker process has already died makes `terminateProcessTree()` throw, so `handleCancel()` never persists the `cancelled` state. The job stays `"running"` forever, and because status/resume trust the stored state without checking pid liveness, the dead job also blocks `--resume-last` resolution. Reproduced three times on 2026-07-03.

## Reproduction

Real flow:

1. On zh-TW Windows, start a background codex task (detached task worker, pid recorded in the job file).
2. Kill the worker out-of-band (hard-close the session, reboot, or `taskkill` it from another terminal). The job file still says `"running"` with the stale pid.
3. Run `/codex:cancel <task-id>`.
4. The command fails with the error below; the job file and state are never updated.

Minimal script (run from the plugin root on any non-English Windows, with `SHELL` unset — i.e. the plugin's normal runtime):

```js
import { spawnSync } from "node:child_process";
const { terminateProcessTree } = await import("./scripts/lib/process.mjs");
// spawnSync waits for exit, so this pid is dead by the next line
const dead = spawnSync("cmd.exe", ["/c", "exit"], { windowsHide: true });
terminateProcessTree(dead.pid); // throws
```

Observed:

```
Error: taskkill /PID 36564 /T /F: exit=128: ���~: �䤣��B�z�{�� "36564"�C
```

The mojibake is the cp950 rendering of `錯誤: 找不到處理程序 "36564"。` — Windows' localized "ERROR: The process "36564" not found."

## Root cause

Three layers in `scripts/lib/process.mjs` (1.0.4):

1. **English-only message matching.** `looksLikeMissingProcessMessage()` (L53-55) only recognizes `not found|no running instance|cannot find|does not exist|no such process`. Localized taskkill output never matches, so the win32 branch of `terminateProcessTree()` falls through to `throw new Error(formatCommandFailure(result))` (L97).
2. **The message can never be matched on most localized systems anyway.** `runCommand()` uses `spawnSync(..., { encoding: "utf8" })` (L8), but taskkill emits the console codepage (cp950 here). The utf8 decode destroys the text into `U+FFFD` runs, so even adding localized patterns to the regex would not help — the exit code is the only locale-independent signal. (On the same machine we have also seen taskkill emit the English message under a different invocation context, so the message text is not even stable per-host — one more reason it cannot be the primary signal.) Empirically, taskkill exits with 128 when the target process does not exist (verified repeatedly on this machine; we found no official Microsoft documentation of taskkill exit codes, which is why keeping the message check as a fallback still makes sense).
3. **Crash-before-persist in `handleCancel()`.** `codex-companion.mjs` calls `terminateProcessTree(job.pid ?? Number.NaN)` (L943) before writing the cancelled state (L956-968), so the throw converts a routine "already dead" situation into a permanently stuck job.

## Suggested fix

Treat exit code 128 as process-not-found in the win32 branch, keeping the message check as fallback:

```diff
     if (!result.error && result.status === 0) {
       return { attempted: true, delivered: true, method: "taskkill", result };
     }
+
+    // taskkill exits 128 when the target process does not exist; its stderr is
+    // localized (and becomes mojibake when a non-UTF-8 codepage is decoded as
+    // utf8), so the exit code is the only locale-independent signal.
+    if (!result.error && result.status === 128) {
+      return { attempted: true, delivered: false, method: "taskkill", result };
+    }
```

We are running exactly this as a local patch; with it, the dead-pid case returns `{ attempted: true, delivered: false }` and `/codex:cancel` completes and persists the cancelled state.

Two further hardening ideas, your call:

* In `handleCancel()`, consider persisting the cancelled state even if `terminateProcessTree()` throws (try/catch, then rethrow or log). Termination failure on a job the user is abandoning anyway probably shouldn't leave the state machine stuck.
* `runCommand()`'s `shell: process.env.SHELL || true` is fragile for this call in another way: if `SHELL` points at a POSIX bash (e.g. Git Bash), MSYS argument conversion mangles `/PID` into `C:/Program Files/Git/PID`, so taskkill fails with exit 1 (invalid argument) even for live processes. `taskkill` needs no shell at all — `shell: false` for this invocation would sidestep both that and the `DEP0190` deprecation warning.

## Impact

Any non-English Windows locale (zh/ja/ko/de/fr/...) where a background task's worker dies before cancellation: the job can never be cancelled through the plugin, stays `"running"` forever, and blocks resume-last. Recovery currently requires manually rewriting the plugin's state files.

HoneyTyagii:
Fixes #423

## Problem

On non-English Windows, cancelling a background task whose worker has already died leaves the job stuck as `"running"` forever (and blocks `--resume-last`).

`terminateProcessTree()` runs `taskkill /PID <pid> /T /F`. When the process is already gone, `taskkill` exits `128` with a **localized** stderr (e.g. zh-TW `錯誤: 找不到處理程序`). Two things then go wrong:

1. `looksLikeMissingProcessMessage()` only matches English strings, so the localized message never matches and the win32 branch falls through to `throw`.
2. `runCommand()` decodes output as `utf8`, but `taskkill` emits the console codepage (cp950, etc.), so the text is mojibake (`U+FFFD` runs) and could never be matched anyway. The exit code is the only locale-independent signal.

Because `handleCancel()` calls `terminateProcessTree()` **before** persisting the cancelled state, that throw turns a routine "already dead" case into a permanently stuck job.

## Changes

- **Treat `taskkill` exit code `128` as process-not-found** in the win32 branch of `terminateProcessTree()`. The message check remains as a fallback.
- **Invoke `taskkill` with `shell: false`** so `SHELL` pointing at Git Bash no longer mangles `/PID` via MSYS path conversion, and to avoid the `DEP0190` deprecation warning. `runCommand()` now accepts a `shell` override.
- **Persist the cancelled state even if `terminateProcessTree()` throws** (try/catch + log in `handleCancel()`), so abandoning a job can never leave the state machine stuck.
- **Tests** for the exit-128/mojibake case and the `shell: false` invocation.

## Testing

`node --test tests/process.test.mjs` — all pass, including the two new cases:

```
✔ terminateProcessTree treats exit 128 as not-found even with localized/mojibake output
✔ terminateProcessTree invokes taskkill without a shell
```

Credit to the detailed root-cause analysis in #423.

### DIFF
--- a/plugins/codex/scripts/codex-companion.mjs
+++ b/plugins/codex/scripts/codex-companion.mjs
@@ -983,7 +983,17 @@ async function handleCancel(argv) {
     );
   }
 
-  terminateProcessTree(job.pid ?? Number.NaN);
+  // Termination failures must not leave the job stuck as "running": the user is
+  // abandoning this job, so persist the cancelled state even if the (possibly
+  // already-dead) worker cannot be terminated.
+  try {
+    terminateProcessTree(job.pid ?? Number.NaN);
+  } catch (error) {
+    appendLogLine(
+      job.logFile,
+      `Process termination failed during cancel${error?.message ? `: ${error.message}` : "."}`
+    );
+  }
   appendLogLine(job.logFile, "Cancelled by user.");
 
   const completedAt = nowIso();
--- a/plugins/codex/scripts/lib/process.mjs
+++ b/plugins/codex/scripts/lib/process.mjs
@@ -1,13 +1,22 @@
 import { spawnSync } from "node:child_process";
 import process from "node:process";
 
+function resolveShellOption(shell) {
+  // Normalize null/undefined to the platform default so callers can pass
+  // through optional values without tripping spawnSync's ERR_INVALID_ARG_TYPE.
+  if (shell === undefined || shell === null) {
+    return process.platform === "win32" ? process.env.SHELL || true : false;
+  }
+  if (typeof shell !== "boolean" && typeof shell !== "string") {
+    throw new TypeError(
+      `runCommand "shell" option must be a boolean or string, received ${typeof shell}`
+    );
+  }
+  return shell;
+}
+
 export function runCommand(command, args = [], options = {}) {
-  const shell =
-    options.shell !== undefined
-      ? options.shell
-      : process.platform === "win32"
-        ? process.env.SHELL || true
-        : false;
+  const shell = resolveShellOption(options.shell);
   const result = spawnSync(command, args, {
     cwd: options.cwd,
     env: options.env,
--- a/plugins/codex/scripts/lib/process.mjs
+++ b/plugins/codex/scripts/lib/process.mjs
@@ -2,6 +2,12 @@ import { spawnSync } from "node:child_process";
 import process from "node:process";
 
 export function runCommand(command, args = [], options = {}) {
+  const shell =
+    options.shell !== undefined
+      ? options.shell
+      : process.platform === "win32"
+        ? process.env.SHELL || true
+        : false;
   const result = spawnSync(command, args, {
     cwd: options.cwd,
     env: options.env,
@@ -9,7 +15,7 @@ export function runCommand(command, args = [], options = {}) {
     input: options.input,
     maxBuffer: options.maxBuffer,
     stdio: options.stdio ?? "pipe",
-    shell: process.platform === "win32" ? (process.env.SHELL || true) : false,
+    shell,
     windowsHide: true
   });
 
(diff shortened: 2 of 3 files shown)


Iszaf (approved):

rajpratham1 (approved):
This is a well-scoped Windows compatibility fix with good regression tests.

What's good
Fixes locale-dependent behavior by relying on the exit code (128) instead of parsing localized taskkill output.
Explicitly runs taskkill with:
shell: false

avoiding MSYS/Git Bash argument mangling (/PID → path conversion) and avoiding unnecessary shell invocation.

Makes cancellation more robust by ensuring a failed terminateProcessTree() call doesn't leave jobs permanently marked as running.
try {
  terminateProcessTree(...);
} catch (...) {
  appendLogLine(...);
}

This matches user expectations: cancelling a job should cancel it even if the process has already exited.

Test coverage

The added tests cover the important cases:

✅ localized/non-English Windows output
✅ exit code 128 handling
✅ taskkill invoked with shell: false

These directly validate the behavior introduced by the patch.

Minor observation

resolveShellOption() changes behavior for runCommand() by allowing null/undefined and validating the type. While this is a reasonable improvement, it's a small behavioral expansion beyond the Windows bug fix. It isn't problematic, but it could have been a separate cleanup commit.

dkundel-openai:
Duplicate of #577.

PR #426, opened 2026-07-04
fix: pass on-request approval policy for write-capable task runs

### LINKED ISSUE #412: /codex:rescue fails to write files — always returns a read-only sandbox error
**Summary**
I want Codex to make an actual file change when I ask `/codex:rescue` to do something that requires writing (create, edit, or delete a file). Every time, it comes back saying it can't because the sandbox is read-only — even for something as simple as creating a plain text file inside my own project folder.

**Environment**
- OS: Windows 11
- Codex CLI: 0.142.4
- Plugin: codex-plugin-cc 1.0.5
- Claude Code, invoking via `/codex:rescue`

**Steps to reproduce**
1. Open a project in Claude Code with the Codex plugin installed and authenticated.
2. Run `/codex:rescue create a file named test.txt containing "hello"` (or any task that clearly requires a write).
3. Wait for the result.

**Expected**
The file gets created.

**Actual**
Codex reports the workspace as read-only and refuses to write, every time — not occasionally.

**Related**
Possibly the same underlying issue as openai/codex#6428, though that one isn't specific to this plugin and I couldn't find a plugin-side issue describing this exact case.

mvanhorn:
## Summary
Write-capable task runs now send `approvalPolicy: "on-request"` to the app server, so `--write` actually gets a writable sandbox. Previously every run sent `never`, and the Codex CLI silently downgrades a `workspace-write` sandbox to effectively read-only under `approval_policy=never`.

## Why this matters

The reporter of #412 isolated this with a plugin-free repro (Windows 11, Codex CLI 0.142.4, plugin 1.0.5):

> `codex exec -s workspace-write -c approval_policy=never` fails to write; `approval_policy=on-request` succeeds with no prompt, because in-sandbox writes never escalate.

The plugin only toggled `sandbox` between `read-only` and `workspace-write` and hardcoded `approvalPolicy: "never"` everywhere, so even a trivial "create test.txt" rescue task failed with a read-only sandbox error. Two threading gaps caused it: `runAppServerTurn` in `plugins/codex/scripts/lib/codex.mjs` built its `startThread`/`resumeThread` options without `approvalPolicy` (so `buildThreadParams`' existing `options.approvalPolicy ?? "never"` could never see a value), and `executeTaskRun` in `codex-companion.mjs` never passed one.

## Changes
- `approvalPolicy` is forwarded through both `runAppServerTurn` branches (start and resume), and `executeTaskRun` sets it from the run mode: `on-request` for `--write` (including `--resume-last` resumes), `never` for read-only runs - reviews and other read-only paths are behaviorally unchanged.

## Testing
`node --test tests/runtime.test.mjs`: new assertions verify the fake app-server records `approvalPolicy: "on-request"` + `sandbox: "workspace-write"` on write runs, `never` + `read-only` on read-only runs, and `on-request` on the `--resume-last --write` resume path. Full suite matches the baseline commit exactly in a clean checkout (the only failures in my environment are three pre-existing status/result tests that also fail on the base commit due to a live shared app-server on this machine).

Fixes #412

### DIFF
--- a/plugins/codex/scripts/codex-companion.mjs
+++ b/plugins/codex/scripts/codex-companion.mjs
@@ -488,6 +488,7 @@ async function executeTaskRun(request) {
     defaultPrompt: resumeThreadId ? DEFAULT_CONTINUE_PROMPT : "",
     model: request.model,
     effort: request.effort,
+    approvalPolicy: request.write ? "on-request" : "never",
     sandbox: request.write ? "workspace-write" : "read-only",
     onProgress: request.onProgress,
     persistThread: true,
--- a/plugins/codex/scripts/lib/codex.mjs
+++ b/plugins/codex/scripts/lib/codex.mjs
@@ -1105,6 +1105,7 @@ export async function runAppServerTurn(cwd, options = {}) {
       emitProgress(options.onProgress, `Resuming thread ${options.resumeThreadId}.`, "starting");
       const response = await resumeThread(client, options.resumeThreadId, cwd, {
         model: options.model,
+        approvalPolicy: options.approvalPolicy,
         sandbox: options.sandbox,
         ephemeral: false
       });
@@ -1113,6 +1114,7 @@ export async function runAppServerTurn(cwd, options = {}) {
       emitProgress(options.onProgress, "Starting Codex task thread.", "starting");
       const response = await startThread(client, cwd, {
         model: options.model,
+        approvalPolicy: options.approvalPolicy,
         sandbox: options.sandbox,
         ephemeral: options.persistThread ? false : true,
         threadName: options.persistThread ? options.threadName : options.threadName ?? null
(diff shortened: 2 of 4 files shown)


rajpratham1 (approved):
This is a focused bug fix with good test coverage.

What's good
Fixes the root cause by forwarding approvalPolicy to both:
startThread()
resumeThread()
Keeps behavior unchanged for read-only tasks (never).
Enables the intended behavior for write tasks (on-request).
Adds regression tests for:
new write thread
read-only thread
resumed write thread
Updates the fake Codex fixture so tests validate the actual parameters being sent.
Code quality

The implementation is small and consistent:

approvalPolicy: options.approvalPolicy,
sandbox: options.sandbox,

is now passed in both thread creation and resume paths, eliminating the inconsistency.

The companion also correctly maps:

request.write

to

approvalPolicy = on-request
sandbox = workspace-write

which matches expected behavior.

Tests

The new tests verify exactly what the fix intends:

✅ write task → on-request
✅ read-only task → never
✅ resume write task → on-request

This provides good regression coverage.

Issue #410, opened 2026-06-30
Background job state machine retains zombie entries; `--fresh` does not clear them


mariohe0007:
## Summary

The background-job state machine (used by `/codex:rescue`, `/codex:status`,
`/codex:cancel`) appears to retain completed/failed job entries
indefinitely. Re-invoking with `--background --fresh` does not clear or
recycle these zombie entries, so subsequent runs can see and act on stale
state.

## Reproduction

Environment:
- Windows 10 / 11
- `@openai/codex-plugin-cc` 1.0.5

Steps:
1. Run a Codex rescue task to completion via
   `codex:rescue --model gpt-5.5 --effort high --background`.
2. Wait until it shows `status: completed`.
3. Start a new task with `--background --fresh`.
4. Run `/codex:status` — the previously-completed job is still listed,
   and in some cases its job id is reused or its state bleeds into the
   new run's status reporting.

## Suggested fix

- Auto-GC `completed` / `failed` job entries after N minutes (e.g. 30 min).
- Have `--fresh` actively wipe terminal-state job entries before starting
  the new run, not just decline to resume.

Happy to clarify with logs if useful — the state files live under
`~/.claude/plugins/data/codex-openai-codex/state/...`.

PR #435, opened 2026-07-05
Fix background job zombie entries and enhance --fresh clearing

SparshGarg999:
This PR fixes background job state machine retaining zombie entries (Issue #410) in `openai/codex-plugin-cc`:

1. **Auto-Garbage Collection (GC) for Job State (Issue #410)**:
   - **Root Cause**: Previously completed, failed, or stopped jobs were stored in the state database indefinitely, leading to a build-up of stale/zombie entries.
   - **Fix**: Implemented an auto-cleanup process in `listJobs()` (inside `state.mjs`) that automatically filters out completed, failed, or stopped jobs older than 30 minutes. The associated JSON state files and log files on disk are pruned automatically during this process via the existing `saveState` logic.

2. **Enhanced `--fresh` Flag Behavior**:
   - **Root Cause**: The `--fresh` flag did not actively clean up existing terminal-state jobs, meaning stale jobs still interfered or showed up in `/codex:status`.
   - **Fix**: Added a `clearTerminalJobs(cwd)` function to wipe all terminal-state jobs (completed, failed, or stopped) while keeping active/running jobs. This is invoked in `codex-companion.mjs`'s `handleTask` when the `--fresh` option is passed, ensuring a completely clean slate for new runs.

3. **Added Unit Tests**:
   - Added unit test cases in `tests/state.test.mjs` verifying that `listJobs` correctly prunes expired jobs and `clearTerminalJobs` removes only terminal jobs.
   - Fixed a test isolation bug in `tests/state.test.mjs` where multiple test cases resolved to the same temp workspace directory on Windows (due to parent directory traversing for Git repo roots), by calling `initGitRepo(workspace)` on the unique temp workspaces.

### DIFF
--- a/plugins/codex/scripts/codex-companion.mjs
+++ b/plugins/codex/scripts/codex-companion.mjs
@@ -27,6 +27,7 @@ import { collectReviewContext, ensureGitRepository, resolveReviewTarget } from "
 import { binaryAvailable, terminateProcessTree } from "./lib/process.mjs";
 import { loadPromptTemplate, interpolateTemplate } from "./lib/prompts.mjs";
 import {
+  clearTerminalJobs,
   generateJobId,
   getConfig,
   listJobs,
@@ -778,6 +779,9 @@ async function handleTask(argv) {
   const fresh = Boolean(options.fresh);
   if (resumeLast && fresh) {
     throw new Error("Choose either --resume/--resume-last or --fresh.");
+  }
+  if (fresh) {
+    clearTerminalJobs(cwd);
   }
   const write = Boolean(options.write);
   const taskMetadata = buildTaskRunMetadata({
--- a/plugins/codex/scripts/lib/state.mjs
+++ b/plugins/codex/scripts/lib/state.mjs
@@ -149,8 +149,9 @@ export function upsertJob(cwd, jobPatch) {
 export function listJobs(cwd) {
   const state = loadState(cwd);
   const now = Date.now();
+  const terminalStatuses = ["completed", "failed", "cancelled", "stopped"];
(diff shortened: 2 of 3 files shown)


rajpratham1 (changes requested):
The intent is good, but I don't think this implementation is safe enough yet. The review bot has already pointed out two correctness issues that can cause active jobs to disappear.

Major issues
1. Queued jobs are incorrectly garbage collected

Current logic:

if (job.status !== "running" && job.updatedAt) {
    ...
    return false;
}

This treats every non-running job as terminal.

But elsewhere the codebase treats queued as an active state.

Possible scenario:

User starts a background task.
Job is stored as queued.
Machine sleeps or startup is delayed.
After 30 minutes, listJobs() runs.
The queued job is deleted.
Worker eventually starts but its metadata/logs are gone.

That is data loss.

GC should only remove known terminal states such as:

completed
failed
cancelled/stopped

not every non-running status.

2. clearTerminalJobs() also removes queued jobs

Current code:

job.status === "running"

This means --fresh deletes every queued job.

A queued worker is still active.

Removing it can cause:

missing job metadata
failed worker startup
/codex:cancel no longer works
/codex:status can't find the job

Again, terminal states should be filtered explicitly instead of keeping only "running".

Tests miss these cases

The new tests only cover:

running
completed
failed

They should also include a queued job to verify that:

listJobs() preserves queued jobs
clearTerminalJobs() preserves queued jobs

Without those tests, the regression remains undetected.

What is good
Adding automatic cleanup for stale terminal jobs is a reasonable improvement.
Calling clearTerminalJobs() for --fresh makes sense.
Test coverage for completed/failed cleanup is good.

PR #436, opened 2026-07-05
docs: fix review gate README anchor

CooperSheroy:
## Summary
- update the README Codex resume section to link to the existing review gate heading
- add a command-docs regression assertion so the stale anchor does not return

## Verification
- node --test tests/commands.test.mjs
- git diff --check
- node -e README anchor check for #enabling-review-gate

Note: an initial full `npm test` run was stopped after it appeared stuck after test 77; all emitted tests up to that point had passed. The focused docs test above passed after the change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -287,7 +287,7 @@ Check out the Codex docs for more [configuration options](https://developers.ope
 
 ### Moving The Work Over To Codex
 
-Delegated tasks and any [stop gate](#what-does-the-review-gate-do) run can also be directly resumed inside Codex by running `codex resume` either with the specific session ID you received from running `/codex:result` or `/codex:status` or by selecting it from the list.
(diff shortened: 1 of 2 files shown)


rajpratham1 (approved):
This is a straightforward documentation fix with an appropriate regression test.

Review
README fix ✅

The link changes from:

[stop gate](#what-does-the-review-gate-do)

to

[review gate](#enabling-review-gate)

This fixes a broken internal anchor and also updates the terminology to match the current documentation.

Regression test ✅

The added assertions verify both:

the new anchor exists
assert.match(readme, /\[review gate\]\(#enabling-review-gate\)/i);
the old broken anchor is gone
assert.doesNotMatch(readme, /#what-does-the-review-gate-do/i);

This prevents future documentation regressions.

Risk

None.

Only documentation and a documentation validation test are modified.

PR #438, opened 2026-07-05
fix(windows): resolve cross-platform test compatibility and path separation issues

mayankpandey0:
This PR addresses cross-platform compatibility issues when running the test suite on Windows. Specifically, it resolves POSIX path joining mismatches for Unix socket assertions, wraps Node.js binary access using .cmd scripts to bypass EPERM permissions errors on symlink creation, and dynamically skips symlink creation tests on git.test if sufficient developer/admin permissions are not present. Finally, a parallel/sequential runner script has been created for manual verification.


rajpratham1 (approved):
This PR addresses legitimate cross-platform test failures with minimal, targeted changes.

Review
1. path.posix.join() for Unix socket endpoints ✅
return `unix:${path.posix.join(sessionDir, "broker.sock")}`;

This is the correct fix.

Previously:

path.join(...)

would generate:

unix:C:\...\broker.sock

or use backslashes when run on Windows, even though the URI represents a Unix socket. Using path.posix.join() guarantees forward slashes regardless of the host OS.

2. Skip symlink test on Windows when privileges are unavailable ✅
try {
    fs.symlinkSync(...)
} catch (err) {
    if (process.platform === "win32" && err.code === "EPERM") {
        t.skip(...)
    }
}

This is standard practice.

Creating symlinks on Windows often requires:

Developer Mode
Administrator privileges

Without this guard, CI can fail for environmental reasons rather than code regressions.

3. node.cmd instead of Unix symlink on Windows ✅

Replacing

fs.symlinkSync(process.execPath, ...)

with

node.cmd

on Windows is appropriate because executable symlinks are less portable there. A simple batch wrapper better matches how Windows resolves commands from PATH.

Test impact

The changes improve portability without weakening assertions:

Unix socket paths remain valid.
Windows CI no longer depends on symlink privileges.
Node discovery works on both Windows and Unix.
Risk

Very low.

All changes are platform-specific and improve compatibility without altering production behavior.

Issue #441, opened 2026-07-06
Stop hook flashes a visible console window on Windows at every turn end — consider not registering it while stopReviewGate is disabled


intpure:
## Summary

On Windows, after installing the plugin, a console window flashes at the end of **every** Claude Code turn. Differential testing traced it to the plugin's `Stop` hook (`scripts/stop-review-gate-hook.mjs`), which `hooks/hooks.json` registers unconditionally — even though `stopReviewGate` defaults to `false`, making the hook a no-op that still costs a `node` process spawn (and a visible window flash) on every turn end.

## Evidence

Process-creation monitor (300ms polling) on Windows 11, Claude Code 2.1.200, plugin v1.0.5:

- Every turn end spawns the plugin hook: `bash.exe -c "node .../stop-review-gate-hook.mjs"` plus a `conhost.exe`.
- Differential: user-defined Stop hooks from `settings.json` (pwsh scripts, same spawn mechanism, running for 2 weeks / ~200 turn ends a day) do **not** produce a visible flash; the flashing started the day the plugin was installed and stopped immediately after removing the `Stop` block from the plugin's cached `hooks.json` (verified over multiple turns with the monitor still recording — the settings.json hooks kept firing, no flash).

## Suggested fix

Since the stop-review gate is opt-in and defaults to off, consider one of:

1. Not registering the `Stop` hook statically in `hooks.json`; register/enable it only when the user enables `stopReviewGate` (if the hooks API allows), or
2. Documenting that Windows users who keep the gate disabled can remove the `Stop` block from `hooks.json` to avoid the per-turn flash and process spawn.

The window-flash mechanics may partly live in Claude Code's hook runner (cf. anthropics/claude-code#19012), but the plugin can avoid the cost entirely for the default-off configuration.

## Environment

- codex plugin v1.0.5 (commit 80c31f9), Claude Code 2.1.200
- Windows 11, Node.js v24.16.0, Git Bash as hook shell

Related: #440 (missing `windowsHide` in `spawnBrokerProcess()` — a different spawn site, filed separately).

PR #451, opened 2026-07-08
fix: hide detached broker window on Windows

### LINKED ISSUE #440: spawnBrokerProcess() missing windowsHide: true — leftover spawn site from #67
## Summary

`spawnBrokerProcess()` in `plugins/codex/scripts/lib/broker-lifecycle.mjs` is missing `windowsHide: true` in its `spawn()` options. On Windows, the detached broker process gets its own visible console window.

#67 fixed this class of bug by adding `windowsHide: true` to the other spawn sites — `lib/app-server.mjs` (codex app-server), `lib/process.mjs` (runCommand), and `codex-companion.mjs` (spawnDetachedTaskWorker) all have it in v1.0.5 — but the broker spawn site was missed.

## Location

`plugins/codex/scripts/lib/broker-lifecycle.mjs`, `spawnBrokerProcess()`:

```js
const child = spawn(process.execPath, [scriptPath, "serve", "--endpoint", endpoint, "--cwd", cwd, "--pid-file", pidFile], {
  cwd,
  env,
  detached: true,
  stdio: ["ignore", logFd, logFd]
  // <-- windowsHide: true missing here
});
```

## Impact

Whenever `ensureBrokerSession()` needs to (re)spawn the broker on Windows, a console window appears and stays open for the lifetime of the broker process. Because the window is visible and closable, users are also likely to close it manually — killing the broker — which makes the next connect probe fail and respawn yet another visible window, so the problem repeats instead of happening once.

## Fix

One line, same as the other spawn sites:

```js
detached: true,
windowsHide: true,
stdio: ["ignore", logFd, logFd]
```

## Environment

- codex plugin v1.0.5 (commit 80c31f9), Claude Code 2.1.200
- Windows 11, Node.js v24.16.0

Verified locally by patching the installed plugin — the visible console window no longer appears.

e345ee:
## Summary

- Add `windowsHide: true` to the detached broker spawn path.
- Add a focused unit test covering the broker spawn options.

Fixes #440.

## Test plan

- `node --test tests/broker-lifecycle.test.mjs`
- `git diff --check`
- `npm test`
- `npm run build`

### DIFF
--- a/plugins/codex/scripts/lib/broker-lifecycle.mjs
+++ b/plugins/codex/scripts/lib/broker-lifecycle.mjs
@@ -56,12 +56,13 @@ export async function sendBrokerShutdown(endpoint) {
   });
 }
 
-export function spawnBrokerProcess({ scriptPath, cwd, endpoint, pidFile, logFile, env = process.env }) {
+export function spawnBrokerProcess({ scriptPath, cwd, endpoint, pidFile, logFile, env = process.env, spawnImpl = spawn }) {
   const logFd = fs.openSync(logFile, "a");
-  const child = spawn(process.execPath, [scriptPath, "serve", "--endpoint", endpoint, "--cwd", cwd, "--pid-file", pidFile], {
+  const child = spawnImpl(process.execPath, [scriptPath, "serve", "--endpoint", endpoint, "--cwd", cwd, "--pid-file", pidFile], {
     cwd,
     env,
     detached: true,
+    windowsHide: true,
     stdio: ["ignore", logFd, logFd]
   });
   child.unref();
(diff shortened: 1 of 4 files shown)


Wintersta7e:
The `spawnImpl` seam plus asserting the captured options is a better test than scanning source text — it proves the option actually reaches `spawn`, rather than that the string appears somewhere in the file.

Worth folding in while you're in here: there's a **third spawn site** with the same problem that neither #440 nor #441 covers. `plugins/codex/scripts/stop-review-gate-hook.mjs` spawns the companion with

```js
spawnSync(process.execPath, [scriptPath, "task", "--json", prompt], {
  cwd, env: childEnv, encoding: "utf8", timeout: STOP_REVIEW_TIMEOUT_MS
});
```

and no `windowsHide`. Unlike the broker, that one runs at **every turn end** while the review gate is enabled, so on Windows it's a recurring console window rather than a one-off. `lib/process.mjs` already sets `windowsHide: true` on its own `spawnSync`, which makes the hook the odd one out.

PR #456, opened 2026-07-08
fix(tests): keep the suite hermetic inside a live Claude Code session

### LINKED ISSUE #455: Test suite is not hermetic inside a live Claude Code session: 4 failures at HEAD + fixture state leaks into the user's real plugin data dir
## Symptom

Running `npm test` from a shell **inside a live Claude Code session** (the natural environment when Claude itself works on this repo):

1. **4 tests fail at HEAD** with no code change:
   - `status shows phases, hints, and the latest finished job`
   - `status preserves adversarial review kind labels`
   - `result returns the stored output for the latest finished job by default`
   - `resolveStateDir uses a temp-backed per-workspace directory`
2. **Fixture state silently pollutes the user's real plugin data.** Two full-suite runs left **~150 fake `codex-plugin-test-*` workspace dirs and ~420 fake job records** under `$CLAUDE_PLUGIN_DATA/state/` — indistinguishable from real workspaces in the on-disk registry (fake jobs like "Investigate flaky test", "check failed auth" sit next to real project jobs).

The same suite passes 91/91 under `env -i PATH=... HOME=... npm test`.

## Root cause — two inherited variables

- **`CLAUDE_PLUGIN_DATA`**: `resolveStateDir` honors it (`lib/state.mjs`), and a live session points it at the real data dir. Both in-process state writes in tests *and* spawned companions (`buildEnv` spreads `process.env`) then write fixture state there instead of the `os.tmpdir()` fallback — which is also exactly why the `resolveStateDir` fallback-path test fails.
- **`CODEX_COMPANION_SESSION_ID`**: set by the plugin's own session-lifecycle hook, inherited by the suite. `filterJobsForCurrentSession` then hides fixture-seeded jobs that carry no `sessionId`, so the status/result tests render empty output.

Note the second one is a test-visible instance of a runtime sharp edge: jobs without a matching `sessionId` are invisible to `status` even with `--all` (`buildStatusSnapshot` applies the session filter before `options.all` is consulted) — filed separately if useful.

## Fix

PR incoming: pin `CLAUDE_PLUGIN_DATA` to a fresh temp root per test process in `tests/helpers.mjs`, clear `CODEX_COMPANION_SESSION_ID` there (session-scoped tests already compose it explicitly), and make the fallback-path state test manage the variable locally — mirroring what its `CLAUDE_PLUGIN_DATA` sibling test already does.

Validated in the reproducing environment (14 live `CLAUDE*`/`CODEX*` vars): 87/91 → **91/91**, zero new entries in the real plugin state dir.

Caleb0796:
Fixes #455.

## Problem

`npm test` run from inside a live Claude Code session fails 4 tests at HEAD and leaks fixture state into the user's **real** plugin data dir (~150 fake `codex-plugin-test-*` workspaces / ~420 job records per two runs). Two inherited variables cause it:

- `CLAUDE_PLUGIN_DATA` → in-process state writes and spawned companions (`buildEnv` spreads `process.env`) resolve state into the real data dir instead of the temp fallback; the `resolveStateDir` fallback test also fails.
- `CODEX_COMPANION_SESSION_ID` (set by the plugin's own session hook) → `status`/`result` session-filter fixture-seeded jobs that carry no `sessionId`, so those tests render empty output.

Full evidence trail in #455.

## Fix (2 files, tests only)

- `tests/helpers.mjs`: pin `CLAUDE_PLUGIN_DATA` to a fresh `mkdtemp` root for the test process (in-process writers and spawned companions then agree on a temp-backed root, so state handling is still exercised end-to-end), and clear `CODEX_COMPANION_SESSION_ID` — every session-scoped test already sets it explicitly in the env it composes (`sess-current` / `sess-other`).
- `tests/state.test.mjs`: the fallback-path test now saves/deletes/restores `CLAUDE_PLUGIN_DATA` locally, exactly like its `uses CLAUDE_PLUGIN_DATA when provided` sibling already does.

Deliberately did NOT strip `CLAUDE*` wholesale in `buildEnv`: broker-mode tests rely on `CLAUDE_ENV_FILE`, and blanket stripping breaks 4 other tests (verified) while still missing the in-process leak path.

## Testing

- Inside a live Claude Code session (14 `CLAUDE*`/`CODEX*` vars in the env): **87/91 → 91/91**, and the real plugin state dir gains **zero** new entries across the run (was ~75 fake workspaces per run before).
- Under a clean env (`env -i PATH HOME`): 91/91 before and after — no behavior change for CI.

### DIFF
--- a/tests/helpers.mjs
+++ b/tests/helpers.mjs
@@ -4,6 +4,18 @@ import path from "node:path";
 import process from "node:process";
 import { spawnSync } from "node:child_process";
 
+// Keep the suite hermetic when it runs from inside a live Claude Code
+// session. Two inherited variables otherwise change behavior:
+// - CLAUDE_PLUGIN_DATA points at the user's real plugin data directory, so
+//   in-process state writes and spawned companions (which inherit it through
+//   buildEnv) leak fixture workspaces and job records into that real
+//   directory. Pin it to a fresh temp root for the whole test process.
+// - CODEX_COMPANION_SESSION_ID makes status/result session-filter jobs that
+//   fixtures seed without a sessionId, hiding them. Session-scoped tests set
+//   the variable explicitly in the env they compose.
+process.env.CLAUDE_PLUGIN_DATA = fs.mkdtempSync(path.join(os.tmpdir(), "codex-plugin-cc-tests-"));
+delete process.env.CODEX_COMPANION_SESSION_ID;
+
 export function makeTempDir(prefix = "codex-plugin-test-") {
   return fs.mkdtempSync(path.join(os.tmpdir(), prefix));
 }
--- a/tests/state.test.mjs
+++ b/tests/state.test.mjs
@@ -9,11 +9,22 @@ import { resolveJobFile, resolveJobLogFile, resolveStateDir, resolveStateFile, s
 
 test("resolveStateDir uses a temp-backed per-workspace directory", () => {
   const workspace = makeTempDir();
-  const stateDir = resolveStateDir(workspace);
+  const previousPluginDataDir = process.env.CLAUDE_PLUGIN_DATA;
+  delete process.env.CLAUDE_PLUGIN_DATA;
 
-  assert.equal(stateDir.startsWith(os.tmpdir()), true);
-  assert.match(path.basename(stateDir), /.+-[a-f0-9]{16}$/);
-  assert.match(stateDir, new RegExp(`^${os.tmpdir().replace(/[.*+?^${}()|[\]\\]/g, "\\$&")}`));
+  try {
+    const stateDir = resolveStateDir(workspace);
+
+    assert.equal(stateDir.startsWith(os.tmpdir()), true);
+    assert.match(path.basename(stateDir), /.+-[a-f0-9]{16}$/);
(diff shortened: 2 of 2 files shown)


BlueHarmel:
Independent reproduction on a different machine, in case a second data point helps.

**Environment:** Windows 11, Node 22, plugin 1.0.6, run from inside a live Claude Code session (`CLAUDE_PLUGIN_DATA=C:/Users/<me>/.claude/plugins/data/codex-openai-codex`).

`tests/state.test.mjs` alone:

```
as-is                        # pass 2  # fail 1
env -u CLAUDE_PLUGIN_DATA    # pass 3  # fail 0
```

The failing one is `resolveStateDir uses a temp-backed per-workspace directory` — `resolveStateDir` returns the real plugin data dir, so the `startsWith(os.tmpdir())` assertion is false:

```
os.tmpdir() : C:\Users\<me>\AppData\Local\Temp
stateDir    : C:\Users\<me>\.claude\plugins\data\codex-openai-codex\state\ws-probe-d627752adbffd412
```

Worth noting for anyone triaging: this is not Windows-specific and CI cannot catch it, because GitHub Actions never sets `CLAUDE_PLUGIN_DATA`. It only shows up for the people most likely to be running the suite — plugin developers working inside Claude Code.

I also see the fixture leak this PR describes: `~/.claude/plugins/data/codex-openai-codex/state/` accumulated `codex-plugin-test-*` workspace dirs from test runs.

PR #461, opened 2026-07-09
fix: cap aggregate untracked content in review context (#405)

### LINKED ISSUE #405: adversarial-review builds an unbounded prompt — large working trees fail with "Input exceeds the maximum length of 1048576 characters"
## Summary

`adversarial-review` (and any `runAppServerTurn`-based review that interpolates the working-tree context into a text prompt) builds an **unbounded aggregate prompt**. `collectReviewContext` in `scripts/lib/git.mjs` caps each untracked file at 24KB individually, but never caps the **total** assembled size. With enough untracked files, the prompt exceeds Codex's input limit and the run dies with:

```
Input exceeds the maximum length of 1048576 characters.
```

…exit 1, empty stdout. The failure is opaque: nothing in the message points at "your working tree is too big for an inline review."

## Repro

In a repo with many untracked files (e.g. agent/session tooling artifacts that aren't gitignored):

```
node scripts/codex-companion.mjs adversarial-review --scope working-tree --model gpt-5.5 --json "any focus"
# -> exit 1
# stderr: Input exceeds the maximum length of 1048576 characters.
```

Real case: a working tree with 309 untracked files (~6.8MB). `git ls-files --others --exclude-standard` returns them all; each ≤24KB file is inlined in full into the `Untracked Files` section, summing to ~1.7MB of prompt — over the 1,048,576-char ceiling.

## Root cause

`scripts/lib/git.mjs`:

- `MAX_UNTRACKED_BYTES = 24 * 1024` — per-file cap on untracked content (good).
- `DEFAULT_INLINE_DIFF_MAX_BYTES = 256 * 1024` — cap on the inline **diff** portion (good).
- **No aggregate cap** on the untracked section, nor on the final assembled `REVIEW_INPUT`. `state.untracked.map(formatUntrackedFile).join(...)` grows linearly with file count with no ceiling.

So two independently-bounded sections (diff ≤256KB, each untracked ≤24KB) compose into an unbounded whole.

## Suggested fix

Bound the assembled context, ideally with a clear truncation notice rather than a downstream hard error. Options:

1. Track a running byte budget across the untracked section (e.g. reuse/raise a single cap like `DEFAULT_INLINE_DIFF_MAX_BYTES`); once exceeded, stop inlining and append `(N more untracked files omitted: aggregate exceeds <limit> bytes)`.
2. Enforce a final guard on `REVIEW_INPUT` length before the turn, truncating with a visible marker and/or returning an actionable error like "review context exceeds Codex input limit; narrow scope or commit/ignore untracked files" instead of the raw `Input exceeds the maximum length` from the model layer.

Either keeps reviews working on large/dirty trees and makes the boundary explicit.

## Impact / how it surfaced

Downstream callers (e.g. the `council` Claude Code plugin) invoke this headless and report it as a generic "Codex unavailable", which misdirects debugging toward auth/runtime rather than prompt size. The interactive native reviewer (`runAppServerReview` with a git target) is unaffected because it doesn't inline the diff into a prompt — making the failure look mode-specific.

Version: openai-codex plugin 1.0.4 (codex-cli 0.142.2).

ayobamiseun:
Fixes #405

## Problem

`collectReviewContext` bounds the inline diff (256KB) and each untracked file (24KB), but never bounds the **total** Untracked Files section. The section is also inlined even in the lightweight `includeDiff: false` path, so the `maxInlineDiffBytes` heuristic never protects it. With enough untracked files the two independently-bounded sections compose into an unbounded prompt and `adversarial-review` dies with the opaque model-layer error:

```
Input exceeds the maximum length of 1048576 characters.
```

## Repro

Repo with 300 untracked text files just under the 24KB per-file cap (mirrors the issue's real case of 309 files):

| | assembled context |
|---|---|
| before | **7,086,684 chars** (6.7x over the 1,048,576 limit → exit 1) |
| after | **271,243 chars** (review proceeds) |

## Fix

New `formatUntrackedSection` tracks a running byte budget across the untracked entries — default 256KB (`DEFAULT_UNTRACKED_TOTAL_MAX_BYTES`, matching the existing inline-diff cap), overridable via `collectReviewContext(..., { maxUntrackedTotalBytes })`. Once the next entry would exceed the budget, it stops inlining and appends a visible notice instead of failing downstream:

```
(N untracked file(s) omitted: aggregate untracked content exceeds the 262144 byte limit. See the Git Status section for the full file list; inspect the omitted files directly, or commit or gitignore them to narrow the review.)
```

Nothing becomes silently invisible: the Git Status section (`--untracked-files=all`) still names every untracked file, and skip markers (directory/binary/oversized) count toward the budget too. This is option 1 of the fixes suggested in the issue.

## Tests

- `collectReviewContext caps aggregate untracked content instead of growing unbounded` — cap kicks in, omission notice present, Git Status still lists omitted files.
- `collectReviewContext keeps all untracked content when under the aggregate cap` — no behavior change for small trees.
- Full suite: 92/93 pass; the one failure (`resolveStateDir uses a temp-backed per-workspace directory`) fails identically on `main` when run inside a live Claude Code session (pre-existing #455, unrelated).

### DIFF
--- a/plugins/codex/scripts/lib/git.mjs
+++ b/plugins/codex/scripts/lib/git.mjs
@@ -7,6 +7,7 @@ import { formatCommandFailure, runCommand, runCommandChecked } from "./process.m
 const MAX_UNTRACKED_BYTES = 24 * 1024;
 const DEFAULT_INLINE_DIFF_MAX_FILES = 2;
 const DEFAULT_INLINE_DIFF_MAX_BYTES = 256 * 1024;
+const DEFAULT_UNTRACKED_TOTAL_MAX_BYTES = 256 * 1024;
 
 // Git is directly executable on Windows. Repository-derived arguments must never pass through a shell.
 function git(cwd, args, options = {}) {
@@ -33,6 +34,14 @@ function normalizeMaxInlineDiffBytes(value) {
   const parsed = Number(value);
   if (!Number.isFinite(parsed) || parsed < 0) {
     return DEFAULT_INLINE_DIFF_MAX_BYTES;
+  }
+  return Math.floor(parsed);
+}
+
+function normalizeMaxUntrackedTotalBytes(value) {
+  const parsed = Number(value);
+  if (!Number.isFinite(parsed) || parsed < 0) {
+    return DEFAULT_UNTRACKED_TOTAL_MAX_BYTES;
   }
   return Math.floor(parsed);
 }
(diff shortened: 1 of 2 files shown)


sunampe-de:
## Additional downstream observation: oversized Stop-review input

Related to #405 and this PR (#461).

A downstream Windows execution report dated 2026-09-22 records the same `Input exceeds the maximum length of 1048576 characters` error in the Stop-review task. The reported Claude Code version is 2.1.278; the local Companion cache is identified as 1.0.6. This installation has project-specific supervision/integrity controls and prior local transport changes. This is NOT a clean-install upstream reproduction, and the installed sources were not compared against an upstream tag for this report.

Reported measurements, without repository contents:

| Measurement | Value |
| --- | ---: |
| Staged / unstaged / untracked | 0 / 6 / 332 |
| Assembled ctx.content | 1,237,194 characters |
| UTF-8 representation | 1,240,430 bytes |
| Limit stated by the service error | 1,048,576 characters |
| Excess in ctx.content alone | 188,618 characters |
| Inline sections | 349 |

The downstream diagnosis attributes growth to per-file untracked limits without an aggregate limit, including repeated generated and source material. The Stop task failed without an accepted review verdict. These observations come from that execution report; no new model request or Windows reproduction was run for this comment.

The aggregate cap proposed here addresses prompt growth, but our review contract also requires accounting for every required original file. A notice that lists omitted files does not, by itself, establish that the reviewer inspected them. We have not applied or tested this PR and are not reporting its test results as our own.

Is there a supported release or maintenance path for bounded, file-backed/native collection that preserves complete review coverage, detects unreadable or changed inputs, and does not accept a review when required evidence is incomplete? We need to preserve the existing gate, review-state/counters, integrity checks and genuine final verdict. We are not requesting a bypass or a larger service limit.

No private source, credentials, repository identifiers or raw session logs are attached. A synthetic upstream reproduction would be a separate validation step; it has not been performed here.

Issue #452, opened 2026-07-08
Stop-review-gate hook masks the real failure: Node 24 DEP0190 warning displaces the actual error in stderr-first reporting


Reveludden:
Labels (suggestion): bug, windows

Summary
When the stop-time review task fails on Windows with Node 24, the hook reports only a Node DeprecationWarning instead of the actual failure cause. The real error (in our case: ChatGPT usage limit reached) is only discoverable by manually reading the job logs under CLAUDE_PLUGIN_DATA/state/<workspace>/jobs/*.log.

Environment
Windows 11 Pro (10.0.26200)
Node v24.14.1
codex-cli 0.142.5 (logged in via ChatGPT)
Claude Code 2.1.202
codex plugin 1.0.5
What happens
At turn end with stopReviewGate enabled, Claude Code shows:

Stop hook blocking error: The stop-time Codex review task failed:
(node:98328) [DEP0190] DeprecationWarning: Passing args to a child process
with shell option true can lead to security vulnerabilities, as the arguments
are not escaped, only concatenated.
The actual failure (quota exhausted — "You've hit your usage limit...") never reaches the user. We initially chased the deprecation warning as the root cause; the real error was only visible in the companion job log.

Root-cause chain
scripts/lib/app-server.mjs — SpawnedCodexAppServerClient.initialize():

this.proc = spawn("codex", ["app-server"], {
  ...
  shell: process.platform === "win32" ? (process.env.SHELL || true) : false,
});
shell: true (introduced to fix #85, spawn ENOENT for .cmd shims) combined with an args array makes Node 24 emit DEP0190 on the hook process's stderr.

scripts/stop-review-gate-hook.mjs — runStopReview():

if (result.status !== 0) {
  const detail = String(result.stderr || result.stdout || "").trim();
  return { ok: false, reason: detail ? `The stop-time Codex review task failed: ${detail}` : ... };
}
stderr is preferred over stdout. Since stderr now always starts with the DEP0190 warning on Windows/Node 24, the warning displaces whatever useful diagnostics exist. The companion's structured JSON (with rawOutput, threadId) goes to stdout and is never shown.

Impact
Every stop-gate failure on Windows + Node 24 presents as a deprecation warning, regardless of the true cause (quota, auth, network, app-server crash). Users debug the wrong thing. This also compounds #248 (gate blocking on infrastructure errors) by making those errors undiagnosable from the hook output.

Suggested fixes (any one of these helps)
[...]

PR #469, opened 2026-07-10
fix: resolve imported transfer threads when the ledger record diverges (#417)

### LINKED ISSUE #417: ` /codex:transfer` fails with "did not record an imported thread" for both auto-detected and explicit --source
## Description
`codex-companion.mjs transfer` consistently fails with an opaque error, both when auto-detecting the current Claude Code session and when passing an explicit, verified-valid `--source` path.

## Steps to Reproduce
1. From within an active Claude Code session (Windows, Git Bash), run: `node "<plugin-root>/scripts/codex-companion.mjs" transfer`
2. Observe failure.
3. Retry with an explicit, confirmed-existing `.jsonl` path instead of relying on auto-detection: `node "<plugin-root>/scripts/codex-companion.mjs" transfer --source "C:\Users\lukas\.claude\projects\C--Users-lukas-source-repos-pfs-web-app\d845c3e0-b354-4895-8723-c0669d0aa03a.jsonl"`
4. Same failure, byte-for-byte identical error message.

## Expected behavior
The session transfers into a resumable Codex thread, and the command prints a `codex resume <id>`.

## Actual behavior
Both invocations fail identically:

(node:40764) [DEP0190] DeprecationWarning: Passing args to a child process with shell option true can lead to security vulnerabilities, as the arguments are not escaped, only concatenated.
(Use node --trace-deprecation ... to show where the warning was created)
Codex reported that the Claude import completed, but did not record an imported thread. Check the Codex app-server logs for the underlying import error.

Exit code 1. No further detail is surfaced — `codex-companion.mjs status --all` shows no job entry for the transfer attempt at all, so there's no `job-id` to inspect via `status`/`result` for more context.
## Notes
- The `--source` file was confirmed to exist, be a valid non-empty `.jsonl`, and live under the expected `~/.claude/projects/<project>/` directory — ruling out a source-resolution problem. Since the no-argument (auto-detect via `CODEX_COMPANION_TRANSCRIPT_PATH`) and explicit `--source` runs fail identically, the bug appears to be in the Codex-side import/thread-registration step itself, not in path resolution.
- The `[DEP0190]` deprecation warning (child_process spawned with `shell: true` and unescaped args) fires on every invocation and may be worth addressing separately regardless of the main bug.
## Environment
- OS: Windows 11 Pro (10.0.26200)
- Node.js: v24.15.0
- Plugin path/version: `.../plugins/cache/openai-codex/codex/1.0.5/scripts/codex-companion.mjs`
- Shell: Git Bash (MSYS)

ayobamiseun:
Fixes #417

## Problem

`/codex:transfer` consistently fails with `Codex reported that the Claude import completed, but did not record an imported thread` — identically for auto-detected and explicit `--source` paths — even though the import actually succeeds on the Codex side.

After the `externalAgentConfig/import` notification fires, `importedThreadIdForSource` looks up the imported thread in `$CODEX_HOME/external_agent_session_imports.json` by **strict equality** on `source_path` (Node `realpathSync` output) **and** a content sha256 **computed at lookup time**. That match misses real imports two ways:

1. **Path canonicalization mismatch (Windows).** Codex canonicalizes the recorded `source_path` itself; Rust's `fs::canonicalize` yields verbatim paths (`\\?\C:\Users\...`) that never compare equal to Node's `realpathSync` output (`C:\Users\...`). The strict path match can never succeed — matching the reporter's Git Bash / Windows 11 environment where both invocations fail byte-for-byte identically.
2. **Live-transcript hash race (all platforms).** Running transfer from inside an active Claude Code session means the transcript `.jsonl` keeps growing (the transfer invocation itself gets logged) between Codex hashing it at import time and the companion hashing it at lookup time — so `content_sha256` diverges and the record is rejected.

Either way the command exits 1 with an opaque error while a perfectly good imported thread sits in the ledger.

## Fix

`importExternalAgentSession` snapshots the ledger before the import; when the strict match misses, the lookup falls back to **records Codex appended during this import**:

- prefer an appended record whose source path matches after normalization (strip the `\\?\` verbatim prefix, unify `\`/`/` separators, case-fold on win32),
- else accept a lone appended record.

The strict path+hash match stays first so Codex-side dedup of an unchanged, already-imported session still resolves without any appended record. The failure error now also names the ledger path it searched instead of only pointing at app-server logs.

## Tests

New fake-codex behavior `external-import-divergent-ledger` simulates real Codex: the ledger record gets a verbatim-prefixed, backslash-separated `source_path` and a hash of content that differs from what the companion reads back (both #417 divergences at once). The new regression test:

- **fails on `main`**: `transfer` exits 1 with `did not record an imported thread`
- **passes with this fix**: resolves `thr_1` and prints the resume command

All 5 transfer tests pass; full suite 91/92 — the one failure (`resolveStateDir uses a temp-backed per-workspace directory`) fails identically on `main` when run inside a live Claude Code session (pre-existing #455, unrelated).

Note: the `[DEP0190]` deprecation warning mentioned in the issue is a separate stderr-noise concern (#452 territory) and is not addressed here.

### DIFF
--- a/plugins/codex/scripts/lib/codex.mjs
+++ b/plugins/codex/scripts/lib/codex.mjs
@@ -658,16 +658,36 @@ function sourceContentSha256(sourcePath) {
   return crypto.createHash("sha256").update(fs.readFileSync(sourcePath)).digest("hex");
 }
 
-function importedThreadIdForSource(sourcePath) {
-  const ledgerPath = path.join(resolveCodexHome(), "external_agent_session_imports.json");
+function importLedgerPath() {
+  return path.join(resolveCodexHome(), "external_agent_session_imports.json");
+}
+
+function readImportLedgerRecords() {
+  const ledgerPath = importLedgerPath();
   if (!fs.existsSync(ledgerPath)) {
-    return null;
+    return [];
   }
   const ledger = readJsonFile(ledgerPath);
+  return Array.isArray(ledger?.records) ? ledger.records : [];
+}
+
+// Codex canonicalizes the recorded source path itself; on Windows Rust's
+// canonicalization yields verbatim paths (\\?\C:\...) that never compare
+// equal to Node's realpathSync output.
+function normalizeLedgerSourcePath(value) {
+  if (typeof value !== "string" || value === "") {
+    return null;
+  }
+  let normalized = value.startsWith("\\\\?\\") ? value.slice(4) : value;
+  normalized = normalized.replaceAll("\\", "/");
+  return process.platform === "win32" ? normalized.toLowerCase() : normalized;
+}
+
+function importedThreadIdForSource(sourcePath, previousRecords = null) {
+  const records = readImportLedgerRecords();
   const canonicalSource = fs.realpathSync(sourcePath);
   const contentSha256 = sourceContentSha256(canonicalSource);
-  const records = Array.isArray(ledger?.records) ? ledger.records : [];
-  const match = records
+  const strictMatch = records
     .filter(
       (record) =>
         record?.source_path === canonicalSource &&
@@ -675,7 +695,32 @@ function importedThreadIdForSource(sourcePath) {
         typeof record?.imported_thread_id === "string"
     )
     .at(-1);
-  return match?.imported_thread_id ?? null;
+  if (strictMatch) {
+    return strictMatch.imported_thread_id;
+  }
+
+  // Fall back to records Codex appended during this import. The strict match
+  // misses real imports in two known ways: a live Claude transcript keeps
+  // growing while the transfer runs, so the hash computed here trails the
+  // content Codex imported, and Codex may canonicalize the recorded source
+  // path differently than realpathSync does.
+  if (!previousRecords) {
+    return null;
+  }
+  const appended = records
+    .slice(previousRecords.length)
+    .filter((record) => typeof record?.imported_thread_id === "string");
+  const normalizedSource = normalizeLedgerSourcePath(canonicalSource);
+  const pathMatch = appended
+    .filter((record) => normalizeLedgerSourcePath(record?.source_path) === normalizedSource)
+    .at(-1);
+  if (pathMatch) {
+    return pathMatch.imported_thread_id;
+  }
+  if (appended.length === 1) {
+    return appended[0].imported_thread_id;
+  }
+  return null;
 }
 
 function externalAgentSessionMigration(sourcePath, cwd) {
@@ -1066,6 +1111,7 @@ export async function importExternalAgentSession(cwd, options = {}) {
 
   return withDirectAppServer(cwd, async (client) => {
     emitProgress(options.onProgress, "Importing Claude session into Codex.", "transferring");
+    const previousRecords = readImportLedgerRecords();
     try {
       await requestExternalAgentSessionImport(client, externalAgentSessionMigration(options.sourcePath, cwd));
     } catch (error) {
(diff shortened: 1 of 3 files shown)


AndrewAvery7:
Independent confirmation from a Windows install, in case it helps this along: both failure modes this PR handles are real, and they are independent of each other.

I root-caused the same bug from the issue side (#513, #417, #514) and checked it against a live ledger here:

- All 58 records in `~/.codex/external_agent_session_imports.json` carry the `\\?\` extended-length prefix, so `record.source_path === fs.realpathSync(...)` can never match. That alone is enough to return `null`.
- 12 of the 22 source transcripts still on disk no longer hash to their recorded `content_sha256`. That is expected rather than anomalous: the Claude transcript is append-only and keeps growing while the transfer runs, so the hash computed at lookup time trails the one Codex recorded at import.

The second point is why the snapshot-then-fall-back-to-appended-records shape here looks right to me. A path-only fix is not sufficient — I wrote one myself (#551) before I understood the hash half, and it would not have fixed this. It is closed and unmerged.

One small gap, as a review note rather than a blocker. `normalizeLedgerSourcePath()` strips `\\?\` but not the `\\?\UNC\` variant, so a UNC path normalises to `unc/server/share/...` while `realpathSync` yields `//server/share/...` and the two still miss. One extra branch ahead of the existing one covers it:

```js
let normalized = value.startsWith("\\\\?\\UNC\\")
  ? `\\\\${value.slice(8)}`
  : value.startsWith("\\\\?\\")
    ? value.slice(4)
    : value;
```

Everything after that line stays as you have it. This only affects roaming/UNC home directories, and the appended-records fallback would rescue those anyway, so it is genuinely minor — happy to send it as a follow-up rather than churn this PR.

latent-sre:
Independent live confirmation that the notification/history result already carries the authoritative imported thread ID, so the ledger can be a compatibility fallback rather than the primary success path.

**Environment**

- Windows
- `codex@openai-codex` 1.0.6
- stable `codex-cli 0.147.0`
- Node 24.16.0

A bare `/codex:transfer` correctly identified the active Claude transcript. Codex completed the import, but the companion then emitted the false `did not record an imported thread` error.

The expected ledger mismatches were present:

- the ledger stored the source as a Rust-canonicalized `\\?\C:\...` path while Node's `fs.realpathSync()` returned `C:\...`;
- the live transcript's hash after completion differed from the snapshot hash recorded by Codex.

More importantly, a read-only `externalAgentConfig/import/readHistories` request returned the completed import with:

```json
{
  "successes": [
    {
      "itemType": "SESSIONS",
      "source": "<redacted Claude transcript>",
      "target": "<imported thread UUID>",
      "title": "<imported session title>"
    }
  ],
  "failures": []
}
```

Using that exact `successes[].target` value with `thread/read` returned the stored thread in the correct working directory with its title intact. With `includeTurns: true`, it contained 30 turns and 694 items. This verifies on stable 0.147.0 that `target` is not merely metadata: it is the resumable imported thread ID.

That suggests a stronger primary path than post-import ledger inference:

1. retain the `importId` returned by `externalAgentConfig/import`;
2. correlate the matching `externalAgentConfig/import/completed` notification;
3. surface any SESSIONS `failures[]`;
4. require one unambiguous SESSIONS success with a non-empty `target`;
5. use that `target` as `threadId`;
6. keep this PR's normalized, append-aware ledger lookup only as a fallback for older Codex versions or incomplete notification payloads.

This would also remove the live-file hash TOCTOU and ledger-write timing dependency from the normal success path. Issue #618 independently proposed this notification-first direction; this comment adds a live stable-0.147.0 result and a successful `thread/read` verification directly to this PR.

Issue #309, opened 2026-05-09
[BUG] adversarial-review / review fail with HTTP 400 ('gpt-5.5 requires newer Codex') on CLI 0.130.0


jianjettfu-oss:
## Symptom

`/codex:adversarial-review` (and presumably `/codex:review`) fails immediately with:

```
[codex] Codex error: {"type":"error","status":400,"error":{"type":"invalid_request_error",
  "message":"The 'gpt-5.5' model requires a newer version of Codex. Please upgrade to the latest app or CLI and try again."}}
[codex] Turn failed.
```

Result file shows: `Codex did not return valid structured JSON. — Parse error: <same 400 above>`.

## Environment

- Codex CLI: `0.120.0` → upgraded to `0.130.0` (latest as of 2026-05-09). Same failure on both.
- Plugin: `codex@openai-codex` v1.0.4 (latest, reinstalled cleanly to confirm).
- Auth: ChatGPT login active (verified via `node scripts/codex-companion.mjs setup --json` → `ready: true`, `auth.loggedIn: true`).
- `~/.codex/config.toml`: `model = "gpt-5.5"` (default after CLI upgrade).
- **Direct `codex review --base <ref>` works correctly with same model and same config** — the issue is plugin-specific.

## Reproduction

1. Codex CLI 0.130.0, ChatGPT login, `model = \"gpt-5.5\"` in config.toml
2. Run `/codex:adversarial-review --base <ref> --background` (or `--wait`)
3. Plugin task-thread returns HTTP 400 from Codex API

## Root Cause Hypothesis

`scripts/codex-companion.mjs adversarial-review` (and `review`) only accept \`[--wait|--background|--base|--scope] [focus]\` — no \`--model\` flag (only \`task\` subcommand accepts \`--model\`). The companion script falls back to the config.toml default model, which on a freshly-upgraded CLI is \`gpt-5.5\`. The plugin's task-thread protocol layer apparently doesn't advertise sufficient version to the API for gpt-5.5 dispatch — direct \`codex review --base <ref>\` (CLI-native, not task-thread) succeeds with the same model and config.

## Workaround

Edit \`~/.codex/config.toml\`: \`model = \"gpt-5.4\"\` (older but supported by plugin's task-thread protocol). Revert after invocation.

## Suggested Fix

Either:
- Expose \`--model <m>\` flag on \`review\` and \`adversarial-review\` subcommands (mirror what \`task\` already does), so users can pin a known-good model per-invocation, or
[...]

PR #471, opened 2026-07-10
Support GPT-5.6 models and refresh stale brokers

### LINKED ISSUE #468: Current Plugin does not support gpt-5.6 model family
ERROR: {"type":"error","status":400,"error":{"type":"invalid_request_error","message":"The 'gpt-5.6-terra' model requires a newer version of Codex. Please upgrade to the latest app or CLI and try again."}}
ERROR: {"type":"error","status":400,"error":{"type":"invalid_request_error","message":"The 'gpt-5.6-terra' model requires a newer version of Codex. Please upgrade to the latest app or CLI and try again."}}

alexandrereyes:
## Summary

- support `max` and `ultra` reasoning efforts across task, review, and adversarial-review flows
- validate known OpenAI model/effort combinations from the app-server model catalog while preserving fallback behavior for older CLIs, custom providers, and unknown models
- reject unsupported combinations such as `gpt-5.6-luna` + `ultra` locally with a clear error
- validate inherited model/effort configuration, including the catalog default model, before creating persistent task threads
- recycle shared app-server brokers when the plugin version or Codex CLI version changes, including legacy broker state
- preserve reachable stale brokers that report an active turn, keep cancellation interrupts on that broker, let unrelated new work use a direct app-server, and guard broker shutdown against active requests
- serialize broker lifecycle changes across processes and recover abandoned locks
- expose and document `--model` / `--effort` for review commands
- make `task --write` and write-mode resumes request the official `danger-full-access` sandbox so localhost services and browser Mach ports remain available
- bump release metadata from 1.0.6 to 1.0.7

## Root cause

The reported false upgrade error was not caused by `clientInfo.version` or `serviceName`. Controlled probes against Codex CLI 0.144.0 returned the same GPT-5.6 catalog for plugin client version 1.0.6 and CLI client version 0.144.0, and a real plugin-path GPT-5.6 turn succeeded with the existing service name.

The failure occurs when the plugin reuses a broker that was started before the Codex CLI upgrade. That broker keeps its original `codex app-server` child in memory, so a locally installed 0.144.0 CLI can still route plugin turns through an older app-server. Broker state now records both plugin and CLI versions and is invalidated atomically when either changes or when legacy state has no runtime identity.

A separate sandbox issue affected write-capable tasks: the companion explicitly sent `workspace-write`, which can deny localhost networking and Chromium Mach ports. The `--write` path now sends the app-server's supported `danger-full-access` mode for both new and resumed tasks.

Review follow-up also hardens lifecycle edges: inherited and default-model selections are rejected before persistent thread creation; stale broker replacement probes for active work before shutdown; cancellation retains the stale broker when it owns the active turn; and idle brokers still rotate normally.

## Validation

- `npm ci`
- `npm run check-version`
- `npm run build` (including app-server binding generation)
- `npm test` — 114/114 passing
- deterministic catalog tests for Sol/Terra/Luna, including Ultra support, Luna rejection, and default-model resolution
- integration coverage for foreground, resume, background, review, adversarial review, inherited/default configuration, custom providers, older CLIs without `model/list`, stale CLI/plugin versions, legacy broker state, concurrent startup, abandoned-lock recovery, active stale brokers, and stale-broker cancellation
- regression coverage for unrestricted sandbox selection on new and resumed `--write` tasks
- real read-only plugin-path smokes on Codex CLI 0.144.0:
  - `gpt-5.6-sol` / `high`
  - `gpt-5.6-terra` / `max`
  - `gpt-5.6-luna` / `high`
- Ultra was covered deterministically and was not used for real smokes
- independent diff review completed; all findings were fixed and revalidated

Fixes #468.
Related to #309, which reports the same false “requires a newer version” symptom and direct-CLI/plugin-runtime mismatch; this PR fixes the stale shared-broker path and also exposes review model/effort selection.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -5,13 +5,13 @@
   },
   "metadata": {
     "description": "Codex plugins to use in Claude Code for delegation and code review.",
-    "version": "1.0.6"
+    "version": "1.0.7"
   },
   "plugins": [
     {
       "name": "codex",
       "description": "Use Codex from Claude Code to review code or delegate tasks.",
-      "version": "1.0.6",
+      "version": "1.0.7",
       "author": {
         "name": "OpenAI"
       },
--- a/README.md
+++ b/README.md
@@ -86,7 +86,7 @@ Use it when you want:
 - a review of your current uncommitted changes
 - a review of your branch compared to a base branch like `main`
 
-Use `--base <ref>` for branch review. It also supports `--wait` and `--background`. It is not steerable and does not take custom focus text. Use [`/codex:adversarial-review`](#codexadversarial-review) when you want to challenge a specific decision or risk area.
+Use `--base <ref>` for branch review. It also supports `--wait`, `--background`, `--model`, and `--effort`. It is not steerable and does not take custom focus text. Use [`/codex:adversarial-review`](#codexadversarial-review) when you want to challenge a specific decision or risk area.
 
 Examples:
 
@@ -94,6 +94,7 @@ Examples:
 /codex:review
 /codex:review --base main
 /codex:review --background
+/codex:review --model gpt-5.6-sol --effort max
 ```
 
 This command is read-only and will not perform any changes. When run in the background you can use [`/codex:status`](#codexstatus) to check on the progress and [`/codex:cancel`](#codexcancel) to cancel the ongoing task.
@@ -105,7 +106,7 @@ Runs a **steerable** review that questions the chosen implementation and design.
 It can be used to pressure-test assumptions, tradeoffs, failure modes, and whether a different approach would have been safer or simpler.
 
 It uses the same review target selection as `/codex:review`, including `--base <ref>` for branch review.
-It also supports `--wait` and `--background`. Unlike `/codex:review`, it can take extra focus text after the flags.
+It also supports `--wait`, `--background`, `--model`, and `--effort`. Unlike `/codex:review`, it can take extra focus text after the flags.
 
 Use it when you want:
 
@@ -119,6 +120,7 @@ Examples:
 /codex:adversarial-review
 /codex:adversarial-review --base main challenge whether this was the right caching and retry design
 /codex:adversarial-review --background look for race conditions and question the chosen approach
+/codex:adversarial-review --model gpt-5.6-terra --effort xhigh challenge the retry design
 ```
 
 This command is read-only. It does not fix code.
@@ -160,6 +162,8 @@ Ask Codex to redesign the database connection to be more resilient.
 
 - if you do not pass `--model` or `--effort`, Codex chooses its own defaults.
 - if you say `spark`, the plugin maps that to `gpt-5.3-codex-spark`
+- reasoning efforts are `none`, `minimal`, `low`, `medium`, `high`, `xhigh`, `max`, and `ultra`; the current Codex model catalog is used to reject unsupported combinations such as `gpt-5.6-luna` with `ultra`
+- model names are otherwise passed through, so custom providers and newly released models are not blocked by a plugin allowlist
 - follow-up rescue requests can continue the latest Codex task in the repo
 
 ### `/codex:transfer`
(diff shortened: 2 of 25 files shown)


gabimoncha:
up up

gbertb:
we're still waiting for this!

alexandrereyes:
@codex review

axisrow:
Independently verified the effort path here on `codex-cli 0.144.1`: a raw `thread/start` with `config: { model_reasoning_effort: "high" }` comes back with `reasoningEffort: "high"` (no error), which matches the approach this PR takes in `codex.mjs` (`config: options.effort ? { model_reasoning_effort: options.effort } : null`). So the per-thread config route is confirmed working against a live app-server, not just the fake fixture. 👍

One thing worth adding to the description: this also closes #476 (`review` / `adversarial-review` silently ignoring `--effort`). The root cause there is the old code trying to pass effort as a top-level field on `turn/start` / `review/start`, which the server drops — routing it through `thread/start.config` (as you do here) is exactly the fix. Might be worth listing `#476` alongside `#468`/`#309` in the fixes.

One honest question on the boot-time class of failures (#309, and the closely related #408): when the app-server's *default* model is itself unsupported on the account (e.g. a `gpt-5.3-codex` default on ChatGPT/Plus), does `thread/start.config` still recover it? The failure in those reports happens before the first `thread/start` lands, so I wasn't sure whether the per-thread override gets a chance to apply. I can't reproduce it myself — my account resolves the default fine — so this is a question, not a claim. If you smoke-tested that path on a Plus account, it'd be great to note it, since it's the one case where a spawn-time `-c model=...` might still be needed as a fallback.

Pinging @hiirott (whose #408 tackles this exact boot-time case via a spawn-time `-c model=...`) and @jianjettfu-oss (#309 reporter) — you two have hit this directly, so your take on whether the per-thread config route covers it would be valuable.

Either way, nice work — the version-based broker invalidation is the right root-cause fix for the stale-broker half of #468.

axisrow:
Quick question on `a6e501a` ("Fix unrestricted sandbox for write tasks"): it maps `task --write` to `danger-full-access` instead of the previous `workspace-write`. Was that intentional? Curious what use-case motivated it — does Codex routinely need to edit outside the repo on a `--write` task (global package installs, `~/.config`, etc.)?

Asking because under `approvalPolicy: "never"` this removes the last containment boundary — an ordinary-looking `--write` flag silently grants full filesystem authority of the user account (can read/modify/delete `~/.ssh`, other `~/Projects/*`, etc.). An adversarial reviewer flagged it as a privilege expansion, and I'm inclined to restore `workspace-write` for `--write` (keeping `danger-full-access` behind an explicit opt-in if broad access is genuinely needed sometimes). But before I do, wanted to check whether the unrestricted mode here was a deliberate workaround for something `workspace-write` couldn't do — if so, maybe the fix belongs in a separate flag rather than reverting.

axisrow:
Small UX suggestion: this PR documents the full model ids (`gpt-5.6-sol` / `terra` / `luna`) but leaves `MODEL_ALIASES` with only `spark → gpt-5.3-codex-spark`, so users have to remember the exact id to pass `--model gpt-5.6-sol`. Adding short aliases by the same pattern — `sol` / `terra` / `luna → gpt-5.6-{sol,terra,luna}` — would let `--model sol` work out of the box, consistent with how `spark` already behaves.

I prototyped this in a fork (axisrow/codex-plugin-cc@94e2b90) — it's a one-map change plus updating the `argument-hint` strings in `printUsage` and the command `.md` files. Happy to open a focused PR against this branch if it's welcome, or fold it in here if preferred.

axisrow:
Heads up — the "no focus text in `/codex:review`" constraint (which this PR's native-review mapping relies on) has an interface-parity and i18n cost: `/codex:adversarial-review` accepts focus text while `/codex:review` rejects it wholesale, and a user who writes the model/effort value as a natural phrase in a non-English language (e.g. `/codex:review モデル ソル エフォート エックスハイ`) hits the rejection because there's no positional/script-agnostic path to express `--model`/`--effort`. Filed as #522. If this PR (or a follow-up) adds leniency — e.g. recognizing model/effort intent in any language before the focus-text check — it would close that gap alongside the 5.6 support.

PR #475, opened 2026-07-11
Add task --resume-thread <id> for explicit thread resume

Guardiannw:
## Summary

Adds `task --resume-thread <id>` so callers can resume an **explicit** Codex thread instead of the newest tracked one.

`--resume` / `--resume-last` picks the most recent tracked task thread, which is ambiguous for orchestrators driving several concurrent Codex jobs in one repository: which thread gets resumed depends on job-store state and session scoping rather than the caller's intent. The app-server layer already supports targeting a thread (`runAppServerTurn`'s `resumeThreadId`); this exposes it on the `task` subcommand so a caller that recorded a `threadId` from a previous run (`task --json`) can continue that exact conversation deterministically.

## Changes

- `task --resume-thread <id>` resumes the given thread (foreground and `--background`; the worker path serializes `resumeThreadId` through the stored request).
- Conflicts with `--resume` / `--resume-last` / `--fresh` are rejected with a clear error.
- A prompt remains optional, matching `--resume-last` semantics (the default continue-prompt is used when none is given).

## Tests

- `task --resume-thread resumes the specified thread` — round 1 `task --json` captures a `threadId`, round 2 resumes it and asserts the same thread + continued output.
- `task --resume-thread rejects --resume-last and --fresh`.
- Full existing suite green (`npm test`): 93 passing.

## Motivation

We drive Codex as a multi-round peer reviewer from Claude Code. Round 1 records the `threadId` from `task --json`; subsequent rounds need to resume *that* thread. Today the only CLI path is `--resume-last`, which is a heuristic; under any concurrency it can resume the wrong conversation. This flag makes the common "continue the thread I just started" case explicit and deterministic, using machinery the app-server already exposes.

### DIFF
--- a/plugins/codex/scripts/codex-companion.mjs
+++ b/plugins/codex/scripts/codex-companion.mjs
@@ -79,7 +79,7 @@ function printUsage() {
       "  node scripts/codex-companion.mjs setup [--enable-review-gate|--disable-review-gate] [--json]",
       "  node scripts/codex-companion.mjs review [--wait|--background] [--base <ref>] [--scope <auto|working-tree|branch>]",
       "  node scripts/codex-companion.mjs adversarial-review [--wait|--background] [--base <ref>] [--scope <auto|working-tree|branch>] [focus text]",
-      "  node scripts/codex-companion.mjs task [--background] [--write] [--resume-last|--resume|--fresh] [--model <model|spark>] [--effort <none|minimal|low|medium|high|xhigh>] [prompt]",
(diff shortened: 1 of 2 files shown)


zknpr:
Heads up: this was useful enough for multi-job orchestration that #346 cherry-picked it as a808632 (your authorship preserved). It integrated cleanly with #346's stored-request worker flow — `resumeThreadId` now rides the persisted request, so foreground, `--background`, and the detached worker all resume the exact thread. Thanks for the clean patch.

Issue #476, opened 2026-07-11
review / adversarial-review silently ignore reasoning effort — --effort unparsed, and turn/start effort omitted on the adversarial path


ezraaaa:
## Summary

`review` and `adversarial-review` silently ignore reasoning effort. `--effort` is not a recognized flag on either command, and `adversarial-review` omits `effort` when calling `runAppServerTurn` — even though the underlying `turn/start` request already accepts it and `task` threads it through correctly.

The net effect: **the review commands always run at whatever `model_reasoning_effort` sits in `~/.codex/config.toml`, and there is no way to raise effort for a review specifically.** For users who treat the adversarial review as their pre-merge quality gate, this is the one place extra reasoning is most worth paying for, and it's the one place they can't buy it.

This is also silent — no "unknown flag" error. `... adversarial-review --effort xhigh` is accepted and runs at the config default, which reads as if it worked.

## Evidence (v1.0.6)

`--model` is threaded end-to-end; `--effort` is missing at every touchpoint on this path.

**1. `--effort` is not parsed** — `scripts/codex-companion.mjs:714`

```js
async function handleReviewCommand(argv, config) {
  const { options, positionals } = parseCommandInput(argv, {
    valueOptions: ["base", "scope", "model", "cwd"],   // <-- no "effort"
    booleanOptions: ["json", "background", "wait"],
    aliasMap: { m: "model" }
  });
```

Compare `handleTask` (`:764`), which has `valueOptions: ["model", "effort", "cwd", "prompt-file"]` and normalizes via `normalizeReasoningEffort`.

**2. The call site doesn't forward it** — `scripts/codex-companion.mjs:742`

```js
executeReviewRun({
  cwd,
  base: options.base,
  scope: options.scope,
  model: options.model,     // <-- no effort
  ...
})
```

**3. `adversarial-review` drops it into a call that supports it** — `scripts/codex-companion.mjs:411`

```js
const result = await runAppServerTurn(context.repoRoot, {
  prompt,
  model: request.model,
  sandbox: "read-only",     // <-- no effort
  outputSchema: readOutputSchema(REVIEW_SCHEMA),
  onProgress: request.onProgress
});
```

…but `runAppServerTurn` (`scripts/lib/codex.mjs:1139`) already forwards effort to `turn/start`:

```js
client.request("turn/start", {
  threadId,
  input: buildTurnInput(prompt),
  model: options.model ?? null,
  effort: options.effort ?? null,   // always null on the review path
[...]

PR #477, opened 2026-07-11
Add model and effort flags to review commands

axisrow:
Propagate `--model` and `--effort` to `/codex:review` and `/codex:adversarial-review`, matching the existing `task` behavior.

Validation:
- `node --test tests/commands.test.mjs`
- `node --test tests/runtime.test.mjs`

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,13 +86,14 @@ Use it when you want:
 - a review of your current uncommitted changes
 - a review of your branch compared to a base branch like `main`
 
-Use `--base <ref>` for branch review. It also supports `--wait` and `--background`. It is not steerable and does not take custom focus text. Use [`/codex:adversarial-review`](#codexadversarial-review) when you want to challenge a specific decision or risk area.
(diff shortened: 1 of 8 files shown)


axisrow:
Closing this as the wrong fix.

The approach here — forwarding `model`/`effort` into the `review/start` RPC params — doesn't actually make the flags work. Probing the real app-server (codex-cli 0.144.1), `review/start` accepts `model`/`effort` without error but the Rust side drops unknown fields (serde, no `deny_unknown_fields`), and `ReviewStartParams` has no `model`/`effort` field in the first place (`TS2353` on a direct type-probe). The `model` field is inert on the server across `thread/start`/`turn/start`/`review/start` — the effective override is `-c model="…"` at spawn time (#408), not an RPC field.

For `effort`: `adversarial-review` can be fixed via `turn/start` (its params include `effort`) per #476; native `review` has no effort field in the protocol, so it's not fixable on the plugin side without the #408-style spawn override.

Left detailed notes + the server-probe evidence on #476 and #408. Folding into those instead of keeping a separate incompatible change open.

Issue #487, opened 2026-07-12
Windows: broker child processes leak and stale broker.json poisons setup tests on persistent machines


nategarelik:
Two related Windows process-hygiene gaps found while running the suite repeatedly on a persistent (non-CI) Windows 11 box. Both are pre-existing (independent of any local change) and reproducible.

## 1. Stale broker.json poisons setup-family tests

`%TEMP%\codex-companion\<repo-hash>\broker.json` persists across runs and can point at a dead broker PID / stale named pipe. The `setup`-family tests key shared state on the real repo path (the test fixture's `buildEnv()` inherits `process.env` unmodified rather than using an ephemeral fixture dir), so a stale pointer makes every subsequent run report false "not ready" results. Symptom: ~6 spurious failures that vanish permanently after deleting that one directory. On ephemeral CI this is masked because the temp dir is always fresh.

Suggested fix: have the setup tests use an isolated $TEMP/HOME per run, or validate/ignore a broker.json whose PID is dead before treating it as authoritative.

## 2. Broker child processes leak on Windows

Repeated test runs accumulate orphaned `app-server-broker.mjs` node children (observed ~200 after a day of runs), each holding its temp workspace open. Same family as the known `taskkill /T /F` flake ("operation attempted is not supported" on a nested child). Correctness impact is low (each is isolated to a random temp workspace), but it blocks `git worktree remove` on any worktree a broker still holds open, and exhausts the process table over time.

Environment: Windows 11, codex-cli 0.144.1, plugin at current main. Happy to provide fuller repro details.

axisrow:
This issue conflates three axes: (1) the **broker leak** — stillborn/orphaned brokers = scenario 3 in the #163 map, *not* covered by PR #490 (the idle-timer never reaches stillborn brokers that never bind a socket); (2) **stale `broker.json` poisoning tests** — a separate validation problem (same PID-liveness family noted in #478); (3) **env-inheritance / directory accumulation** — the fixture spreads `process.env` (`fake-codex-fixture.mjs:658`), so tests write into the real plugin data dir, an orthogonal axis — see my stats in #163. Worth splitting. PR #490 closes only the "live-but-idle broker" subset.

Issue #380, opened 2026-06-18
SessionEnd cleanup fails when review cwd ≠ session cwd — broker.json looked up by cwd-hash, leaving orphan brokers even on graceful /quit


hongsu:
## Summary

`handleSessionEnd` cleans up the broker by calling `loadBrokerSession(input.cwd)`, but `broker.json` is stored under a **cwd-derived hash** (`resolveStateDir(cwd)`). When the broker was spawned for a **different cwd** than the session's cwd, the lookup misses, `endpoint` is null, no `broker/shutdown` is sent, and the broker is orphaned — **even when the session exits gracefully (`/quit`, SIGTERM)**.

This is distinct from #108 (which attributes orphans to the hook *not firing* on crash/kill). Here the hook fires normally (exit 0) but **cannot locate the broker**.

## Why this happens in normal use

The broker cwd is decided by `resolveCommandCwd` (`codex-companion.mjs:149`):

```js
options.cwd ? path.resolve(process.cwd(), options.cwd) : process.cwd()
```

→ flows to `connect(cwd)` → `ensureBrokerSession(cwd)` → broker `--cwd` and `broker.json` location.

A very common workflow breaks this:
- Session starts in the **main tree** (session cwd = `/repo`).
- A review runs against a **git worktree** — either `cd worktree && node companion …` or `node companion … --cwd worktree` → broker cwd = `worktree`.
- `/quit` → SessionEnd `input.cwd` = `/repo` → `loadBrokerSession(/repo)` looks at the `/repo` hash, never finds the `worktree` broker.json → broker survives.

Reviewing a worktree's *uncommitted* working tree **requires** broker cwd = worktree, so for a main-tree session this mismatch is structural, not accidental.

## Reproduction (verified)

```bash
# broker registered for cwd=A
ensureBrokerSession("/tmp/probe")            # broker --cwd=/tmp/probe, alive

# SessionEnd with matching cwd  -> broker is killed (OK)
echo '{"cwd":"/tmp/probe","session_id":"s1","hook_event_name":"SessionEnd"}' \
  | node session-lifecycle-hook.mjs SessionEnd   # exit 0, broker DEAD

# SessionEnd with mismatched cwd -> hook exits 0 but broker SURVIVES
echo '{"cwd":"/tmp","session_id":"s2","hook_event_name":"SessionEnd"}' \
  | node session-lifecycle-hook.mjs SessionEnd   # exit 0, broker ALIVE  <-- bug
```

[...]

Issue #288, opened 2026-05-02
`sendBrokerShutdown` has no timeout — SessionEnd hook can hang indefinitely


SmidaOps:
## Summary

`sendBrokerShutdown` in `plugins/codex/scripts/lib/broker-lifecycle.mjs` awaits a Promise that resolves only on socket `data`, `error`, or `close` events. If the broker accepts the connection but stops responding mid-RPC (no data, no error, no close), the Promise never resolves and the SessionEnd hook hangs indefinitely.

## Affected code

`plugins/codex/scripts/lib/broker-lifecycle.mjs:43-57` (v1.0.3)

```javascript
export async function sendBrokerShutdown(endpoint) {
  await new Promise((resolve) => {
    const socket = connectToEndpoint(endpoint);
    socket.setEncoding("utf8");
    socket.on("connect", () => {
      socket.write(`${JSON.stringify({ id: 1, method: "broker/shutdown", params: {} })}\n`);
    });
    socket.on("data", () => {
      socket.end();
      resolve();
    });
    socket.on("error", resolve);
    socket.on("close", resolve);
  });
}
```

## Asymmetry with startup path

`waitForBrokerEndpoint` in the same file (line 24) explicitly takes `timeoutMs = 2000` and bounds its wait loop. Startup path is timed; shutdown path isn't. Same module — looks like an oversight rather than intent.

## Observed symptoms

Running Claude Code in unattended (cron/launchd) invocations on macOS:

- Multiple sessions hung at SessionEnd after content rendered cleanly. One held its lock 19.5 hours before something external (a new Claude Code session creating a fresh broker socket) unwedged it.
- In a separate session the harness's hook timeout fired with `Hook cancelled` and exit 1 — content was already delivered, but the misleading exit code surfaced as a failure in launchd telemetry.
- Correlation suggests macOS DarkWake during long-running SSE streams wedges the broker. We mitigated with `caffeinate -ims` on the wrapper, but the shutdown await still hangs when a wedge does happen.

## Proposed fix

Mirror the timeout pattern used in `waitForBrokerEndpoint`:

```javascript
export async function sendBrokerShutdown(endpoint, timeoutMs = 5000) {
  await new Promise((resolve) => {
    const socket = connectToEndpoint(endpoint);
    let settled = false;
    const finish = () => {
      if (settled) return;
      settled = true;
      clearTimeout(timer);
      try { socket.destroy(); } catch {}
      resolve();
    };
[...]

Issue #163, opened 2026-04-07
Test suite leaks broker processes — 158 orphans found


peterdrier:
## Problem

Running the codex plugin test suite spawns `app-server-broker.mjs` processes for each test workspace (`/tmp/codex-plugin-test-*`), but these processes are never cleaned up after tests complete.

Found 158 stale broker processes on a development machine, each holding a `/tmp/cxc-*` socket directory. All were from old test runs (dated Apr 3–6), pointing at temp dirs that may or may not still exist.

## Evidence

```
$ ps aux | grep 'app-server-broker.mjs' | grep -v grep | wc -l
167

$ ps aux | grep 'app-server-broker.mjs' | grep -v grep | grep -oP '(?<=--cwd )\S+' | sort | uniq -c | sort -rn | head -5
1 /tmp/codex-plugin-test-zxrQqX
1 /tmp/codex-plugin-test-znwec8
... (158 test dirs total)

$ ls -d /tmp/cxc-* | wc -l
168
```

## Expected behavior

Test teardown should kill broker processes it spawned (or the broker should self-terminate when its cwd disappears / after idle timeout).

## Impact

- Resource leak: 167 node processes × ~35MB RSS = ~5.8GB wasted
- Socket dir accumulation in /tmp
- May contribute to Codex review failures when the system is under resource pressure

brettimus:
also did some spring cleaning on my remote box and hit this. 159 codex app-server processes, never cleaned up. 

claude's opinion of the issue (yet to verify manually): 

***

The leak is in SpawnedCodexAppServerClient spawning codex app-server processes that never get cleaned
  up.

  The flow:
  1. Each Claude session with the Codex plugin calls CodexAppServerClient.connect() (line 332 of
  app-server.mjs)
  2. This calls ensureBrokerSession() which spawns a detached, unref'd broker process
  (broker-lifecycle.mjs:61-68) — detached: true + child.unref() means Node won't wait for it to exit
  3. The broker in turn spawns codex app-server as a child (app-server.mjs:189)
  4. Cleanup depends on SessionEnd hook firing (session-lifecycle-hook.mjs:81), which calls
  sendBrokerShutdown() then teardownBrokerSession()

  The problem: When a Claude session dies uncleanly (killed, crashes, terminal closed, OOM'd), the
  SessionEnd hook never fires. The broker and its codex app-server child are detached and orphaned —
[...]

PR #490, opened 2026-07-13
Stop orphaned Codex companion brokers

ckfchampion:
## Problem

The shared app-server broker outlives the Claude Code session that spawned it. When a session exits without a clean teardown, the broker is reparented to `init` and keeps listening forever, holding its `codex app-server` child open with it. Nothing ever reaps them, so they accumulate for as long as the machine stays up.

On one developer machine this had reached **8 orphaned brokers, ~1.3 GB RSS**, the oldest alive for 41 hours — every one with `PPID=1`, i.e. no session attached. Running the plugin's own test suite makes it worse, since each run leaves its helper brokers behind too.

## Fix

Shut the broker down after 15 minutes with no connected clients.

- Foreground and background jobs hold a broker connection for their whole lifetime, so an **active job always keeps its broker alive** — the timer only ever fires on a broker nobody is using.
- `CODEX_COMPANION_BROKER_IDLE_TIMEOUT_MS` overrides the window; `0` disables the timer entirely.
- No client-side change is needed: `ensureBrokerSession()` already probes the endpoint and respawns when it finds none, so an idle-exited broker is transparently recreated on the next call. A successful probe also resets the idle timer, so there is no window where a client can connect to a broker that is about to exit.

Two shutdown-path corrections come with it:

- Close the listener **before** awaiting async child cleanup. Previously a reconnect could slip into the async shutdown window and inherit a closing app-server client.
- Reject connections that are delivered after `server.close()`, rather than adding them to the live socket set.

## Tests

Adds `tests/broker-idle.test.mjs`, covering: exit + runtime-file cleanup after the last client disconnects; idle shutdown deferred while a client is still connected; the timeout being disabled via `0`; and clean shutdown while new clients race to connect.

`tests/fake-codex-fixture.mjs` now sets a 2s idle window so the suite stops leaving detached helpers behind.

Full suite: 91 passing. The 4 failures on this branch (`status`/`result`/`resolveStateDir`) reproduce identically on an unmodified `main`, so they are pre-existing and unrelated.

### DIFF
--- a/plugins/codex/CHANGELOG.md
+++ b/plugins/codex/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## Unreleased
+
+- Stop an orphaned shared Codex app-server broker after 15 minutes with no connected clients. Active foreground and background jobs keep their broker connection open, and the timeout can be configured with `CODEX_COMPANION_BROKER_IDLE_TIMEOUT_MS` (`0` disables the safety timer).
+- Close the broker listener before asynchronous child cleanup and safely reject reconnects already queued during shutdown.
+
 ## 1.0.0
 
 - Initial version of the Codex plugin for Claude Code
--- a/plugins/codex/scripts/app-server-broker.mjs
+++ b/plugins/codex/scripts/app-server-broker.mjs
@@ -10,6 +10,21 @@ import { BROKER_BUSY_RPC_CODE, CodexAppServerClient } from "./lib/app-server.mjs
 import { parseBrokerEndpoint } from "./lib/broker-endpoint.mjs";
 
 const STREAMING_METHODS = new Set(["turn/start", "review/start", "thread/compact/start"]);
+const BROKER_IDLE_TIMEOUT_ENV = "CODEX_COMPANION_BROKER_IDLE_TIMEOUT_MS";
+const DEFAULT_BROKER_IDLE_TIMEOUT_MS = 15 * 60 * 1000;
+
+function resolveIdleTimeoutMs(env = process.env) {
+  const rawValue = env[BROKER_IDLE_TIMEOUT_ENV];
+  if (rawValue == null || rawValue === "") {
+    return DEFAULT_BROKER_IDLE_TIMEOUT_MS;
+  }
+
+  const parsed = Number(rawValue);
+  if (!Number.isFinite(parsed) || parsed < 0) {
+    return DEFAULT_BROKER_IDLE_TIMEOUT_MS;
+  }
+  return Math.floor(parsed);
+}
 
 function buildStreamThreadIds(method, params, result) {
   const threadIds = new Set();
@@ -70,6 +85,16 @@ async function main() {
   let activeStreamSocket = null;
   let activeStreamThreadIds = null;
   const sockets = new Set();
+  const idleTimeoutMs = resolveIdleTimeoutMs();
+  let idleTimer = null;
+  let shuttingDown = false;
+
+  function cancelIdleShutdown() {
+    if (idleTimer) {
+      clearTimeout(idleTimer);
+      idleTimer = null;
+    }
+  }
 
   function clearSocketOwnership(socket) {
     if (activeRequestSocket === socket) {
@@ -100,11 +125,20 @@ async function main() {
   }
 
   async function shutdown(server) {
+    if (shuttingDown) {
+      return;
+    }
+    shuttingDown = true;
+    cancelIdleShutdown();
+    // Stop accepting connections before awaiting child cleanup. Otherwise a
+    // reconnect can slip into the async shutdown window and inherit a closing
+    // app-server client.
+    const serverClosed = new Promise((resolve) => server.close(resolve));
     for (const socket of sockets) {
       socket.end();
     }
     await appClient.close().catch(() => {});
-    await new Promise((resolve) => server.close(resolve));
+    await serverClosed;
     if (listenTarget.kind === "unix" && fs.existsSync(listenTarget.path)) {
       fs.unlinkSync(listenTarget.path);
     }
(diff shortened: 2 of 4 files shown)


axisrow:
Nice — this overlaps directly with a race I already caught and patched on the fork: the shutdown window where a pid can be reused. Your "close the listener before awaiting async child cleanup" and "reject connections delivered after `server.close()`" are the right corrections — they're the same class of gap as my PID-reuse guard (only tree-kill by pid when the endpoint probe confirmed it was live).

One thing worth pressure-testing: the interaction between this idle-timer and an active respawn. `ensureBrokerSession()` respawns when the probe finds nothing — but does a successful probe **reset** the idle timer in the same atomic step that a racing idle-shutdown might fire? Your description says "a successful probe also resets the idle timer," but if the reset and the fire are on different ticks, there's a window where a client reconnects to a broker mid-self-shutdown. Your `tests/broker-idle.test.mjs` race case covers new clients connecting during shutdown — does it also cover a probe-reset racing the fire on the same broker? If not, that's the one case I'd add.

axisrow:
A further note on coverage after mapping this against the open broker-leak issues. The idle-timer arms only when `sockets.size === 0 && !activeRequestSocket && !activeStreamSocket` (`app-server-broker.mjs:152`), so it covers exactly one case: a **live broker with no connected clients**. Breaking the leak scenarios down against that:

| Scenario | Covered by #490? | Detail |
|---|---|---|
| CC crashes / terminal closed / unclean exit → detached broker reparents to init | ✅ Yes (the main win) | connection drops → 15 min no clients → `process.exit` |
| `sendBrokerShutdown` hangs on `/quit` | ⚠️ **Partial** | The hang is on the *connect* step of `sendBrokerShutdown` (`broker-lifecycle.mjs:44`), not a broker response. If the abandoned broker is alive-but-idle, #490's timer eventually reaps it (~15 min) — so the orphaned process goes away. What #490 does **not** fix is the `/quit` UX: the SessionEnd hook blocks for that window without a client-side bounded timeout (#288/#245), and a broker wedged with an *active* socket never arms the timer. So: reaps the process, not the hang. |
| Broker spawned but endpoint never came up (stillborn) | ⚠️ **Not covered — and it DOES accumulate, see #487** | I first assumed this "self-cleans on read" via the liveness check in `ensureBrokerSession` (`broker-lifecycle.mjs:107` → teardown + clear), but #487 shows the opposite on Windows: (a) the setup-fixture path trusts `broker.json` as authoritative **without** a liveness check, so a stale pid/named-pipe poisons every subsequent run (~6 spurious failures until manually deleted); (b) ~200 orphaned `app-server-broker` children accumulate over a day, because the teardown path doesn't reap them reliably (`taskkill /T /F` breaks under Git Bash, and `shutdown()` → `appClient.close()` is not a hard guarantee on children). The liveness self-heal exists in `ensureBrokerSession`, but it's path-specific — it is not a substitute for real teardown on the setup/test paths. This is spawn-path robustness, orthogonal to the idle-timer. |

I posted the three-scenario map with the evidence in #163. Net: #490 fully closes scenario 1, partially mitigates scenario 2 (process side only), and is orthogonal to scenario 3 (which genuinely accumulates per #487, despite `ensureBrokerSession`'s liveness self-heal). The earlier "❌ out of scope" framing undersold #490's reach on scenario 2 and was too optimistic about scenario 3's cleanup path — correcting after #487 reminded that the self-heal is path-specific and does not extend to setup/Windows paths.

One behavioral nuance worth documenting: the idle-timer can't tell "client died" from "long pause (>15 min) between turns with the socket closed and not yet reopened" — the latter will trigger a respawn on the next call via `ensureBrokerSession`, which is fine but adds respawn latency. A connection held open across such a pause is of course safe (socket kept alive → idle-timer not armed).

afspies:
Bumping this — we're independently affected and would like to see it land. On one macOS machine orphaned brokers had reached **25 processes / 1.31 GB RSS, oldest 15 days**, 15 of them owned by no live Claude Code session. Idle self-termination is the only mechanism that covers the `SessionEnd`-never-fires cases (SIGKILL, force quit, crash), so it's needed regardless of how #380/#381 lands.

**The listener-first ordering here is the detail worth keeping.** We'd written our own idle timeout that closed the listener *last* — `await appClient.close()` before `server.close()` — and it had exactly the race in #526: a client can connect during the child-cleanup window and inherit a broker that is already tearing down. Kicking off `server.close()` first, so the listener stops accepting immediately while the promise resolves only once connections drain, is the right fix, and we've since adopted this PR's ordering over our own. The `shuttingDown` idempotence flag and the re-check of `sockets.size / activeRequestSocket / activeStreamSocket` inside the timer callback both matter too — the tick-time check alone isn't enough.

On the bot's P2, *"Clear stale broker session on idle exit"* — we hit that same symptom class and can confirm it. Two notes from fixing it:

**1. Only clear `broker.json` if it still names *this* broker.** Between this broker going idle and its shutdown completing, another broker may already have claimed the workspace entry, and clearing unconditionally would strip a live broker's pointer:

```js
try {
  if (loadBrokerSession(cwd)?.endpoint === endpoint) {
    clearBrokerSession(cwd);
  }
} catch { /* never block shutdown on state cleanup */ }
```

**2. The logfile and session dir leak even on a *clean* shutdown.** `shutdown()` unlinks the socket and pidfile but never `broker.log` or the `cxc-*` mkdtemp dir, so every well-behaved exit still leaves a directory behind — we cleared out 17 stale `cxc-*` dirs holding orphaned logfiles. Worth removing both here while you're in this path.

**One suggestion for the idle path specifically.** Unlike a hook-driven teardown, nothing supervises this one, so anything that can stall leaves the broker resident forever — the exact failure being fixed. We bound `appClient.close()` and `server.close()` and put an absolute forced-exit ceiling on every teardown path (SIGTERM, SIGINT, `broker/shutdown`, idle). In practice a broker exits ~1 s after a bare SIGTERM, child included, so these never fire — they're just insurance against the unattended case.

If you do add escalation to the `codex app-server` child on timeout, one trap: **signal its positive PID, not a process group.** The child is spawned without `detached` (`plugins/codex/scripts/lib/app-server.mjs:190`), so it shares the broker's group and is not a group leader. `terminateProcessTree(childPid)` would call `kill(-childPid)`, which hits `ESRCH` — and `process.mjs`'s `terminateProcessTree` has no positive-PID fallback on that branch, so the escalation silently does nothing — or, after PID recycling, signals an unrelated process group. We lost some time to that one.

---

*Written by Claude (Claude Code) on behalf of @afspies, who reviewed and sent it. The measurements and file paths above are from a real machine, not inferred.*

Issue #482, opened 2026-07-12
Hardcoded sandbox values always override config.toml sandbox_mode


cubicj:
## Summary

The plugin sends an explicit `sandbox` value on every `thread/start` / `thread/resume`, so the `sandbox_mode` a user configured in `~/.codex/config.toml` never applies to plugin-launched jobs. A user who sets `sandbox_mode = "workspace-write"` expecting delegated tasks to edit files still gets `read-only` unless every single invocation remembers `--write`; other config keys (`model`, reasoning effort) flow through to plugin jobs as defaults, but `sandbox_mode` is the one setting that silently cannot.

## Evidence (at HEAD db52e28, plugin v1.0.6)

- `buildThreadParams` and `buildResumeParams` always materialize a sandbox: `sandbox: options.sandbox ?? "read-only"` (`plugins/codex/scripts/lib/codex.mjs:68`, `:81`) — the config value can never surface.
- Task path hardcodes the pair: `sandbox: request.write ? "workspace-write" : "read-only"` (`plugins/codex/scripts/codex-companion.mjs:491`).
- Review paths pin `read-only` (`codex.mjs:1012`, `codex-companion.mjs:414`).
- The app-server does fall back to config when the parameter is omitted. Verified live (Codex CLI 0.145.0-alpha.2) with a minimal JSON-RPC client:
  - `thread/start` with `sandbox` omitted → the response `sandbox` matched the `sandbox_mode` configured in config.toml, alongside the config-resolved `model` and `reasoningEffort`.
  - `thread/start` with `sandbox: "read-only"` → response `sandbox: {"type":"readOnly","networkAccess":false}`, overriding config as expected.

So the `?? "read-only"` default and the two hardcoded call sites are the only thing standing between config.toml and plugin jobs.

## Repro

1. Set `sandbox_mode = "workspace-write"` in `~/.codex/config.toml`.
2. Run `node scripts/codex-companion.mjs task --background "any prompt"` without `--write`.
3. The job executes `read-only` regardless of config. There is no flag, config key, or environment variable that lets the configured `sandbox_mode` through.

## Proposed fix

When the caller passes no explicit sandbox-affecting flag, omit `sandbox` from the thread params and let the app-server resolve it from config.toml — the same fallback the app-server already performs for `model` and reasoning effort. `--write` stays as an explicit override, and the review paths can reasonably remain pinned `read-only` if reviews are meant to be side-effect-free; that call is yours.

[...]

Issue #478, opened 2026-07-11
codex-companion background jobs: no turn timeout, no PID-liveness reaping, shell-mangled taskkill (Windows) — jobs stick as 'running' forever


bilclinti:
# Bug report: codex-companion job runner has no turn timeout, no PID-liveness check, and a shell-mangling taskkill call

**Component:** `openai-codex` Claude Code plugin, version `1.0.6`, `scripts/codex-companion.mjs` + `scripts/lib/{tracked-jobs,state,process}.mjs`
**Platform:** Windows (win32), jobs launched via the plugin's `task --background` / `rescue` code path
**Severity:** High — a single wedged turn permanently occupies a scheduler slot and is invisible to `status` as "stuck"; only a process-tree kill + manual `state.json` surgery recovers it.

## Bug 1 — no wall-clock timeout on a tracked job's turn

`runTrackedJob()` (`scripts/lib/tracked-jobs.mjs:142`) does:

```js
const execution = await runner();
```

with no timer, `AbortController`, or any other mechanism racing the runner. If the underlying Codex turn wedges — e.g. a shell child command it spawned mid-turn hangs (stdin wait, dead network call, orphaned handle) — the awaited promise never resolves and the job record never leaves `status: "running"`.

**Evidence:** a real job was observed frozen at `phase: "verifying"`, with its last recorded progress event being a `custom_tool_call` whose output never arrived (the child that should have produced it never returned). By the time it was noticed, the thread had accumulated **~57M input tokens** — a runaway loop re-feeding stale context with no timeout to cut it off.

## Bug 2 — no PID-liveness check when jobs are read back from `state.json`

`listJobs()` (`scripts/lib/state.mjs:149`) returns the `jobs` array from `state.json` verbatim:

```js
export function listJobs(cwd) {
  return loadState(cwd).jobs;
}
```

There is no check anywhere in the read path that a job reported `running`/`queued` still has a live process behind its `pid`. If the worker process dies (terminal closed, Ctrl+C, crash, reboot) without going through the normal completion path in `runTrackedJob`, its `state.json` entry is permanently stuck at `running`. Because `state.json` is an on-disk registry independent of any live process, every subsequent `status`/`task` invocation rehydrates the dead job and the scheduler continues to treat it as occupying a slot — indefinitely, with no self-healing.

## Bug 3 — `taskkill` invoked with a shell layer by default on Windows

[...]

PR #506, opened 2026-07-15
Capture resolved model/effort/sandbox in job records

### LINKED ISSUE #481: Job records never capture the resolved model/effort/sandbox a job ran with
## Summary

`jobs/<id>.json` never records what model, reasoning effort, or sandbox a job actually executed with. Background task jobs persist the raw CLI flags, which are `null` whenever the caller relied on `~/.codex/config.toml` defaults; foreground task jobs and all review jobs record nothing at all. External tooling that wants to display "this job ran gpt-5.4 / medium / read-only" has to reverse-engineer it from Codex CLI rollout files.

## Evidence (at HEAD db52e28, plugin v1.0.6)

- Base job records carry no model/effort fields: `createCompanionJob` (`plugins/codex/scripts/codex-companion.mjs:567-578`) plus `createJobRecord` (`plugins/codex/scripts/lib/tracked-jobs.mjs:60-68`) produce `id`, `kind`, `kindLabel`, `title`, `workspaceRoot`, `jobClass`, `summary`, `write`, `createdAt`, `sessionId`.
- Background tasks additionally persist the raw request — `enqueueBackgroundTask` stores `request` (`codex-companion.mjs:689-698`), built from the CLI flags as passed (`buildTaskRequest`, `codex-companion.mjs:604-614`). If `--model`/`--effort` were omitted, `request.model`/`request.effort` are `null`, while the job really runs with the config.toml defaults resolved by the app-server.
- `runTrackedJob` never adds resolved values to the running or completed record (`tracked-jobs.mjs:142-204`).
- Review jobs and foreground tasks go through `runForegroundCommand` with no `request` persisted at all (`codex-companion.mjs:731-752`, `807-822`).
- The resolved values are available to the plugin: the `thread/start` / `thread/resume` response already carries `model`, `reasoningEffort`, `sandbox`, and `approvalPolicy` — the repo's own fake app-server mirrors this shape (`tests/fake-codex-fixture.mjs:316`).

## Repro

1. `node scripts/codex-companion.mjs task --background "any prompt"` (no `--model`, no `--effort`).
2. Inspect `$CLAUDE_PLUGIN_DATA/state/<slug>-<hash>/jobs/<id>.json`: `request.model` and `request.effort` are `null`; no other field names the model/effort/sandbox.
3. Run `node scripts/codex-companion.mjs review`: the stored job has no model/effort trace at all.
4. The actual values only exist in the Codex CLI rollout `turn_context` entries under `$CODEX_HOME/sessions/YYYY/MM/DD/rollout-<ts>-<threadId>.jsonl`.

## Proposed fix

At thread/turn start, write the resolved `model`, `effort`, and sandbox (as reported by the app-server response) into the job record on all paths — background tasks, foreground tasks, and reviews — e.g. alongside `threadId`/`turnId` in the running record written by `runTrackedJob`.

Happy to follow up with a PR if this direction sounds right.

## Why it matters

The downstream log watcher [@cubicj/codex-watch](https://github.com/cubicj/codex-watch) displays a per-job session-info line. To get it, it joins the job's `threadId` against rollout filenames and parses `turn_context` payloads. That workaround is fragile in ways the plugin can't see: rollout date directories use local time while job timestamps are UTC, resumed threads live under the thread's original date directory, and rollouts are pruned independently of job records. One field written at turn start removes the entire join.

## Related

- #476 — review paths ignore `--effort`; same underlying theme (effort is invisible/unthreaded), different bug.
- #418 — forwarding extra args to the Codex CLI; adjacent, about inputs rather than records.
- #359 — `/codex:observe` structured event stream; complementary observability interest.

## Environment

- Plugin `codex@openai-codex` v1.0.6 (upstream HEAD db52e28)
- Codex CLI 0.145.0-alpha.2
- Claude Code on WSL2 Ubuntu 26.04, Node v24.18.0

cubicj:
## Summary

Job records never capture the model, reasoning effort, or sandbox a job actually ran with (#481). Background task jobs persist the raw CLI request, but those fields are `null` when flags are omitted; foreground tasks and review jobs record nothing at all. The only place the resolved values exist today is the Codex rollout file, and joining a job record to its rollout fails in practice: date directories use local time while job timestamps are UTC, and rollout filenames carry a session id that does not match the job's `threadId` (both hit independently in the issue thread).

## Change

The app-server already reports what it resolved: `thread/start` and `thread/resume` responses echo `model`, `modelProvider`, `reasoningEffort`, and the full `sandbox` policy with config.toml fallback applied, and the plugin currently discards everything but `thread.id`. This PR captures that echo instead of adding any new resolution logic, so it stays truthful regardless of how the pending override-mechanism discussions (#471, #408, #476, #482) land.

- `lib/codex.mjs`: `runAppServerTurn` and `runAppServerReview` build `resolved = { model, modelProvider, reasoningEffort, sandbox }` from the thread response, attach it to the existing "Thread ready" progress event, and include it in the returned result. `turn/start` responses carry no effort field, so an explicit per-turn `--effort` is overlaid onto `resolved` only after the `turn/start` request is accepted — a rejected or failed turn never records a requested effort as if it had run.
- `lib/tracked-jobs.mjs`: the job progress updater patches `resolved` into the job file and state index mid-run, the same way it already handles `threadId`, so failed and cancelled jobs still record what they ran with. The completion write adds `resolved` explicitly because it spreads the in-memory running record.
- `codex-companion.mjs`: each execution object lifts `result.resolved` next to `threadId`/`turnId`; all job kinds share `runTrackedJob`, so foreground tasks, background workers, native review, and adversarial review are covered without per-path changes.

Job records gain one nested field, next to the existing raw `request` fields, so both layers stay visible — requested versus server-confirmed:

```json
"resolved": {
  "model": "gpt-5.4",
  "modelProvider": "openai",
  "reasoningEffort": "medium",
  "sandbox": { "type": "readOnly", "access": { "type": "fullAccess" }, "networkAccess": false }
}
```

Human-readable output is unchanged: renderers read job fields selectively, the progress-event patch rides the same write the `threadId` update already performs, and no log line formats change. `status --json` and `result --json` gain the field additively.

## Verification

- `npm test`: 92/92 passing. New coverage asserts `resolved` in the persisted job JSON for native review, adversarial review, resumed tasks, and the detached background worker (checked mid-run, while still `running`); an explicit `--effort` overriding the thread echo on success; and a `turn/start` failure case with an explicit `--effort` proving the failed record keeps the thread echo rather than the never-applied request.
- `npm run build` (typecheck via `tsconfig.app-server.json`): clean.
- `git diff --check` against the base commit: clean.
- The fake-codex fixture's `thread/start`/`thread/resume` responses now mirror the real app-server's resolved echo (shape confirmed against the generated `ThreadStartResponse`/`ThreadResumeResponse` types and a live JSON-RPC probe against codex-cli 0.144.1).

## Related Issues

Fixes #481


axisrow:
Independently verified this end-to-end against a live app-server on **codex-cli 0.144.4** (your probe was 0.144.1, so this doubles as a newer-CLI check). Everything holds:

- **Echo is real on 0.144.4:** raw `thread/start` returns `model`/`modelProvider`/`reasoningEffort`/`sandbox` (my config resolved to `gpt-5.6-sol`/`medium`); `config: { model_reasoning_effort: "high" }` comes back `reasoningEffort: "high"`.
- **End-to-end on real jobs:** foreground `task` persists `resolved` with live values; `task --effort high` records `resolved.reasoningEffort: "high"` — so the `onResponse` overlay works against the real server, not just the fixture.
- **The echo is truthful:** cross-checked the job record against the Codex rollout — `turn_context` entries carry `reasoning_effort: "high"`, matching the record. This is exactly the join #481 made painful, now unnecessary.
- **Resume:** `task --resume-last` recorded the thread-inherited `high` via the resume echo — state that was previously invisible anywhere in the plugin.
- **Forensics on dead jobs:** SIGKILL'd a background worker mid-run; the orphaned record (stuck at `running`, per #478) still had `resolved` from the mid-run patch. That's the scenario that motivated my comment on #481, and it works.
- **Tests:** 88/92 here; the 4 failures (`status`/`result`/`resolveStateDir`) reproduce identically on unmodified `main`, so delta vs main is zero. Typecheck clean. No garbage log line from the empty-message overlay event (`appendLogLine` trims and skips).

Two minor observations, neither blocking:

1. **Sandbox echo shape drifts across CLI versions.** On 0.144.4 the echo is `{ "type": "readOnly", "networkAccess": false }` — no `access` field, unlike the fixture and the PR-body example (`access: { type: "fullAccess" }`, the 0.144.1 shape). Since the plugin stores it verbatim as opaque data this is harmless, but consumers of `resolved.sandbox` shouldn't assume `access` exists, and the fixture now mirrors an older shape than current CLIs return.
2. **`thread/resume` of a turn-less thread fails on 0.144.4** ("no rollout found") — the rollout is created lazily at the first turn. Irrelevant to this PR (real resumes always follow turns), just a protocol nuance worth knowing while touching this surface.

LGTM — this closes the observability half of the #476/#481 invisibility pair, and the mid-run patch means even wedged jobs finally tell you what they ran with.

PR #511, opened 2026-07-16
fix(status): avoid HTML breaks in action cells

### LINKED ISSUE #495: `/codex:status` renders literal `<br>` in Actions column instead of a line break
### Summary
`/codex:status` renders a literal `<br>` in the Actions column of the job table instead of a line break, because Claude Code's markdown table renderer does not interpret raw HTML.

### Steps to reproduce
1. Start a background review, e.g. `/codex:review --background`
2. Run `/codex:status`
3. Look at the Actions column of the job table

### Expected
The suggested actions render on separate lines:
```
/codex:status review-xxxx
/codex:cancel review-xxxx
```

### Actual
They appear on one line joined by a literal `<br>`:
```
`/codex:status review-xxxx`<br>`/codex:cancel review-xxxx`
```

### Root cause
In `scripts/lib/render.mjs` (v1.0.4), the Actions cell joins the action code spans with an HTML `<br>` tag:

```js
${actions.map((action) => `\`${action}\``).join("<br>")}
```

Claude Code's markdown table renderer prints `<br>` as plain text rather than interpreting it as a line break.

### Suggested fix
Replace the `<br>` separator with a renderer-supported way to show the actions on separate lines (e.g. join with a space, or emit the actions as separate lines / a nested list outside the cell).

### Environment
- codex-plugin-cc version: 1.0.4
- Claude Code version: 2.1.209
- OS: macOS 26.5.1

tianrking:
## Summary
- replace raw HTML `<br>` separators in `/codex:status` action cells with a Markdown-table-safe semicolon delimiter
- update the runtime regression test to verify both actions remain visible without HTML output

Fixes #495

## Root cause
Claude Code renders raw `<br>` text inside table cells rather than interpreting it as a line break, so the two action commands were displayed with the HTML tag visible.

## Validation
- `node --test --test-name-pattern='renders active jobs with status and cancel actions' tests/runtime.test.mjs`
- `git diff --check`

`npm test` was also started but exceeded the local two-minute command limit without producing a failure; the directly affected runtime test passed.

### DIFF
--- a/plugins/codex/scripts/lib/render.mjs
+++ b/plugins/codex/scripts/lib/render.mjs
@@ -116,7 +116,7 @@ function appendActiveJobsTable(lines, jobs) {
       actions.push(`/codex:cancel ${job.id}`);
     }
     lines.push(
-      `| ${escapeMarkdownCell(job.id)} | ${escapeMarkdownCell(job.kindLabel)} | ${escapeMarkdownCell(job.status)} | ${escapeMarkdownCell(job.phase ?? "")} | ${escapeMarkdownCell(job.elapsed ?? "")} | ${escapeMarkdownCell(job.threadId ?? "")} | ${escapeMarkdownCell(job.summary ?? "")} | ${actions.map((action) => `\`${action}\``).join("<br>")} |`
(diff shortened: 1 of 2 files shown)


dkundel-openai:
Duplicate of #599.

PR #523, opened 2026-07-18
fix(review): accept positional focus text for parity with adversarial-review

### LINKED ISSUE #522: /codex:review rejects focus text — breaks interface parity with /codex:adversarial-review and blocks non-English model/effort entry
## Describe the bug

`/codex:review` throws on any positional argument because `validateNativeReviewRequest` rejects non-empty focus text unconditionally (`codex-companion.mjs:271-274`): *"does not support custom focus text. Retry with `/codex:adversarial-review ...`"*.

**1. It breaks interface parity.** `/codex:adversarial-review` accepts focus text after the flags; `/codex:review` rejects it wholesale. Two sibling review commands that should behave as mirrors differ in a way the user only discovers by hitting the error. The suggested escape hatch (`adversarial-review`) is not a substitute — it changes the review *semantics* (adversarial framing), so the user is forced to pick between "type a sentence" and "get the review I actually want."

**2. It makes non-English model/effort entry unusable.** The short aliases (`spark`, `sol`, `terra`, `luna`) and the latin-only `--flag value` grammar assume the user types in English/Latin script. A user who writes the model/effort **value** as a natural phrase in their own language has no accepted input path. They must know the exact flag syntax *and* the exact latin model id.

## To reproduce

A Japanese-speaking user requesting a high-effort review on the `sol` model, typed the natural way (model → モデル, sol → ソル, effort → エフォート, xhigh → エックスハイ — Japanese phonetically transcribes everything into kana):

```
$ /codex:review モデル ソル エフォート エックスハイ
Error: `/codex:review` now maps directly to the built-in reviewer and does not support custom focus text.
         Retry with `/codex:adversarial-review ...`
```

The intent ("use model sol, effort xhigh") is unambiguous to a human, but the parser treats the non-Latin prose as forbidden focus text — and even if it didn't, there is no `ソル` alias to map back to `sol`.

## Expected behavior

Not mutually exclusive:

1. **Lenient positional parsing**: if a positional argument can be interpreted as model/effort intent, accept it as such instead of rejecting the whole invocation.
2. **Recognize model/effort intent regardless of position/script**: scan raw arguments for model/effort intent in any language before the strict rejection, so `/codex:review モデル ソル` resolves to `--model sol`.
3. **Parity**: accept the same argument surface in both review commands, so the interface is uniform.

## Environment

Plugin v1.0.6, codex-cli 0.144.4, macOS.

## Related

- #333, #512 — focus-text re-tokenization (inverse problem: flags leaking *into* focus text; this issue is benign focus text rejected wholesale).

axisrow:
## What

Removes the unconditional focus-text rejection in `/codex:review` so it no longer aborts where `/codex:adversarial-review` succeeds.

Closes #522.

## Why

`/codex:review` rejected any positional argument as "custom focus text" (`validateNativeReviewRequest`), while `/codex:adversarial-review` accepts it. Two sibling review commands behaved differently in a way the user only discovered by hitting the error, and the suggested escape hatch (`adversarial-review`) changes the review *semantics* — not a substitute for "let me name the model in my own words."

This also blocks non-English model/effort entry: even when a host (e.g. Claude Code reading the command's `.md`) rewrites a natural-language phrase into `--model`/`--effort`, any residual positional word still aborted the native review (see the Japanese example in #522).

## Change

Remove the focus-text branch from `validateNativeReviewRequest`. The native reviewer (`review/start`) does not consume focus text, so leftover positional words are now silently ignored rather than rejecting the invocation. Review semantics are unchanged (still native structured-review via `review/start`); only the spurious abort is gone.

This PR does **not** add language/alias recognition in the companion — that stays the host's job. The fix is purely about interface parity: review no longer fails where adversarial-review succeeds.

## Scope

This is independent of #471 — the two touch different concerns (#471 is model/effort negotiation + broker refresh; this is a review-command parity fix). Based on `main` so it can land on its own.

## Testing

- New runtime test `review accepts (ignores) positional focus text for parity with adversarial-review` (red before, green after).
- Updated the stale `review rejects focus text` test and the `commands.test.mjs` assertion that pinned the old wording.
- 0 regressions vs `main` (the pre-existing suite failures are unchanged).

### DIFF
--- a/plugins/codex/commands/review.md
+++ b/plugins/codex/commands/review.md
@@ -36,7 +36,7 @@ Argument handling:
 - Do not strip `--wait` or `--background` yourself.
 - Do not add extra review instructions or rewrite the user's intent.
 - The companion script parses `--wait` and `--background`, but Claude Code's `Bash(..., run_in_background: true)` is what actually detaches the run.
-- `/codex:review` is native-review only. It does not support staged-only review, unstaged-only review, or extra focus text.
+- `/codex:review` is native-review only. It does not support staged-only review or unstaged-only review. Positional focus text is ignored by the native reviewer (for interface parity with `/codex:adversarial-review`); it does not abort the review.
 - If the user needs custom review instructions or more adversarial framing, they should use `/codex:adversarial-review`.
 
 Foreground flow:
--- a/plugins/codex/scripts/codex-companion.mjs
+++ b/plugins/codex/scripts/codex-companion.mjs
@@ -268,13 +268,13 @@ function buildNativeReviewTarget(target) {
   return null;
 }
 
-function validateNativeReviewRequest(target, focusText) {
-  if (focusText.trim()) {
-    throw new Error(
-      `\`/codex:review\` now maps directly to the built-in reviewer and does not support custom focus text. Retry with \`/codex:adversarial-review ${focusText.trim()}\` for focused review instructions.`
-    );
-  }
-
+function validateNativeReviewRequest(target, _focusText) {
+  // Parity with /codex:adversarial-review: positional focus text no longer
+  // aborts the native review. The native reviewer (review/start) does not
+  // consume focus text, so leftover positional words are silently ignored
+  // rather than rejecting the invocation. This keeps `/codex:review --model sol`
+  // usable when a host forwards residual positional text alongside flags,
+  // and removes an interface-parity gap with /codex:adversarial-review (#522).
   const nativeTarget = buildNativeReviewTarget(target);
   if (!nativeTarget) {
     throw new Error("This `/codex:review` target is not supported by the built-in reviewer. Retry with `/codex:adversarial-review` for custom targeting.");
--- a/plugins/codex/scripts/codex-companion.mjs
+++ b/plugins/codex/scripts/codex-companion.mjs
@@ -726,7 +726,7 @@ async function handleReviewCommand(argv, config) {
     scope: options.scope
   });
 
-  config.validateRequest?.(target, focusText);
+  config.validateRequest?.(target);
   const metadata = buildReviewJobMetadata(config.reviewName, target);
   const job = createCompanionJob({
     prefix: "review",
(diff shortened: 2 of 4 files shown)


axisrow:
## 🔍 Local review (cycle 1)
Reviewed locally (Claude subagent + Codex companion), no bots pinged. Critical-only bar.

| Verdict | Reviewer | Finding | Location |
|---|---|---|---|
| SKIP | claude | Dead `focusText` local + vestigial `_focusText` param in the native-review branch (computed L366, passed to validator L369, used nowhere now) | codex-companion.mjs:366,369 |

**Codex companion**: `approve`, 0 findings — *"no critical security, data-loss, correctness, or performance defect is supported by the PR diff."*

**Claude subagent**: 0 critical findings after verifying 6 claims against the code (focusText cannot leak into `review/start` RPC — payload is exactly `{threadId, delivery, target}`; cannot reach job metadata or rendered output; staged-only rejection still enforced; test is a genuine inversion, not a weakened assertion; parity-only scoping is sound).

No `FIX` verdicts → final cycle. The one SKIP (dead param) will be addressed in the cleanup pass or left as-is since it's harmless.

axisrow:
## 🔍 Local review (cycle 1, round 2 after cleanup commit)
Reviewed locally (Claude subagent + Codex companion), no bots pinged. Critical-only bar.

| Verdict | Reviewer | Finding | Location |
|---|---|---|---|
| SKIP | claude | `validateRequest?.(target, focusText)` still passes a now-dead 2nd arg (validator takes 1). Harmless — JS drops it. One-line cleanup. | codex-companion.mjs:729 |

**Codex companion**: `approve`, 0 findings — *"no defensible critical security, data-loss, correctness, or performance risk."*

**Claude subagent**: 0 critical after verifying 6 claims against the code (native path never reads focusText; cleanup commit did NOT break the adversarial path — focusText still reaches `buildAdversarialReviewPrompt`; no leak into prompt/job metadata/logs; `review --scope working-tree <leftover>` runs native exit 0 without passing leftovers as instructions; tests are genuine inversions not weakened; pre-existing failures confirmed identical on main).

No `FIX` verdicts → final cycle. The one SKIP (dead 2nd arg) applied in the cleanup pass.

PR #536, opened 2026-07-19
Fix result lookup for active jobs

### LINKED ISSUE #498: `result <id>` on a queued/running job reports "No job found" — the still-running message is unreachable when a reference is given
On 1.0.6 (macOS), asking for the result of a job that is still running reports the job as nonexistent:

```
$ node codex-companion.mjs result task-mrkxdpsi-mij20h --cwd /Users/me/repo
No job found for "task-mrkxdpsi-mij20h". Run /codex:status to list known jobs.
$ echo $?
1
$ node codex-companion.mjs status task-mrkxdpsi-mij20h --cwd /Users/me/repo
- task-mrkxdpsi-mij20h | running | rescue | Codex Task
```

There is code in `resolveResultJob` that intends to say `Job X is still running. Check /codex:status and try again once it finishes.` — but it is unreachable when a job id is passed.

`scripts/lib/job-control.mjs` (line numbers at current main):

- `resolveResultJob` (line 256) first calls `matchJobReference(jobs, reference, terminalPredicate)` with the predicate filtering to `completed`/`failed`/`cancelled`.
- `matchJobReference` (line 191) applies the predicate, then matches the reference against the filtered list, and **throws** `No job found for "<ref>"` (line 210) when nothing matches.
- So for a queued/running job looked up by id, the throw fires inside the first call and the `active` check at line 269 never runs. The "still running" message is reachable only in the no-reference path, where `matchJobReference` returns `filtered[0] ?? null` instead of throwing.

This is worse than a cosmetic wrong message because `No job found` is the same string the cwd-mismatch case produces. A caller polling `result <id>` cannot distinguish "wrong workspace" from "not finished yet" from "no such job", and the natural reading — the job does not exist — is the one that is false. We traced a real incident to an orchestrator concluding from this exact string that a dispatched job had never run, and redoing the work while the job was still going.

Possible fixes, either works:

- Make `matchJobReference` return `null` when a reference matches nothing, and let each caller produce its own error message (`resolveResultJob` already has the branching to do this correctly).
- Or in `resolveResultJob`, match the reference against the unfiltered job list first, then check the status of the match — producing "still running" for active jobs, "No finished job" only when the id genuinely is not in the workspace.

https://claude.ai/code/session_01B5Xst2TXwvPQ5ecfMckbBR

Kevinjohn:
Users requesting the result of an active job receive a false “No job found” error.

Fixes #498.

## Observable sequence

- Run `result <job-id>` for a queued or running job.
- Expected: report the job's current state and tell the user to try again after it finishes.
- Actual: report that no matching job exists.

## Root cause and fix

Explicit references were matched only after filtering the job list to terminal states, so active jobs could never reach the existing active-state error. Match explicit references against all jobs first, then branch on the selected job's status.

The CLI regression test covers both queued and running jobs and verifies that neither is reported as missing.

## Validation

- Focused regression: queued and running cases passed
- Full test suite: passed
- `pnpm run build`: passed
- `git diff --check`: clean

## Actual screenshots

Before and after screenshots use the same stored running job and command.

**Before — active job incorrectly reported missing**

The result command on main reports that the stored running job cannot be found

**After — active job reports its actual state and next action**

The same result command on the fix branch reports that the job is still running

### DIFF
--- a/plugins/codex/scripts/lib/job-control.mjs
+++ b/plugins/codex/scripts/lib/job-control.mjs
@@ -256,9 +256,18 @@ export function buildSingleJobSnapshot(cwd, reference, options = {}) {
 export function resolveResultJob(cwd, reference) {
   const workspaceRoot = resolveWorkspaceRoot(cwd);
   const jobs = sortJobsNewestFirst(reference ? listJobs(workspaceRoot) : filterJobsForCurrentSession(listJobs(workspaceRoot)));
+
+  if (reference) {
+    const selected = matchJobReference(jobs, reference);
+    if (selected.status === "queued" || selected.status === "running") {
+      throw new Error(`Job ${selected.id} is still ${selected.status}. Check /codex:status and try again once it finishes.`);
+    }
+    return { workspaceRoot, job: selected };
+  }
+
   const selected = matchJobReference(
     jobs,
-    reference,
+    "",
     (job) => job.status === "completed" || job.status === "failed" || job.status === "cancelled"
   );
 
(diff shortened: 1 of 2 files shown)


dkundel-openai:
Duplicate of #504.

Issue #540, opened 2026-07-22
Ending any Claude session kills the shared broker mid-turn: concurrent sessions' tasks die silently (exit 0) and stay "running" forever


leitxio:
## Environment

- Plugin: `codex@openai-codex` v1.0.6 (repo `openai/codex-plugin-cc`)
- Claude Code on Linux (Fedora, Node 24)
- **Multiple concurrent Claude Code sessions in the same workspace** (interactive sessions plus occasional short-lived/headless ones)

## Summary

When any Claude session in a workspace ends, the `SessionEnd` hook unconditionally shuts down the workspace's **shared** broker. If another session's rescue task is streaming a turn through that broker at that moment, its companion process **exits silently with code 0 mid-turn**: no error, no `failed` state — the job log just stops after an ordinary event line. The job file stays `status: "running"` forever (nothing ever pid-checks it), and the stale record then blocks all further tasks on that session ("Task … is still running") until a manual `/codex:cancel`.

Three defects compose into this:

1. **`SessionEnd` assumes one session per workspace.** `scripts/session-lifecycle-hook.mjs:101` (`handleSessionEnd`) calls `sendBrokerShutdown()` on the broker recorded in the shared per-workspace `broker.json`, with no check whether other live sessions (or in-flight jobs) are using it. The broker's `shutdown()` then calls `socket.end()` on **every connected client** (`scripts/app-server-broker.mjs:104`), including another session's mid-stream turn.

2. **A transport drop mid-turn is a silent exit, not an error.** `captureTurn` awaits `state.completion` (`scripts/lib/codex.mjs:606`), a bare promise resolved only by a `turn/completed` notification (`resolveCompletion`, `lib/codex.mjs:304-310`). It is never raced against the client's exit. On socket close, `handleExit` (`scripts/lib/app-server.mjs:163`) rejects only *pending request* promises — during streaming there are none (`turn/start` already resolved) — so nothing rejects, no catch runs, `runTrackedJob` never records a failure. The closed socket was the last ref'd handle (the completion timer is `unref()`d, `lib/codex.mjs:393`), so the event loop drains and Node exits 0 mid-`await`. The same silent death occurs for any transport loss: direct-mode app-server child crash, broker crash, or the client being SIGTERMed (no signal handler updates job state either).

[...]

PR #541, opened 2026-07-22
Fix test broker leaks, state races, and signal-masked command failures

apedroheringer:
> **Description updated 2026-08-19 to match the current head (`f3204e8`).** The original text
> described an earlier revision: it claimed signal statuses were *normalized*, when review
> feedback led to the opposite (and better) approach, and it did not mention the ownership
> protocol added while addressing that feedback. The code did not change with this edit — only
> this description. Diff and review threads are unchanged.

## Problem

Running `npm test` leaves dozens of orphaned processes behind (broker `app-server-broker.mjs serve` instances plus fake `codex app-server` children spawned from `/tmp/codex-plugin-test-*` workspaces). On one machine a single suite run left ~30 brokers + ~29 fake app-servers running after the temp dirs had already been cleaned up. Root causes, verified against `main` (db52e28):

1. **Test teardown is per-test and easy to miss.** Most tests that lazily start a broker never run the `SessionEnd` hook (e.g. the lazy-broker `status` test at `tests/runtime.test.mjs:2209`, while its neighbours at `:2153`/`:2198` do clean up). Any test failure or early `return` also skips cleanup.
2. **`state.json` writes are a read-modify-write with no lock and no atomic rename** (`lib/state.mjs` — `saveState`/`updateState` use a bare `fs.writeFileSync`). Two concurrent commands can lose updates or tear the file.
3. **`ensureBrokerSession` is check-then-create with no serialization** (`lib/broker-lifecycle.mjs`). Two concurrent callers can both miss the existing broker and spawn duplicates, orphaning one.
4. **Signal-terminated commands are reported as success**: `lib/process.mjs` coalesces `spawnSync`'s `status: null` to `0`, so `runCommandChecked` treats a SIGTERM/SIGKILL death as exit 0.

## Changes

**Test-suite leak**

- **`tests/helpers.mjs`**: `makeTempDir` tracks created workspaces, and a global `after()` hook shuts down any broker a test left running (graceful `broker/shutdown`, then `terminateProcessTree` escalation) and removes its session dir. It only touches sessions carrying the per-process `instanceToken`, so a fixture that deliberately persists an arbitrary endpoint is left alone.
- **`tests/runtime.test.mjs`**: the lazy-broker `status` test now runs `SessionEnd` like its neighbours.

**State races**

- **`lib/locking.mjs` (new)**: advisory lock with no new dependencies. A candidate directory is populated and then `rename`d onto the lock path, so publication is atomic; the owner token is part of the owner filename, so a previous generation cannot unlink a successor's record. Reclaim of an abandoned lock is decided by *process identity* (boot id + start time), not by age alone.
- **`lib/fs.mjs`**: `writeJsonFileAtomic` — temp file, `fsync`, atomic `rename`; private-mode helpers.
- **`lib/state.mjs`**: `saveState`/`updateState` are serialized per workspace and write through `writeJsonFileAtomic`.
- **`lib/broker-lifecycle.mjs`**: the check-then-create sequence in `ensureBrokerSession` is serialized per workspace.

**Safe teardown under PID reuse** — not in the original description; added while addressing review feedback, because killing a broker by persisted PID alone can kill an unrelated process that inherited that PID:

- Broker shutdown is authenticated with a per-instance token carried on the wire (`app-server-broker.mjs`), so a shutdown request cannot be aimed at a listener that is not ours.
- A process-identity subsystem (`lib/process.mjs`) distinguishes "the process we started" from "some process with that PID", with a `/proc` fast path and `ps`/PowerShell fallbacks.
- Sessions persisted by older versions (no token, no identity) fall back to an argv heuristic, and an orphaned process tree is reclaimed without killing what cannot be verified.

**Signal-terminated commands**

- **`lib/process.mjs`**: `runCommand` **preserves Node's `spawnSync` contract** — a signal death keeps `status: null` and reports a non-null `signal`. Consumers (`runCommandChecked`, `binaryAvailable`) now test `signal != null || status !== 0`, and failures render as `signal=SIGTERM`. Normalizing the status to a non-zero number, as the original description proposed, would have misreported what `spawnSync` actually returned; checking the signal explicitly does not. The only other caller of `runCommand` is `lib/git.mjs`, which does not use timeouts.

**Hook timeout**

- **`plugins/codex/hooks/hooks.json`**: the `SessionEnd` timeout goes from 5 s to 30 s. Root cause 4 above is partly a race created by that limit: graceful `broker/shutdown` followed by `terminateProcessTree` escalation does not reliably fit in 5 s, and a hook cut off mid-teardown is what left brokers behind while reporting success. The worst case is a session exit that waits up to 30 s; it only approaches that when a broker ignores graceful shutdown.

## Testing

- `tests/` declares **134 test cases across 10 files** on this branch (`grep -cE '^\s*(test|it)\('`). The "95/95 (91 + 4 new)" figure in the original description predates the ownership-protocol work and no longer matches; the new tests cover the lock (published/reclaimed/contended), process identity, PID reuse, orphaned trees, and cross-process concurrent `saveState`.
- Last recorded full run: **2026-07-25**, against this head, all green, with the post-suite process table verified twice — **zero** leftover brokers, fake app-servers, or temp session dirs. I have not re-run it since; if you want a fresh number before reviewing, say so and I will post one.
- `npm run check-version` passes.
- ⚠️ For anyone reproducing locally: the suite must run with `CLAUDE_*`/`CODEX_*` variables cleared, or they leak into the fixtures' environment and fail tests for reasons unrelated to the change.

## Related work

- #490 attacks orphaned brokers from a complementary angle (idle timeout in the broker runtime); this PR makes the *test suite* clean up after itself and prevents duplicate spawns at the source.
- #518 overlaps on serializing state updates; this PR is a smaller, dependency-free take (directory lock + atomic rename) that also covers the test-suite leak and the signal-status bug.
- #540 reports tasks dying "silently (exit 0)" — the `lib/process.mjs` fix here addresses one path where a signal death is reported as success.

## Known, and deliberately not fixed here

`resolveStateDir` (`lib/state.mjs`) hashes the **canonical** workspace path but takes the directory-name slug from the **non-canonical** one, so two aliases of the same non-Git workspace resolve to different state directories — and therefore different locks:

```
alias-a → alias-a-c3dd6040e0182f82
alias-b → alias-b-c3dd6040e0182f82   # same hash, different directory
```

This is pre-existing on `main` — this PR does not touch those lines — and Git workspaces are unaffected because the toplevel path is already canonical. Fixing it means deriving the slug from the canonical path too. Left out to keep this PR to its stated scope; happy to send it separately.

### DIFF
--- a/plugins/codex/scripts/lib/broker-lifecycle.mjs
+++ b/plugins/codex/scripts/lib/broker-lifecycle.mjs
@@ -1,3 +1,4 @@
+import { randomUUID } from "node:crypto";
 import fs from "node:fs";
 import net from "node:net";
 import os from "node:os";
@@ -5,16 +6,40 @@ import path from "node:path";
 import process from "node:process";
 import { spawn } from "node:child_process";
 import { fileURLToPath } from "node:url";
+
 import { createBrokerEndpoint, parseBrokerEndpoint } from "./broker-endpoint.mjs";
 import { withLock } from "./locking.mjs";
+import {
+  isProcessTreeRunning,
+  processHasLaunchToken,
+  terminateProcessTree,
+  waitForProcessExit
+} from "./process.mjs";
 import { resolveStateDir } from "./state.mjs";
 
 export const PID_FILE_ENV = "CODEX_COMPANION_APP_SERVER_PID_FILE";
 export const LOG_FILE_ENV = "CODEX_COMPANION_APP_SERVER_LOG_FILE";
 const BROKER_STATE_FILE = "broker.json";
+const PRIVATE_DIR_MODE = 0o700;
+const PRIVATE_FILE_MODE = 0o600;
+
+function setMode(filePath, mode) {
+  try {
+    fs.chmodSync(filePath, mode);
+  } catch {
+    // Windows and restrictive filesystems may not implement POSIX modes.
+  }
+}
+
+function ensurePrivateDir(dir) {
+  fs.mkdirSync(dir, { recursive: true, mode: PRIVATE_DIR_MODE });
+  setMode(dir, PRIVATE_DIR_MODE);
+}
 
 export function createBrokerSessionDir(prefix = "cxc-") {
-  return fs.mkdtempSync(path.join(os.tmpdir(), prefix));
+  const sessionDir = fs.mkdtempSync(path.join(os.tmpdir(), prefix));
+  setMode(sessionDir, PRIVATE_DIR_MODE);
+  return sessionDir;
 }
 
 function connectToEndpoint(endpoint) {
@@ -23,15 +48,22 @@ function connectToEndpoint(endpoint) {
 }
 
 export async function waitForBrokerEndpoint(endpoint, timeoutMs = 2000) {
-  const start = Date.now();
-  while (Date.now() - start < timeoutMs) {
+  const deadline = Date.now() + timeoutMs;
+  while (Date.now() < deadline) {
     const ready = await new Promise((resolve) => {
       const socket = connectToEndpoint(endpoint);
-      socket.on("connect", () => {
-        socket.end();
-        resolve(true);
-      });
-      socket.on("error", () => resolve(false));
+      let settled = false;
+      const finish = (value) => {
+        if (settled) {
+          return;
+        }
+        settled = true;
+        socket.destroy();
+        resolve(value);
+      };
+      socket.setTimeout(Math.max(1, Math.min(150, deadline - Date.now())), () => finish(false));
+      socket.on("connect", () => finish(true));
+      socket.on("error", () => finish(false));
     });
     if (ready) {
       return true;
@@ -41,33 +73,90 @@ export async function waitForBrokerEndpoint(endpoint, timeoutMs = 2000) {
   return false;
 }
 
-export async function sendBrokerShutdown(endpoint) {
-  await new Promise((resolve) => {
+export async function sendBrokerShutdown(endpoint, options = {}) {
+  return await new Promise((resolve) => {
     const socket = connectToEndpoint(endpoint);
+    let settled = false;
+    let buffer = "";
+    const finish = (result = null) => {
+      if (settled) {
+        return;
+      }
+      settled = true;
+      socket.destroy();
+      resolve(result);
+    };
     socket.setEncoding("utf8");
+    socket.setTimeout(options.timeoutMs ?? 2000, () => finish(null));
     socket.on("connect", () => {
-      socket.write(`${JSON.stringify({ id: 1, method: "broker/shutdown", params: {} })}\n`);
+      socket.write(
+        `${JSON.stringify({
+          id: 1,
+          method: "broker/shutdown",
+          params: { instanceToken: options.instanceToken ?? null }
+        })}\n`
+      );
     });
-    socket.on("data", () => {
-      socket.end();
-      resolve();
+    socket.on("data", (chunk) => {
+      buffer += chunk;
+      const newlineIndex = buffer.indexOf("\n");
+      if (newlineIndex === -1) {
+        return;
+      }
+      try {
+        const response = JSON.parse(buffer.slice(0, newlineIndex));
+        finish({ result: response?.result ?? null, error: response?.error ?? null });
+      } catch {
+        finish(null);
+      }
     });
-    socket.on("error", resolve);
-    socket.on("close", resolve);
+    socket.on("error", () => finish(null));
+    socket.on("close", () => finish(null));
   });
 }
 
-export function spawnBrokerProcess({ scriptPath, cwd, endpoint, pidFile, logFile, env = process.env }) {
-  const logFd = fs.openSync(logFile, "a");
-  const child = spawn(process.execPath, [scriptPath, "serve", "--endpoint", endpoint, "--cwd", cwd, "--pid-file", pidFile], {
-    cwd,
-    env,
-    detached: true,
-    stdio: ["ignore", logFd, logFd]
-  });
-  child.unref();
-  fs.closeSync(logFd);
-  return child;
+function isValidPid(pid) {
+  return Number.isSafeInteger(pid) && pid > 0;
+}
+
+export function spawnBrokerProcess({
+  scriptPath,
+  cwd,
+  endpoint,
+  pidFile,
+  logFile,
+  instanceToken,
+  env = process.env
+}) {
+  const logFd = fs.openSync(logFile, "a", PRIVATE_FILE_MODE);
+  try {
+    setMode(logFile, PRIVATE_FILE_MODE);
+    const child = spawn(
+      process.execPath,
+      [
+        scriptPath,
+        "serve",
+        "--endpoint",
+        endpoint,
+        "--cwd",
+        cwd,
+        "--pid-file",
+        pidFile,
+        "--instance-token",
+        instanceToken
+      ],
+      {
+        cwd,
+        env,
+        detached: true,
+        stdio: ["ignore", logFd, logFd]
+      }
+    );
+    child.unref();
+    return child;
+  } finally {
+    fs.closeSync(logFd);
+  }
 }
 
 function resolveBrokerStateFile(cwd) {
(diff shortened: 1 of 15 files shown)


plugins/codex/scripts/lib/locking.mjs
```diff
@@ -0,0 +1,88 @@
+import fs from "node:fs";
+
+const DEFAULT_TIMEOUT_MS = 5000;
+const DEFAULT_STALE_MS = 30000;
```
provencher-oai, line 9: Can we record the lock owner's PID and reclaim a dead owner's lock immediately? `staleMs` is 30 seconds, but state acquisition times out after five seconds and broker startup after ten. Killing a worker while it owns the lock can therefore break cancellation and exhaust the SessionEnd cleanup deadline.

provencher-oai (commented):
Found four additional correctness and safety issues beyond the existing automated review comment.

plugins/codex/scripts/lib/locking.mjs
```diff
@@ -0,0 +1,88 @@
...
+      return false;
+    }
+    throw error;
+  }
+}
+
+function reclaimIfStale(lockDir, staleMs) {
+  try {
+    const age = Date.now() - fs.statSync(lockDir).mtimeMs;
+    if (age > staleMs) {
```
provencher-oai: Can we make stale-lock reclamation conditional on the exact lock instance observed here? Another process can replace the directory between `statSync()` and `rmdirSync()`, so removing the replacement lets multiple state writers or broker creators enter the critical section. The unconditional unlock has the same ownership problem.

plugins/codex/scripts/lib/state.mjs
```diff
@@ -89,9 +90,29 @@ function removeFileIfExists(filePath) {
...
 
-export function saveState(cwd, state) {
+function resolveStateLockDir(cwd) {
+  return path.join(resolveStateDir(cwd), ".state.lock");
+}
+
+function writeStateFileAtomic(stateFile, nextState) {
+  const tmpFile = `${stateFile}.${process.pid}.${Math.random().toString(36).slice(2)}.tmp`;
+  try {
+    const fd = fs.openSync(tmpFile, "w");
```
provencher-oai: Can we create the replacement with the existing file's permissions, or explicitly use `0o600`? This temporary file uses the process umask, so replacing a `0600` state file typically changes it to `0644` even though it contains complete background-task prompts.

tests/helpers.mjs
```diff
@@ -3,11 +3,89 @@ import os from "node:os";
...
+  }
+}
+
+// Global teardown: any broker a test started (directly or lazily) and did not
+// tear down is shut down here, so the suite never leaves broker/app-server
+// processes behind — even when a test fails or returns early.
+after(async () => {
+  const leakedPids = [];
+  const leakedSessions = [];
+  for (const dir of trackedTempDirs) {
```
provencher-oai, line 27: Can teardown verify that this endpoint belongs to a broker started by the fixture and put a deadline on shutdown? An existing runtime fixture stores `unix:/tmp/fake-broker.sock` without ownership metadata, so this hook can send `broker/shutdown` to an unrelated process, unlink its socket, or hang indefinitely waiting for a response.

provencher-oai (changes requested):
Some important changes to fix

apedroheringer:
Confirmed — this is a real bug, and it's now fixed in 2f6cc6a + f51a8a4. Thanks for the catch.

**Tracing it.** For an owned session, `sendBrokerShutdown()` resolves `null` on `ECONNREFUSED` (`broker-lifecycle.mjs:113`), so `shutdownAck` is null and `ownershipVerified` stays false. `exited` correctly becomes true for the dead PID, but the endpoint check at `broker-lifecycle.mjs:365` then throws because the stale socket file is still on disk. `ensureBrokerSessionLocked()` awaits that call at `:408` with no try/catch before spawning the replacement, so the workspace stays wedged until the state and socket files are deleted by hand. `canDiscardUnownedSession()` doesn't cover it either: it only applies to sessions without an `instanceToken`, and additionally requires the endpoint artifact to be absent.

**The fix.** That path is now reclaimable instead of fatal, but deliberately not on the absence of an ack alone — `sendBrokerShutdown()` also returns `null` on timeout, which is what a live-but-hung broker produces. Reclaiming requires all of:

1. the endpoint sits inside the `mkdtemp` 0700 session directory the plugin created;
2. neither the recorded PID nor its process group is running;
3. connecting to the endpoint is refused (`ECONNREFUSED`/`ENOENT`).

Anything else — timeout, `EACCES`/`EPERM`, an unexpected error, a malformed endpoint — keeps the current failure, so the bias stays on the safe side: a false "alive" preserves the wedge, but no socket owned by another process is ever unlinked.

The session-directory containment in (1) is what closes the gap a plain refused-connect leaves open: a process that has `bind()`-ed but not yet `listen()`-ed also refuses connections, so "refused" alone does not establish that the socket is ours. Requirement (2) uses both `isProcessTreeRunning()` and `isProcessRunning()`, because the former inspects the process *group* on Linux and a recycled PID in a different group would otherwise read as dead.

**Tests.** Five regression tests in `tests/broker-lifecycle.test.mjs`: a stale socket is reclaimed; `ensureBrokerSession()` recovers end-to-end instead of throwing; and three negative cases that must never reclaim — a live listener on the endpoint, a live owner with a refused endpoint, and an otherwise stale-looking endpoint outside the session directory. Suite is at 105 passing.

---

Two things worth flagging separately, both **pre-existing on `main`** rather than introduced here — happy to open issues or fold either into this PR if you'd prefer:

1. `handleSessionEnd()` calls `shutdownBrokerSession()` directly (`session-lifecycle-hook.mjs:89`) without taking the per-workspace `.broker.lock` that `ensureBrokerSession()` holds. A `SessionEnd` running against a stale snapshot can therefore reach `clearBrokerSession()` after a concurrent `ensure` has already persisted a replacement, orphaning the new broker. The reclaim path makes that window mutate the filesystem, so it seemed worth naming.

2. Four tests fail on `main` today, independent of this branch (verified by running them at `db52e28`): `resolveStateDir uses a temp-backed per-workspace directory` fails whenever `CLAUDE_PLUGIN_DATA` points outside `os.tmpdir()` — which is the case inside Claude Code itself, the plugin's own environment — and three `runtime.test.mjs` cases (`status shows phases…`, `status preserves adversarial review kind labels`, `result returns the stored output…`) fail with "No finished Codex jobs found for this repository yet."

plugins/codex/scripts/lib/broker-lifecycle.mjs
```diff
@@ -88,17 +179,305 @@ export function loadBrokerSession(cwd) {
...
+}
+
+export async function shutdownBrokerSession(cwd, options = {}) {
+  const session = options.session ?? loadBrokerSession(cwd);
+  if (!session) {
+    return { found: false, exited: true, forced: false, reclaimedStaleEndpoint: false };
+  }
+
+  const pid = resolveBrokerPid(session);
+  if (session.endpoint && !session.instanceToken) {
```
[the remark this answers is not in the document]
apedroheringer: Addressed in `c0b5eb2`. A tokenless session is retired only when its endpoint exactly matches the endpoint derived from `sessionDir`, its PID file is canonical, and the live PID command line matches the complete legacy broker launch sequence. The old unauthenticated shutdown is then allowed and awaited. A positive upgrade regression and a negative test for a new authenticated broker whose persisted token was lost are both included. `npm test` passes 112/112 and `npm run check-version` passes.

tests/helpers.mjs
```diff
@@ -3,11 +3,89 @@ import os from "node:os";
...
+  for (const pid of leakedPids) {
+    while (pidAlive(pid) && Date.now() < deadline) {
+      await sleep(50);
+    }
+    if (pidAlive(pid)) {
+      terminateProcessTree(pid);
+      await sleep(200);
+    }
+    if (pidAlive(pid)) {
+      throw new Error(`Leaked broker process ${pid} survived suite teardown.`);
```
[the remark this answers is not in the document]
apedroheringer: Addressed in `96875ff`. Teardown no longer relies on a raw `kill(pid, 0)` liveness check. `isProcessRunning()` now reads Linux process state and treats `Z`/`X` as exited, with regression coverage in `tests/process.test.mjs`. The full suite at current HEAD passes 112/112.

plugins/codex/scripts/lib/locking.mjs
apedroheringer: Addressed in `96875ff`. Each lock now records a private owner file with the PID, process start identity, and a UUID token. Acquisition reclaims immediately when that exact owner is no longer running, before `staleMs`; regression coverage is in `lock immediately reclaims an owner process that exited`.

plugins/codex/scripts/lib/state.mjs
apedroheringer: Addressed in `96875ff`. Atomic state replacements are created with `wx` and mode `0o600`, `fchmod`ed before writing, and `chmod`ed after rename for defense in depth. `state, job, and log artifacts are private` verifies the resulting modes.

plugins/codex/scripts/lib/locking.mjs
apedroheringer: Addressed in `96875ff`. The owner UUID is part of the ownership filename, and release/reclaim unlinks only the exact observed generation before removing an empty directory. The regression test proves that an old handle cannot release a successor lock.

plugins/codex/scripts/lib/broker-lifecycle.mjs
```diff
@@ -88,17 +178,205 @@ export function loadBrokerSession(cwd) {
...
+      killImpl: options.killImpl,
+      platform: options.platform
+    });
+  }
+
+  if (!exited) {
+    throw new Error(`Codex app-server broker ${verifiedPid ?? session.endpoint ?? "unknown"} did not exit.`);
+  }
+  if (!ownershipVerified && endpointArtifactExists(session.endpoint)) {
+    throw new Error("Codex app-server broker endpoint ownership could not be verified; persisted state was preserved.");
```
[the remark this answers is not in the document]
apedroheringer: Addressed in `2f6cc6a` and hardened in `f51a8a4`. Stale socket reclaim now requires all independent signals: the endpoint is inside the private session directory, both the recorded PID and process group are gone, and connecting is refused. Negative regressions preserve live, reused, or external endpoints.

tests/helpers.mjs
apedroheringer: Addressed in `96875ff`. Global teardown only touches sessions carrying the fixture-created `instanceToken` and routes them through `shutdownBrokerSession()`, which verifies the RPC/process identity before kill or unlink. The fixture supplies a 1000 ms timeout, failures are aggregated, and regressions cover both deadline enforcement and preservation of an unowned endpoint.

apedroheringer:
Hi @provencher-oai — I addressed all four requested changes in `96875ff` and replied to each review thread with the corresponding implementation and regression coverage. I also addressed every subsequent automated finding in `2f6cc6a`, `f51a8a4`, `c0b5eb2`, and `64fdef1`, including stale sockets, safe migration of tokenless legacy brokers, and reused-PID lock ownership on macOS and Windows. Current HEAD passes `npm test` (116/116) and `npm run check-version`. Would you mind taking another look when convenient? Thanks.

plugins/codex/scripts/lib/locking.mjs
```diff
@@ -0,0 +1,233 @@
...
+    return true;
+  }
+  if (state.kind === "owned") {
+    const processRunning = options.isProcessRunning ?? isProcessRunning;
+    if (
+      processRunning(state.owner.pid, {
+        identity: state.owner.processIdentity ?? undefined
+      })
+    ) {
+      return false;
```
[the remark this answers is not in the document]
apedroheringer: Addressed in `64fdef1`. Lock owners now persist a platform-specific process start identity: `/proc` start time on Linux, `ps lstart` on macOS/other POSIX systems, and PowerShell `StartTime` ticks on Windows. `isProcessRunning()` compares that identity before accepting a live PID as the owner, so a reused PID is reclaimed immediately. The identity lookup is cached once per owner process to avoid adding cost to every state update. Existing macOS/Windows lock files that already contain no identity are reclaimed after `staleMs`, preventing permanent wedges during upgrade. Added regressions for macOS identity, Windows identity, non-Linux PID reuse, and stale identity-less locks. `npm test` passes 116/116; `npm run check-version` passes.

apedroheringer:
@codex review

apedroheringer:
@codex review

apedroheringer:
@codex review

apedroheringer:
@codex review

apedroheringer:
@codex review

apedroheringer:
@codex review

apedroheringer:
@codex review

apedroheringer:
@codex review

apedroheringer:
@codex review

apedroheringer:
@codex review

apedroheringer:
@codex review

apedroheringer:
@codex review

apedroheringer:
@codex review

apedroheringer:
@codex review

apedroheringer:
@codex review

apedroheringer:
@codex review

apedroheringer:
@codex review

apedroheringer:
@codex review

apedroheringer:
@codex review

apedroheringer:
@codex review

apedroheringer:
@codex review

apedroheringer:
Hi @provencher-oai — just checking in. All four requested changes are addressed (replies in each review thread). Let me know if anything else needs attention. Thanks!

PR #565, opened 2026-07-27
fix: honor stop_hook_active in the stop-review-gate hook

### LINKED ISSUE #548: Stop-review gate hook loops until CLAUDE_CODE_STOP_HOOK_BLOCK_CAP (missing `stop_hook_active` guard)
**File:** `scripts/stop-review-gate-hook.mjs` (v1.0.6)

## Problem

When `stopReviewGate` is enabled, `main()` runs the review and emits `{"decision":"block"}` whenever the review is not `ok` — including the non-`ok` cases where the review task returns no output, times out (15 min), fails, or returns invalid JSON.

After a Stop hook blocks, Claude Code re-runs the turn with `stop_hook_active: true` in the hook input. This hook never checks that flag, so it re-runs the review and blocks again on **every** retry. The result is the harness message:

> A hook blocked the turn from ending 9 consecutive times — overriding and ending turn. For Stop/SubagentStop hooks, check stop_hook_active in the input and return success while it's true.

The failure is worst in exactly the cases the gate is least useful — a Codex review that errors or times out blocks the session 9× before the cap force-ends it.

## Fix

Honor `stop_hook_active` before running the blocking review:

```js
if (input.stop_hook_active) {
  logNote(runningTaskNote);
  return;
}

const review = runStopReview(cwd, input);
```

This lets the gate review once; on the harness's forced retry it returns success instead of blocking again.

`codex-companion.mjs` has the same `config.stopReviewGate` branch — worth auditing it for the same issue.

mittalpk:
Fixes #548.

## Summary

When the review gate is enabled, `main()` in `stop-review-gate-hook.mjs` runs the stop-time Codex review and emits `{"decision":"block"}` on any non-`ok` outcome (no output, timeout, failure, invalid JSON) — without checking whether this invocation is already a forced retry.

Claude Code re-invokes Stop hooks with `stop_hook_active: true` after a `block` decision. Since this hook never checked that flag, a review that times out or errors would re-run and re-block on **every** retry, until the harness's forced-retry cap force-ends the turn — worst in exactly the cases where the gate is least useful (a broken/slow review blocking the session 9× before being overridden).

## Fix

Skip the review and return cleanly when `input.stop_hook_active` is set, matching the fix the issue itself suggested:

```js
if (input.stop_hook_active) {
  logNote(runningTaskNote);
  return;
}
```

The issue also flagged `codex-companion.mjs`'s `config.stopReviewGate` branches as "worth auditing... for the same issue." I checked — those are only config get/set/report code (`buildSetupReport`, `handleSetup`), not the review-blocking logic itself, which lives solely in `stop-review-gate-hook.mjs`. Confirmed via `grep -rl "decision.*block\|runStopReview\|stop_hook_active"` across `plugins/` that only this one file implements it, so no second file needs the fix.

## Tests

Added a regression test that:
1. Enables the review gate and snapshots the fake Codex binary's `appServerStarts` counter.
2. Sends the Stop hook input with `stop_hook_active: true`.
3. Asserts the hook exits with empty stdout (no block decision) and that `appServerStarts` is unchanged — proving the review subprocess was never spawned a second time, not just that the output happened to look right.

Confirmed the test **fails** against the pre-fix code via `git stash` isolation, reproducing the exact stale block decision the issue describes (`Codex stop-time review found issues that still need fixes...`).

## Verification

- `npm test`: 92/92 passed, no regressions.
- No lint config found in this repo (no `.eslintrc*`/`eslint.config.*`).
- `npm run build` (`tsc`) requires a live `codex app-server generate-ts` run to produce `.generated/app-server-types`, which needs an actual Codex CLI installation not available in this environment — this change touches no TypeScript files, so it's out of scope for this fix, but flagging honestly rather than silently skipping it.

### DIFF
--- a/plugins/codex/scripts/stop-review-gate-hook.mjs
+++ b/plugins/codex/scripts/stop-review-gate-hook.mjs
@@ -163,6 +163,16 @@ function main() {
     // failure, invalid JSON) re-blocks unconditionally -- until the harness's
     // forced-retry cap kicks in and ends the turn anyway. Skip the re-run and
     // let this retry succeed instead of repeating it up to the cap.
+    //
+    // logNote() alone is invisible to the user (it only writes to stderr, which
+    // isn't surfaced for a successful hook run) -- a skipped review would look
+    // identical to a passed one. Emit a systemMessage instead: it reaches the
+    // user without blocking (no "decision" key), so the skip is visible right
+    // when it matters -- the previous review blocked on something.
+    emitDecision({
+      systemMessage:
+        "Codex stop-gate review skipped on this continuation turn. Run /codex:review --wait to verify the fixes."
+    });
     logNote(runningTaskNote);
     return;
   }
--- a/plugins/codex/scripts/stop-review-gate-hook.mjs
+++ b/plugins/codex/scripts/stop-review-gate-hook.mjs
@@ -156,6 +156,17 @@ function main() {
     return;
   }
 
+  if (input.stop_hook_active) {
+    // Claude Code re-invokes the Stop hook with stop_hook_active: true when a
+    // prior invocation returned a "block" decision. Running the review again
+    // here would just block again -- every non-ok outcome (no output, timeout,
+    // failure, invalid JSON) re-blocks unconditionally -- until the harness's
+    // forced-retry cap kicks in and ends the turn anyway. Skip the re-run and
+    // let this retry succeed instead of repeating it up to the cap.
+    logNote(runningTaskNote);
+    return;
+  }
+
   const setupNote = buildSetupNote(cwd);
   if (setupNote) {
     logNote(setupNote);
--- a/tests/runtime.test.mjs
+++ b/tests/runtime.test.mjs
@@ -1979,6 +1979,50 @@ test("stop hook runs a stop-time review task and blocks on findings when the rev
   assert.match(status.stdout, /Codex Stop Gate Review/);
 });
 
+test("stop hook does not re-run the review and does not block on a forced retry (stop_hook_active)", () => {
+  const repo = makeTempDir();
+  const binDir = makeTempDir();
+  const fakeStatePath = path.join(binDir, "fake-codex-state.json");
+  installFakeCodex(binDir);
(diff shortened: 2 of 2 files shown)


Wintersta7e:
Confirming this reproduces, and I arrived at the same placement independently — putting the guard before `buildSetupNote` also keeps the retry turn from paying for the two `codex` availability probes.

One thing worth adding: **the skip is currently invisible to the user.** `logNote` writes to stderr, and a successful hook's stderr isn't surfaced — so from the user's side a skipped review looks identical to a review that passed. With the gate enabled that difference matters, since the whole reason there's a retry turn is that the previous review blocked on something.

Emitting a `systemMessage` does reach them, and doesn't block:

```js
if (input.stop_hook_active) {
  emitDecision({
    systemMessage:
      "Codex stop-gate review skipped on this continuation turn. Run /codex:review --wait to verify the fixes."
  });
  logNote(runningTaskNote);
  return;
}
```

No `decision` key means the turn still ends normally.

Separately — your `appServerStarts` snapshot is a better assertion than the one I had written. It proves no review turn was spawned, where I was inferring it from `lastTurnStart` being unset.

RossMora:
Ran into this exact gap independently and wanted to add a data point in support of merging.

We arm this hook on unattended, multi-day Claude Code sessions. On 2026-08-09, the missing stop_hook_active check let a real block re-fire 3 times inside a 22-minute window before an external circuit breaker we'd built specifically for this scenario stepped in and disarmed the gate -- without it, nothing in the plugin would have stopped the loop. Confirmed stop_hook_active is absent from both 1.0.4 and 1.0.6 by grepping the whole plugin.

The fix here is the same one we'd independently drafted, plus a regression test ours didn't have, so linking here rather than opening a duplicate. Hope this helps move it forward.

mittalpk:
Good catch, added in bdfcd86 — emits a systemMessage on skip now instead of only logging to stderr.

mittalpk:
A couple of others have independently hit this same gap (see comments above) - hoping this can get a maintainer look.

Issue #450, opened 2026-07-08
Shared/co-owned broker orphaned when an owning session exits without SessionEnd


hongsu:
## Context

PR #381 keys broker cleanup on the owning session(s): `broker.json` records `sessionIds`, a reused broker adds the reusing session as a co-owner, and `teardownBrokersForSession` shuts a broker down only once **no owner remains**. This correctly stops a single session from tearing down a broker its siblings are still using.

## Gap

If a co-owning session disappears **without** running its `SessionEnd` hook (SIGKILL, OOM, crash, host reboot), its `sessionId` stays in `broker.json` forever. When the other owners later exit gracefully, `teardownBrokersForSession` sees a remaining owner and preserves the broker — but no future hook will ever fire for the dead owner. The shared/worktree broker is orphaned indefinitely.

This is the same class as the original single-owner leak in #108, but surfaced by the new co-ownership model. Codex flagged it on the PR: https://github.com/openai/codex-plugin-cc/pull/381#discussion_r3535561589

## Why not fix it in the SessionEnd hook

PR #381 briefly recorded a per-owner session pid (`process.ppid` captured at `SessionStart`) and pruned owners whose pid was gone. That was reverted because:

- The packaged `hooks.json` runs `SessionStart` in **shell form** (no `args`), so `node`'s `process.ppid` is the ephemeral `sh` wrapper, not the Claude session. The recorded pid is dead almost immediately, which made **every** shared-broker co-owner look dead and tore down brokers still in use (regression flagged in https://github.com/openai/codex-plugin-cc/pull/381#discussion_r3535328451).
- Even with a correct anchor, `process.kill(pid, 0)` is unsound across PID reuse, and there is no cross-platform way to map a `sessionId` to a live process (macOS has no `/proc`; exec-form hooks aren't available in the hook schema).

A hook-side liveness signal is therefore not a reliable fix.

## Proposed resolution

Solve it broker-side with the **idle timeout** already proposed in #108: `app-server-broker.mjs` self-terminates after N minutes with no active client connection. This is platform-independent, needs no PID/liveness signal, and covers **both** the abnormal-exit orphan and the dead-co-owner orphan in one mechanism.

This issue tracks the co-owner facet specifically; it can be closed as covered once #108's idle timeout lands (they likely share one implementation).

## Related

- #108 — umbrella idle-timeout issue
- #380 / PR #381 — graceful cwd-mismatch teardown (the in-scope part, fixed)

[...]

Issue #416, opened 2026-07-01
Windows: zombie broker + codex app-server trees accumulate — broker/shutdown acks before unbounded cleanup, and taskkill tree-kill breaks under Git Bash SHELL


JethroTseng:
# Summary

On Windows (plugin v1.0.5), zombie broker + `codex app-server` process trees accumulate at scale — we found **16 brokers, 16 `codex app-server` children, and 16 shell-wrapper processes (48 processes total)** that had survived for 2 days. Every broker's named pipe was still alive and accepting connections. The visible symptom is the recurring startup/exit message:

```
SessionEnd hook [node "${CLAUDE_PLUGIN_ROOT}/scripts/session-lifecycle-hook.mjs" SessionEnd] failed: Hook cancelled
```

This is a *compound* failure chain. Individual links are partially covered by #288, #331, #403, #380, #410, but the ack-before-cleanup shutdown handler and the resulting child-tree leak are not reported anywhere, and it's the combination that makes brokers effectively immortal.

# Environment

- Windows 11 Pro (10.0.26200), plugin v1.0.5, Node from `C:\Program Files\nodejs`
- Brokers spawned from Claude Code Bash-tool background jobs, so their environment has `SHELL` = Git Bash (`sh.exe`)

# Observed evidence (verified)

1. 16 broker processes (`app-server-broker.mjs serve`) created 2026-06-30 → 07-01 were still running on 07-02. Each had a matching `codex app-server` child — wrapped in `powershell.exe -c "codex app-server"` or `sh.exe .../npm/codex app-server` (a consequence of `shell: process.env.SHELL || true` in `app-server.mjs:190-194`).
2. All 16 named pipes still connected successfully (tested via `NamedPipeClientStream`).
3. Sending `broker/shutdown` manually to one zombie broker returned `{"id":1,"result":{}}` in ~250 ms, but the process was **still alive 2+ seconds later** (it exited some minutes afterwards).
4. The SessionEnd hook, run manually against this accumulated state, took long enough to exceed the hook's own `timeout: 5` (hooks.json) — Node cold start alone costs ~1–2 s on Windows, leaving almost no budget for shutdown + teardown work. Hence the recurring `Hook cancelled`.
5. Because the hook is cancelled before teardown completes, stale `broker.json` and `cxc-*` session dirs persist, and the next SessionEnd repeats the same work — the failure is self-perpetuating.

# Root-cause chain (code analysis, v1.0.5)

[...]

PR #566, opened 2026-07-28
fix(broker): terminate the broker on teardown instead of only unlinking its pid file

srdrkr:
## The leak

The broker is spawned `detached: true` and `unref()`'d so it can outlive the process that started it. That is correct for a broker, but it makes explicit termination mandatory.

`teardownBrokerSession` unlinks the pid file, the log file and the endpoint socket. It only *terminates* the process when a caller injects `killProcess`, whose default was `null`:

```js
export function teardownBrokerSession({ ..., pid = null, killProcess = null }) {
  if (Number.isFinite(pid) && killProcess) {   // false when nothing is injected
    killProcess(pid);
  }
  // ... pid file, log file and socket are unlinked unconditionally
```

`session-lifecycle-hook.mjs` does pass `terminateProcessTree`, so session end was fine. **`ensureBrokerSession` does not.** Both of its teardown paths pass `options.killProcess ?? null`, and its only caller (`app-server.mjs:344`) passes just `{ env: options.env }`.

So two paths leaked:

1. **Stale-session replacement** — an existing session whose endpoint is not ready is torn down before a new one is spawned. The old broker survives.
2. **Spawn readiness timeout** — a broker that does not open its endpoint within the timeout is torn down and `null` is returned. The broker it just spawned survives.

In both cases the pid file is deleted while the process keeps running, so **nothing can ever find it again**. There is no reaper and no age-based sweep.

## Observed impact

On one workstation, after roughly a week of ordinary use:

- **206 orphaned `app-server-broker.mjs` processes**, all `PPID 1`, oldest cohort 7 to 9 days old, one per invocation
- ~1050 total user processes, 419 of them `node`
- New process spawns began failing with `fork failed: Device not configured`

Terminating just those 206 dropped the machine from 1047 processes to 565, since each broker also held child processes. Likely a contributor to battery drain as well.

## The fix

Default `killProcess` to `terminateProcessTree` in `teardownBrokerSession`, and stop coercing an absent injection to `null` at both `ensureBrokerSession` call sites. Injection still works unchanged, so the session hook and existing callers are unaffected.

Eight lines changed in one file.

## Tests

`tests/broker-lifecycle.test.mjs`, three cases. The behavioural one spawns a real stub process that never opens its endpoint, lets readiness time out, and asserts nothing survives afterwards. It deliberately injects **no** killer, because injecting one exercises the path that always worked.

Verified A/B/A against the unfixed tree:

```
unfixed:  teardown left 1 broker process(es) running after unlinking the pid file
fixed:    3 pass, 0 fail
unfixed:  fails again
```

Full suite on this branch: **90 pass, 4 fail**. Clean `main` at `db52e28`: **88 pass, 6 fail**. The same four failures are present on `main` and are unrelated to this change (`status shows phases…`, `status preserves adversarial review kind labels`, `result returns the stored output…`, `resolveStateDir uses a temp-backed per-workspace directory`). This branch fixes two and breaks none.

## Note on scope

This stops new orphans. It does not reap ones already on disk from earlier versions, which have to be cleared by hand. If you would like a startup sweep for pre-existing strays, I am happy to add it in a separate PR rather than widen this one.

### DIFF
--- a/plugins/codex/scripts/lib/broker-lifecycle.mjs
+++ b/plugins/codex/scripts/lib/broker-lifecycle.mjs
@@ -7,6 +7,7 @@ import { spawn } from "node:child_process";
 import { fileURLToPath } from "node:url";
 import { createBrokerEndpoint, parseBrokerEndpoint } from "./broker-endpoint.mjs";
 import { resolveStateDir } from "./state.mjs";
+import { terminateProcessTree } from "./process.mjs";
 
 export const PID_FILE_ENV = "CODEX_COMPANION_APP_SERVER_PID_FILE";
 export const LOG_FILE_ENV = "CODEX_COMPANION_APP_SERVER_LOG_FILE";
@@ -123,7 +124,7 @@ export async function ensureBrokerSession(cwd, options = {}) {
       logFile: existing.logFile ?? null,
       sessionDir: existing.sessionDir ?? null,
       pid: existing.pid ?? null,
-      killProcess: options.killProcess ?? null
+      killProcess: options.killProcess ?? terminateProcessTree
     });
     clearBrokerSession(cwd);
   }
@@ -154,7 +155,7 @@ export async function ensureBrokerSession(cwd, options = {}) {
       logFile,
       sessionDir,
       pid: child.pid ?? null,
-      killProcess: options.killProcess ?? null
+      killProcess: options.killProcess ?? terminateProcessTree
     });
     return null;
   }
@@ -170,7 +171,11 @@ export async function ensureBrokerSession(cwd, options = {}) {
   return session;
 }
 
-export function teardownBrokerSession({ endpoint = null, pidFile, logFile, sessionDir = null, pid = null, killProcess = null }) {
+export function teardownBrokerSession({ endpoint = null, pidFile, logFile, sessionDir = null, pid = null, killProcess = terminateProcessTree }) {
+  // The broker is spawned detached and unref'd so it can outlive the process that
+  // started it. That makes termination here mandatory rather than optional: this
+  // function also unlinks the pid file, so a broker left running after its pid file
+  // is gone can never be found again by anything.
   if (Number.isFinite(pid) && killProcess) {
     try {
       killProcess(pid);
--- a/tests/broker-lifecycle.test.mjs
+++ b/tests/broker-lifecycle.test.mjs
@@ -1,0 +1,111 @@
+import test from "node:test";
+import assert from "node:assert/strict";
+import fs from "node:fs";
+import os from "node:os";
+import path from "node:path";
+import { execFileSync } from "node:child_process";
+
+import {
+  ensureBrokerSession,
+  teardownBrokerSession
+} from "../plugins/codex/scripts/lib/broker-lifecycle.mjs";
+
+// The broker is spawned detached and unref'd, so it outlives the process that
+// started it by design. Teardown also unlinks the pid file. If teardown removes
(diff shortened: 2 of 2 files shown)


srdrkr:
I'm a lowly pm, hire me please!

afspies:
Bumping this — we're independently affected and would like to see it land. On one macOS machine the missing default kill (plus the cwd-keyed lookup in #380) had accumulated **25 orphaned brokers, 1.31 GB RSS, oldest 15 days**, 15 of them owned by no live session.

On the bot's P2, *"Verify stale broker PID before terminating it"* — that risk is real, but the suggested remedy (*"validate that the pid file still exists and matches the stored"*) is unreliable in practice, at least on macOS.

**`broker.pid` disappears on its own.** We found it missing from every `cxc-*` session dir older than ~3 days while `broker.sock` and `broker.log` were still present. Cause is macOS periodic cleanup of untouched files under `/var/folders/.../T` (confirmed via atime/mtime — the pidfile is never touched after creation, the socket is). Teardown reads the pid from `broker.json` rather than the pidfile, so this isn't a leak cause today, but it does mean a pidfile-existence check would refuse to terminate exactly the long-lived stale brokers this PR is meant to reap, on any machine that has been up a few days.

**An alternative that doesn't depend on the pidfile:** match the `--endpoint` path in the process's own argv. It's unique per broker (`mkdtemp`), so it identifies *this* instance rather than merely "a broker" — which matters, because a script-name-only check would happily accept a *different live broker* that inherited the recycled PID, and killing that drops another session's in-flight request.

Two details that turned out to matter:

1. **Make it tri-state and fail closed.** If `ps` itself fails, signal nothing *and* delete no tracking files, so a later teardown can retry. Deleting the pidfile/socket/session dir while failing to signal is precisely what converts a live broker into an untracked orphan — the bug this PR fixes — so an unverifiable pid should leave everything reapable rather than half-cleaned.
2. **Windows can't do this cheaply.** argv isn't readily readable there, so it stays permissive: no worse than current behaviour, but it cannot detect a recycled PID. Worth a comment rather than silence, given #416/#487.

```js
// "match" | "mismatch" | "unknown" — callers must treat "unknown" as fail-closed.
export function verifyBrokerProcess(pid, endpoint, platform = process.platform) {
  if (!Number.isFinite(pid)) return "mismatch";
  if (platform === "win32") return "unknown-permissive";

  const result = runCommand("ps", ["-p", String(pid), "-o", "command="]);
  if (result.error) return "unknown";          // ps unavailable: do not guess
  if (result.status !== 0) return "mismatch";  // no such process: definitively gone

  const command = result.stdout;
  if (!command.includes("app-server-broker.mjs")) return "mismatch";
  if (endpoint) {
    const socketPath = endpoint.startsWith("unix:") ? endpoint.slice("unix:".length) : endpoint;
    return command.includes(socketPath) ? "match" : "mismatch";  // not just *a* broker
  }
  return "match";
}
```

Then in `teardownBrokerSession`, `unknown` returns early without touching any file, and only `match` (or the Windows permissive case) reaches `killProcess(pid)`.

We've been running this together with idle self-termination (#450/#457/#490) and session-id teardown (#380/#381) locally for a day, with a regression harness covering the PID-identity case specifically: two live brokers, teardown called with broker A's pid and broker B's endpoint, asserting A survives. Happy to open a PR against this branch with the guard and that test if useful.

---

*Written by Claude (Claude Code) on behalf of @afspies, who reviewed and sent it. The measurements and file paths above are from a real machine, not inferred.*

PR #567, opened 2026-07-28
Fix rescue thread persistence across Claude session resume

Asher123452:
## Summary
- retain completed Codex task records when Claude's SessionEnd hook cleans up active work
- add a regression test covering task -> SessionEnd -> resumed task
- fall back to the tracked root process when Windows taskkill only partially terminates a tree
- make Windows runtime fixtures use platform-correct executable and home-directory environment names
- use POSIX path joining when constructing simulated/non-Windows Unix broker endpoints

## Why
Claude conversations preserve their session ID when reopened, but SessionEnd deleted every job for that ID. As a result, `/codex:rescue --resume` and the automatic resume candidate lookup reported that no previous Codex task existed even though the Codex thread itself still existed.

The full Windows run also exposed partial taskkill failures that made cancellation report failure after successfully interrupting the Codex turn, plus several test fixtures that assumed POSIX executable and HOME lookup behavior.

## Verification
- `npm test`: 93/93 pass on Windows
- live two-process Claude Code smoke test resumed the same Codex thread after SessionEnd
- resume candidate and stored result both retain the same thread ID
- syntax checks and `git diff --check` pass

### DIFF
--- a/plugins/codex/scripts/lib/broker-endpoint.mjs
+++ b/plugins/codex/scripts/lib/broker-endpoint.mjs
@@ -13,7 +13,7 @@ export function createBrokerEndpoint(sessionDir, platform = process.platform) {
     return `pipe:\\\\.\\pipe\\${pipeName}`;
   }
 
-  return `unix:${path.join(sessionDir, "broker.sock")}`;
+  return `unix:${path.posix.join(sessionDir, "broker.sock")}`;
 }
 
 export function parseBrokerEndpoint(endpoint) {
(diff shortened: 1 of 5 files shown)


Asher123452:
Local verification is complete: `npm test` passes 93/93 on Windows, including the new Claude SessionEnd -> reopened conversation -> Codex thread resume regression and repeated partial-`taskkill` cancellation coverage. A live two-process Claude Code smoke test also resumed the same Codex thread ID after SessionEnd.

The fork workflow is currently `action_required`, and the repository ruleset requires one approval from the configured OpenAI team. A maintainer will need to approve the Actions run and review this PR before it can squash-merge.

PR #572, opened 2026-07-31
feat: support max reasoning effort

MattSyska:
## What we are solving

Codex supports `max` reasoning effort, but the Claude Code bridge currently rejects `--effort max` before it reaches Codex. Users therefore cannot select the highest supported effort through `/codex:rescue`.

## Why

Max is useful for difficult bounded work and lets orchestrators exhaust a cheaper execution model's reasoning range before escalating to a more expensive model.

## Changes

- Accept `max` in the companion bridge validator and CLI help.
- Document `max` in the rescue command and runtime skill.
- Cover the accepted value and app-server passthrough in tests.

## Testing

- `npm test` — 91 tests passed
- `git diff --check`


dkundel-openai:
Duplicate of #616.

PR #577, opened 2026-08-01
fix: make cancel work under Git Bash on a non-English Windows (#423)

### LINKED ISSUE #423: `/codex:cancel` throws and leaves the job stuck as "running" when the pid is already dead on non-English Windows

xoonjaeho:
## Problem

Two defects make `/codex:cancel` unusable when Claude Code runs from Git Bash on a non-English Windows. The second one is invisible until the first is fixed, which is why they are in one PR.

**1. MSYS rewrites `taskkill`'s `/PID` flag into a path.**

`runCommand` spawns through `$SHELL` on Windows (`process.mjs:12`), which is Git Bash in this setup. MSYS then converts the `/PID` argument before `taskkill.exe` sees it:

```
$ node codex-companion.mjs cancel <job-id>
taskkill /PID 5548 /T /F: exit=1: ERROR: Invalid argument/option - 'C:/Program Files/Git/PID'
```

Every cancel fails. This is the same defect as #411 and #220, and this PR uses **#411's approach** — `MSYS_NO_PATHCONV=1` + `MSYS2_ARG_CONV_EXCL=*` scoped to the `taskkill` child only. Scoping matters: setting those on the node process instead breaks the `/c/...` path translation node needs to resolve its own modules.

**2. Dead-pid detection is English-only, so an already-stopped process throws.**

Once the flags survive, `taskkill` runs and reports "no such process" — but `looksLikeMissingProcessMessage` matches English text only. On a localized Windows nothing matches, so `terminateProcessTree` falls through to `throw new Error(formatCommandFailure(result))` instead of returning `delivered: false`. This is #423; #424 also addresses it but currently conflicts and carries unrelated changes.

The fix uses the exit code, which is locale-independent.

## Verification

Measured on ko-KR Windows 11, Git Bash, node v24.14.0:

| case | exit | message | English regex matches |
|---|---|---|---|
| live pid | `0` | `성공: PID 35580인 프로세스가 종료되었습니다` | — |
| dead pid | `128` | `오류: 프로세스 "35580"을(를) 찾을 수 없습니다` | **false** |

Before the env change, the same call fails with `Invalid argument/option - 'C:/Program Files/Git/PID'`. After it, the flags arrive intact.

Test suite on the same host: **92 tests / 83 pass / 9 fail**, against a pre-change baseline of **91 / 79 / 12**. No new failures, and three previously-failing tests now pass:

- `cancel stops an active background job and marks it cancelled`
- `cancel sends turn interrupt to the shared app-server before killing a brokered task`
- `session end fully cleans up jobs for the ending session`

(The 9 remaining failures are pre-existing on Windows — Unix-socket, temp-dir and real-CLI assertions. CI is ubuntu-only so they do not appear there.)

Both changes were mutation-checked: reverting either production line makes its test fail, and restoring makes it pass.

## Tests

- extended `terminateProcessTree uses taskkill on Windows` to assert the spawn env (from #411)
- added `terminateProcessTree treats exit code 128 as already stopped on a non-English Windows`

Both inject `platform`, `runCommandImpl` and `killImpl`, so they perform no real spawn and run identically on Linux CI.

## Relation to existing PRs

This supersedes **#411** (defect 1 only) and **#424** (defect 2 only, currently conflicting). Credit to #411 for the env approach — I verified it empirically and kept it as-is. Happy to rebase, split, or close this in favour of either if a maintainer prefers.

Fixes #423

---

## Why this matters beyond `cancel`

`terminateProcessTree` is not only reached from `/codex:cancel`. It is also the killer
`session-lifecycle-hook.mjs:105` passes as `killProcess` when the `SessionEnd` hook reaps a
broker. On a Windows host running Claude Code from Git Bash, that reap is silently a no-op
today for the same MSYS reason — `taskkill` never receives `/PID`.

That turns out to matter for the app-server broker leak. `broker-lifecycle.mjs` spawns brokers
`detached` + `unref`, the broker has no idle timeout, and the stale-teardown path at
`broker-lifecycle.mjs:120` calls `teardownBrokerSession` **without** `killProcess`, so a broker
that misses the 150 ms endpoint probe is replaced while the old process keeps running and its
state files are deleted — invisible to the reaper forever. On this host that produced four live
brokers for a single `cwd` where `broker.json` named one, and a per-orphan footprint of ~9
processes.

Wiring `terminateProcessTree` into that teardown is the natural fix for the leak — **but on
Windows + Git Bash it would still be a no-op without the patch in this PR.** So this change is a
prerequisite for that fix landing effectively on Windows, not just a `cancel` bug. Filing the
broker fix separately; flagging the dependency here so the two are not evaluated in isolation.

### DIFF
--- a/plugins/codex/scripts/codex-companion.mjs
+++ b/plugins/codex/scripts/codex-companion.mjs
@@ -983,7 +983,17 @@ async function handleCancel(argv) {
     );
   }
 
-  terminateProcessTree(job.pid ?? Number.NaN);
+  const termination = terminateProcessTree(job.pid ?? Number.NaN);
+  if (termination.attempted && !termination.treeConfirmed) {
+    // The job's own process is gone, but the kill could not account for what was under
+    // it. Say so rather than reporting a clean cancellation: a surviving codex or broker
+    // child keeps running, and nothing else makes a second pass at it.
+    appendLogLine(
+      job.logFile,
+      `Cancelled, but the process tree under pid ${job.pid} could not be fully confirmed stopped. ` +
+        "Check for leftover codex processes if the workspace behaves oddly."
+    );
+  }
   appendLogLine(job.logFile, "Cancelled by user.");
 
   const completedAt = nowIso();
--- a/plugins/codex/scripts/lib/process.mjs
+++ b/plugins/codex/scripts/lib/process.mjs
@@ -82,6 +82,24 @@ function processIsGone(pid, killImpl) {
   }
 }
 
+// A process that is already on its way out -- taskkill losing the race with a turn that
+// was just interrupted reports "Access is denied", exit 1 -- is still briefly alive, so a
+// single instantaneous probe calls it a failure. Give it a bounded moment to finish.
+// This whole function is synchronous, hence the Atomics sleep rather than a timer.
+function waitForProcessGone(pid, killImpl, timeoutMs) {
+  const deadline = Date.now() + timeoutMs;
+  const idle = new Int32Array(new SharedArrayBuffer(4));
+  for (;;) {
+    if (processIsGone(pid, killImpl)) {
+      return true;
+    }
+    if (Date.now() >= deadline) {
+      return false;
+    }
+    Atomics.wait(idle, 0, 0, Math.min(50, Math.max(1, deadline - Date.now())));
+  }
+}
+
 export function terminateProcessTree(pid, options = {}) {
   if (!Number.isFinite(pid)) {
     return { attempted: false, delivered: false, method: null };
@@ -109,13 +127,14 @@ export function terminateProcessTree(pid, options = {}) {
       return { attempted: true, delivered: false, method: "taskkill", result };
     }
 
-    // `/T` makes taskkill walk the tree, and it exits non-zero when it killed the target
-    // but could not reap one descendant -- "The operation attempted is not supported",
-    // exit 255, which is what a console host or an already-exiting child produces. The
-    // exit code does not separate that from a genuine failure, and the message is
-    // localized, so ask the OS whether the target itself is gone rather than parsing
-    // either. The target being dead is what every caller of this actually asked for.
-    if (!result.error && processIsGone(pid, killImpl)) {
+    // taskkill exits non-zero in two cases where the target still ends up dead: it killed
+    // the target but could not reap a descendant ("The operation attempted is not
+    // supported", exit 255), and it raced a process that was already terminating
+    // ("Access is denied", exit 1 -- what `cancel` hits after interrupting the turn).
+    // The exit code separates neither from a genuine failure, and both messages are
+    // localized, so ask the OS whether the target is gone instead of parsing either.
+    // That is what every caller of this actually asked for.
+    if (!result.error && waitForProcessGone(pid, killImpl, options.killWaitMs ?? 1000)) {
       return { attempted: true, delivered: true, method: "taskkill", result };
     }
 
--- a/plugins/codex/scripts/lib/process.mjs
+++ b/plugins/codex/scripts/lib/process.mjs
@@ -125,7 +125,12 @@ export function terminateProcessTree(pid, options = {}) {
       !result.error &&
       (result.status === TASKKILL_PROCESS_NOT_FOUND || looksLikeMissingProcessMessage(combinedOutput))
     ) {
-      return { attempted: true, delivered: false, treeConfirmed: true, method: "taskkill", result };
+      // "No such process" is about the root at this instant, not about what it spawned
+      // while it was alive. A worker that died mid-turn leaves its `codex app-server`
+      // child orphaned, and with the root gone `/T` has nothing left to walk from. That
+      // is not a corner case here: `cancel` only reaches this branch for a job the state
+      // still calls running, which is exactly the crashed-worker case.
+      return { attempted: true, delivered: false, treeConfirmed: false, method: "taskkill", result };
     }
 
     // taskkill exits non-zero both after failing to reap a descendant ("The operation
@@ -137,8 +142,8 @@ export function terminateProcessTree(pid, options = {}) {
     // But a non-zero taskkill never establishes that the REST of the tree went with it,
     // and `/T` leaves no way to enumerate what survived once the root is gone. So the
     // two facts are reported separately -- `delivered` is about the target, and
-    // `treeConfirmed` is about everything under it. This is the one branch that cannot
-    // account for the whole tree.
+    // `treeConfirmed` is about everything under it. Like the missing-root branch above,
+    // this one cannot account for the whole tree.
     if (!result.error && waitForProcessGone(pid, killImpl, options.killWaitMs ?? 1000)) {
       return { attempted: true, delivered: true, treeConfirmed: false, method: "taskkill", result };
     }
--- a/plugins/codex/scripts/lib/process.mjs
+++ b/plugins/codex/scripts/lib/process.mjs
@@ -8,6 +8,7 @@ export function runCommand(command, args = [], options = {}) {
     encoding: "utf8",
     input: options.input,
     maxBuffer: options.maxBuffer,
+    timeout: options.timeout,
     stdio: options.stdio ?? "pipe",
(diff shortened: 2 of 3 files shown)


xoonjaeho:
Pushed a second commit to this branch: `d5c0b31`, same function, adjacent case.

`taskkill /PID <n> /T /F` walks the tree, and it exits **255** when it killed the target but could not reap one descendant — *"The process with PID <child> (child process of PID <target>) could not be terminated. Reason: The operation attempted is not supported."* A console host or an already-exiting child produces it. `terminateProcessTree` tolerated only exit 0 and exit 128, so that threw, and the throw propagates out of `handleCancel` — `/codex:cancel` fails on Windows even though the process it was asked to kill is dead.

The exit code cannot separate a partial kill from a real failure, and that message is localized too, so keying on it would reintroduce exactly the bug the exit-128 check in this PR exists to avoid. It probes the target with signal 0 instead: gone ⇒ the kill did what the caller asked; still alive ⇒ throws as before. Two tests, each checked by breaking the line it protects.

How it was found is worth a line: `tests/runtime.test.mjs:1740` ("cancel sends turn interrupt to the shared app-server before killing a brokered task") had been written off as flaky on Windows. It is not flaky in that direction — it failed **2/2** in a clean environment, and the taskkill 255 above is why. After this commit the dominant failure mode is gone, though the test still fails intermittently under full-suite parallelism from some other cause I have not isolated yet, so I am not claiming it as a full fix for that test.

xoonjaeho:
The finding on this PR — reporting tree termination after checking only the parent — is real and fixed in `bfdf8db`.

You are right that the branch conflated two different facts. They are separate now: `delivered` answers "did the target die", and a new `treeConfirmed` answers "is everything under it accounted for". Exit 0 confirms both. Exit 128 means there was no process, so no tree is in doubt. Only the branch that tolerates a non-zero taskkill whose root turned out to be gone — exit 255's unreapable descendant, and exit 1's race with an already-exiting root — reports `treeConfirmed: false`.

Windows offers no way to enumerate what survived once the root is gone, so honest reporting is as far as this can go. To keep that from being a pure relabel, `handleCancel` now writes it to the job log instead of reporting a clean cancellation — nothing read the return value before, which was the substance of your point.

Also carries a `timeout` passthrough for `runCommand`, which had none; the broker work stacked on this branch needs it to bound a probe.

xoonjaeho:
The finding on `bfdf8db` — the missing-root branch reporting `treeConfirmed: true` — is real and fixed in `f9d718d`.

You were right that the previous round drew the line in the wrong place. `taskkill` exiting 128 says the pid is not alive *at that instant*; it says nothing about what the process spawned while it was. `/T` has nothing left to walk from once the root is gone, so the tree is unaccounted for and the branch now reports `treeConfirmed: false`.

One thing worth recording, since it also answers the obvious objection that this makes cancellation warn all the time: it does not, and the reason is upstream of `terminateProcessTree`. `resolveCancelableJob` (`lib/job-control.mjs:281`) only ever returns a job whose status is `queued` or `running`, and `handleCancel` writes `pid: null` after a cancel, so a repeat cancel returns `attempted: false` and never reaches the warning. Exit 128 on the cancel path therefore means "the state still calls this job running, but its recorded worker pid is dead" — a worker that died mid-turn. That is precisely the case where its `codex app-server` child (the direct spawn at `lib/app-server.mjs:190`, taken when there is no broker record) is orphaned and still running. Rare path, and the warning fires only there.

The POSIX side keeps `treeConfirmed: true` on `kill(-pid)` ESRCH: an empty process group is genuine evidence about non-detached descendants, which is more than Windows can offer here. A detached grandchild escapes the group either way, so that branch is weaker than it looks — noting it rather than changing it in this PR.

`tests/process.test.mjs` flips the assertion for the Korean-locale exit-128 case. Suite is unchanged otherwise: same 9 pre-existing failures before and after, none in these paths.

PR #578, opened 2026-08-01
feat: recover a lost result from the Codex rollout transcript

xoonjaeho:
## Problem

When a job ends without writing a result back to its record, `/codex:result` has nothing to show:

```
# Codex Result

Job: task-xxxx
Status: failed
Codex session ID: 019fbad1-b569-7d30-8bb1-e9ee525dde8d

No captured result payload was stored for this job.
```

This happens whenever the turn and the record get out of sync — the companion returns before the turn completes, the app-server turn is interrupted, the launcher dies. `status` and `result` are both dead ends at that point, so the leg reads as a total loss.

**It usually isn't.** Codex has already written every assistant message to its own transcript at `$CODEX_HOME/sessions/<YYYY>/<MM>/<DD>/rollout-<timestamp>-<threadId>.jsonl`, and the job record already carries `threadId`. Nothing new needs to be captured — the data is on disk and simply never read.

Recovering it has repeatedly turned a "lost" leg into a usable answer for me, including one genuine high-severity finding that would otherwise have been thrown away.

## Change

When `readStoredJob` yields no output and the job has a `threadId`, look up the transcript and render its last assistant message under an explicit heading:

```
## Recovered from the Codex transcript (PARTIAL)

No result was stored for this job, so this is the last assistant message
codex recorded. The turn may not have finished. Source: <path>

<message>
```

- **Read-only.** It never writes to the job record, and a job that *did* store output takes the existing path untouched — `storedJobHasOutput` gates it.
- **Labelled PARTIAL on purpose.** The turn may have been mid-thought; this is a lead, not a verdict.
- Also exposed on `--json` as `payload.recovered`.
- `resolveCodexHome()` was already implemented in `lib/codex.mjs`; this only exports it.

New file `lib/rollout.mjs` does the filesystem work so `render.mjs` stays pure.

## Verification

**Against real data, not just fixtures.** Run over the 832 real transcripts on this machine, sampling four threads:

| thread | recovered | lookup |
|---|---|---|
| `019fba9c-063a-70a1…` | 1506 chars | 4 ms |
| `019fbab1-07d6-7d83…` | 994 chars | 3 ms |
| `019fbace-b8a5-79c2…` | 547 chars | 1 ms |
| `019fbad1-b569-7d93…` | 1790 chars | 6 ms |

Every one returned that thread's real final message. The directory walk is flat and uncached, and still costs single-digit milliseconds at 832 files.

Four mutations, each caught by its own test:

| mutation | test that fails |
|---|---|
| take the first assistant message instead of the last | `returns the last assistant message, not the first` |
| let a malformed line throw instead of skipping | `survives the half-written final line a killed turn leaves` |
| drop the `role === "assistant"` filter | `returns null when the transcript holds no assistant text` |
| render ignores the recovered message | `surfaces a recovered message instead of the empty-payload notice` |

The malformed-line case is not hypothetical: a killed turn leaves a truncated final line, which is exactly the situation this feature exists for.

Suite: **98 tests / 86 pass / 12 fail**, against a **91 / 79 / 12** baseline on the same host. No new failures. (Those 12 are pre-existing Windows-only failures — Unix-socket, temp-dir and real-CLI assertions. CI is ubuntu-only and does not see them.)

## Notes

- Format confirmed empirically: assistant turns are `record.type === "response_item"` with `payload.type === "message"`, `payload.role === "assistant"`, and `payload.content` a list of `{ type: "output_text", text }`. String `content` is handled too.
- The thread id in the filename matches `session_meta.id`, so the `rollout-*-<threadId>.jsonl` match is exact rather than heuristic.
- Environment: Windows 11, node v24.14.0.

---

## Update — turn scoping, and using the turn's own final message

Three commits since the original review, all in this path. The first fixes the P2 the automated review raised on this PR.

**`b1b1697` — scope the recovery to the job's own turn.** `--resume-last` reuses a thread, so one transcript can hold turns from several jobs; 47 of 836 transcripts on my host do. Recovering by thread alone returned the file's *last* assistant message, which for an older job is a newer job's answer — worse than returning nothing. The job record already carries `turnId`, and it holds the same value as the transcript's `payload.turn_id`.

The assistant records themselves carry no turn id — only `task_started`, `turn_context` and `task_complete` do — so the scan tracks the turn those markers open and keeps only the messages inside the wanted one. A job with no recorded `turnId` now recovers nothing rather than guessing.

**`fe3514f` — prefer `task_complete.last_agent_message`, and stop mislabelling finished turns.**

- `task_complete` was setting `openTurnId` instead of clearing it, so records between one turn's end and the next turn's start were attributed to the finished turn.
- The raw assistant records keep codex's internal citation markup — present on 91 of 843 assistant messages I sampled. `task_complete.last_agent_message` is the same answer without it, and it is recorded for every completed turn (169/169 sampled). Prefer it; fall back to the scan when the turn never completed, or when that message is empty, which is what an errored turn records.
- Its presence also proves the turn finished, so a recovered answer no longer claims "the turn may not have finished" when it demonstrably did. Only the cut-short case keeps the **PARTIAL** label.
- `storedJobHasOutput` counted a whitespace-only stored result as output, so a job that stored `"\n"` suppressed the recovery that had the real answer.

**`7597426`** — `renderStoredJobResult`'s structured-review branch returned early on any truthy `rendered`, so a whitespace-only one printed just that whitespace and dropped the recovered text. Both now gate on the same `hasText`.

**Verification.** 12 tests in `tests/rollout.test.mjs`; each was checked by breaking the line it protects and confirming that test — and only that test — fails. Also exercised against two real (threadId, turnId) pairs from actual job records on disk: both return the right text with no citation markup, and a deliberately wrong `turnId` returns null.

### DIFF
--- a/plugins/codex/scripts/codex-companion.mjs
+++ b/plugins/codex/scripts/codex-companion.mjs
@@ -923,7 +923,10 @@ function handleResult(argv) {
   // only exist in codex's own rollout transcript. Recover the last one so a job
   // that ended without writing back is not a total loss.
   const threadId = storedJob?.threadId ?? job.threadId ?? null;
-  const recovered = storedJobHasOutput(storedJob) ? null : readLastAssistantMessage(threadId);
+  const turnId = storedJob?.turnId ?? job.turnId ?? null;
+  const recovered = storedJobHasOutput(storedJob)
+    ? null
+    : readLastAssistantMessage(threadId, { turnId });
   const payload = {
     job,
     storedJob,
--- a/plugins/codex/scripts/codex-companion.mjs
+++ b/plugins/codex/scripts/codex-companion.mjs
@@ -57,12 +57,14 @@ import {
   renderNativeReviewResult,
   renderReviewResult,
   renderStoredJobResult,
+  storedJobHasOutput,
   renderCancelReport,
   renderJobStatusReport,
   renderSetupReport,
   renderStatusReport,
   renderTaskResult
 } from "./lib/render.mjs";
+import { readLastAssistantMessage } from "./lib/rollout.mjs";
 
 const ROOT_DIR = path.resolve(fileURLToPath(new URL("..", import.meta.url)));
 const REVIEW_SCHEMA = path.join(ROOT_DIR, "schemas", "review-output.schema.json");
@@ -917,12 +919,18 @@ function handleResult(argv) {
   const reference = positionals[0] ?? "";
   const { workspaceRoot, job } = resolveResultJob(cwd, reference);
   const storedJob = readStoredJob(workspaceRoot, job.id);
+  // Nothing was stored: the turn may still have produced assistant messages that
+  // only exist in codex's own rollout transcript. Recover the last one so a job
+  // that ended without writing back is not a total loss.
+  const threadId = storedJob?.threadId ?? job.threadId ?? null;
+  const recovered = storedJobHasOutput(storedJob) ? null : readLastAssistantMessage(threadId);
   const payload = {
     job,
-    storedJob
+    storedJob,
+    recovered
   };
 
-  outputCommandResult(payload, renderStoredJobResult(job, storedJob), options.json);
+  outputCommandResult(payload, renderStoredJobResult(job, storedJob, recovered), options.json);
 }
 
 function handleTaskResumeCandidate(argv) {
--- a/plugins/codex/scripts/lib/codex.mjs
+++ b/plugins/codex/scripts/lib/codex.mjs
@@ -650,7 +650,7 @@ async function withDirectAppServer(cwd, fn) {
   }
 }
 
-function resolveCodexHome() {
+export function resolveCodexHome() {
   return path.resolve(process.env.CODEX_HOME || path.join(os.homedir(), ".codex"));
 }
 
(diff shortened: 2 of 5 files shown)


xoonjaeho:
The P2 on `codex-companion.mjs` (recovering by `threadId` alone can return a later job's answer for an older job) is real and is fixed in `b1b1697`.

Both premises checked out: jobs do store `turnId`, and `findLatestResumableTaskJob` does reuse `job.threadId` across jobs — 47 of the 836 transcripts on my host hold more than one turn.

One correction to the obvious fix, in case it helps: the assistant records carry no turn id of their own. Only `task_started`, `turn_context` and `task_complete` carry `payload.turn_id` (0 of 292 sampled assistant messages had one), so it needs turn *segmentation* rather than a per-record filter. `fe3514f` then goes one better and takes `task_complete.last_agent_message` for the wanted turn when the turn completed.

Details and verification in the PR description update above.

xoonjaeho:
Second finding on this PR (`render.mjs`, trimmed output checks) — real, and fixed in `eb8cf8b`.

You were right that `storedJobHasOutput` and the render branches had drifted apart. `7597426` had changed only the structured-review branch; the `result.rawOutput` / `codex.stdout` branch you pointed at still returned early on untrimmed truthiness, and so did the bare `rendered` fallback below it, which the comment did not mention but has the same defect. A job that stored `"\n"` printed that whitespace and discarded the recovered transcript message.

All of them now go through one `hasText`. When those two functions disagree about what "has output" means, one suppresses the recovery and the other throws it away.

Test covers all four stored shapes, and was checked by restoring the untrimmed check and confirming it fails.

PR #592, opened 2026-08-05
fix: saveState merges with on-disk jobs so stale snapshots cannot destroy concurrent writes

scompel:
`saveState` treats its in-memory snapshot as authoritative: it prunes any job present on disk but absent from the snapshot and unlinks that job's result file and log. Under concurrent same-workspace use (process A loads state; process B records job-B; process A saves), job-B's ledger entry, .json, and .log are destroyed. Reproduced directly; the included regression test fails on baseline and passes with the fix. Fix: merge the snapshot with the current on-disk jobs list by id (newer updatedAt wins, matching pruneJobs' idiom). Intentional removal is preserved via an explicit removedJobIds option (session-end cleanup updated); MAX_JOBS pruning unchanged. Dependency-free and cross-platform.

### DIFF
--- a/plugins/codex/scripts/lib/state.mjs
+++ b/plugins/codex/scripts/lib/state.mjs
@@ -89,10 +89,36 @@ function removeFileIfExists(filePath) {
   }
 }
 
-export function saveState(cwd, state) {
+function isNewerJob(candidate, incumbent) {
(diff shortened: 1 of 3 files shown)


dkundel-openai:
Duplicate of #460.

PR #648, opened 2026-08-14
Accept max and ultra reasoning efforts

apedroheringer:
## Problem

The Codex model catalog advertises reasoning efforts that the companion rejects.

With `codex-cli 0.147.0`, `~/.codex/models_cache.json` reports:

| model | supported efforts |
|---|---|
| `gpt-5.6-sol` | low, medium, high, xhigh, **max, ultra** |
| `gpt-5.6-terra` | low, medium, high, xhigh, **max, ultra** |
| `gpt-5.6-luna` | low, medium, high, xhigh, **max** |

`VALID_REASONING_EFFORTS` in `codex-companion.mjs` stops at `xhigh`, so the request dies locally, before reaching the API:

```
$ node scripts/codex-companion.mjs task --wait --fresh --model gpt-5.6-luna --effort max "Reply OK."
Unsupported reasoning effort "max". Use one of: none, minimal, low, medium, high, xhigh.
$ echo $?
1
```

The same request succeeds through the native CLI:

```
$ codex exec --sandbox read-only -m gpt-5.6-luna -c model_reasoning_effort="max" "Reply OK."
OK
```

So the ceiling is the plugin's, not the account's or the model's. `plan_mode_reasoning_effort = "max"` in `~/.codex/config.toml` works natively for the same reason.

## Change

- Add `max` and `ultra` to `VALID_REASONING_EFFORTS`.
- Build the error message from the set instead of repeating the list by hand — the two had already drifted apart once, and this keeps them from doing it again.
- Update the usage string, the `rescue` command `argument-hint`, and the `codex-cli-runtime` skill docs (the skill now also notes that support is per-model).

Per-model support still varies, and the API remains the authority: asking a model for an effort it does not advertise fails server-side, exactly as it did before this change. This only stops the companion from refusing efforts the models do support.

## Testing

```
$ npm test
# tests 91
# pass 91
# fail 0
```

Both assertions in `tests/commands.test.mjs` that pin the effort list were updated in the same commit.

### DIFF
--- a/plugins/codex/commands/rescue.md
+++ b/plugins/codex/commands/rescue.md
@@ -1,6 +1,6 @@
 ---
 description: Delegate investigation, an explicit fix request, or follow-up rescue work to the Codex rescue subagent
-argument-hint: "[--background|--wait] [--resume|--fresh] [--model <model|spark>] [--effort <none|minimal|low|medium|high|xhigh>] [what Codex should investigate, solve, or continue]"
+argument-hint: "[--background|--wait] [--resume|--fresh] [--model <model|spark>] [--effort <none|minimal|low|medium|high|xhigh|max|ultra>] [what Codex should investigate, solve, or continue]"
 allowed-tools: Bash(node:*), AskUserQuestion, Agent
 ---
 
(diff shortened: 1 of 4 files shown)


apedroheringer:
Withdrawing this — the local workflow it was meant to unblock is already covered by calling `codex exec` directly, so I'd rather not add review load for something we don't need on our side. The observation still stands if anyone wants to pick it up: `models_cache.json` advertises `max`/`ultra` for gpt-5.6-sol/terra (and `max` for luna), and `VALID_REASONING_EFFORTS` stops at `xhigh`. Thanks!

PR #652, opened 2026-08-17
fix: bound the lifetime of detached brokers and task workers

principalwater:
## Problem

Background task workers are spawned fully detached (`detached: true`, `stdio: "ignore"`, `child.unref()`), and the per-workspace broker they use outlives the client that started it. That is intentional — a background task should survive the session that queued it — but neither has a wall-clock ceiling, a parent to watch, or anything that reaps it. A wedged one therefore survives indefinitely, holding its `codex app-server` and every MCP server underneath it.

Found while investigating memory pressure on a 16 GB machine:

```
12 orphaned trees, 2–3 days old
51 MCP server processes held by them
37.8 GB combined physical footprint
27.3 / 28.6 GB swap in use
```

The MCP servers in question happened to be memory servers that load an ONNX model, so each cost several hundred MB — but the leak itself is independent of what the MCP servers do.

Verified along the way that this is *not* an fd-inheritance problem: `lsof` shows exactly one holder of each server's stdin write end, so `Stdio::piped()` hygiene is correct. The servers stay alive simply because the client that owns them never exits.

## Changes

**Broker idle shutdown.** After ten minutes with no connected client and no in-flight request or stream, the broker shuts itself down through its existing `shutdown()`. This also reclaims a broker orphaned by a crashed or timed-out SessionEnd hook, and the losing side of a concurrent `ensureBrokerSession` race.

**Broker shutdown is now atomic with respect to new clients.** `shutdown()` previously awaited `appClient.close()` before `server.close()`, so a client could connect in that window and reach a broker whose backend was already going away. The listener now closes and a `shuttingDown` flag is set before the first `await`, and connections arriving during shutdown are refused so the caller falls back to starting its own broker.

**Broker startup is bounded.** Until the broker is listening there is no idle timer, and the spawning client stops waiting after two seconds without killing anything (`killProcess` is `null` on the normal path). A wedged app-server or MCP startup would strand the whole tree permanently. The broker now bounds its own startup and takes its process group down if it overruns.

**Detached workers get a wall-clock ceiling.** On expiry the worker terminates its process group via the existing `terminateProcessTree`, after recording a terminal job status — otherwise the registry keeps a `running` entry pointing at a dead pid, which is what makes these leaks invisible in `status`. The default is deliberately long: this is a runaway guard, not a task deadline.

A parent watcher was considered and rejected: the worker's immediate parent exits right after enqueue by design, so watching it would kill healthy background tasks.

**`CodexAppServerClient.connect()` closes on failure.** If `initialize()` throws, the app-server has usually already been spawned and the caller never receives the object, so nothing can close it.

## Configuration

| Variable | Default | Purpose |
|---|---|---|
| `CODEX_BROKER_IDLE_SHUTDOWN_MS` | 10 min | Idle window before the broker exits |
| `CODEX_BROKER_STARTUP_TIMEOUT_MS` | 5 min | Ceiling on broker startup |
| `CODEX_TASK_WORKER_TTL_MS` | 24 h | Ceiling on a detached worker |

All three accept `0` to disable. Values are clamped to what `setTimeout` can actually hold — an operator asking for 30 days would otherwise overflow to about a millisecond and kill precisely the long-running work they meant to protect.

## Tests

`tests/lifecycle-limits.test.mjs` covers the defaults, overrides, disabling, rejection of unusable input, and the clamp. `node --test tests/lifecycle-limits.test.mjs` passes 8/8.

One pre-existing failure in `tests/state.test.mjs` (`resolveStateDir uses a temp-backed per-workspace directory`) reproduces on a clean checkout in this environment and is unrelated to these changes.

Not covered by automated tests, and worth a maintainer's eye: reconnect-during-shutdown, wedged broker startup, and real process-group termination all need integration coverage that does not exist here yet.

### DIFF
--- a/plugins/codex/scripts/app-server-broker.mjs
+++ b/plugins/codex/scripts/app-server-broker.mjs
@@ -26,12 +26,18 @@ const STREAMING_METHODS = new Set(["turn/start", "review/start", "thread/compact
 /** How long each step of a shutdown waits before giving up and going down without it. */
 const SHUTDOWN_GRACE_MS = 5000;
 
+/** How many just-completed turns to remember while their handoff may still be racing us. */
+const COMPLETED_HANDOFF_MEMORY = 32;
+
 /** A deadline that never keeps the event loop alive on its own. */
 function grace(ms = SHUTDOWN_GRACE_MS) {
   return new Promise((resolve) => {
     setTimeout(resolve, ms).unref?.();
   });
 }
+
+/** Whether the app-server — and with it every configured MCP server — has been started. */
+let backendStarted = false;
 
 function buildStreamThreadIds(method, params, result) {
   const threadIds = new Set();
@@ -100,6 +106,7 @@ async function main() {
   });
 
   const appClient = await CodexAppServerClient.connect(cwd, { disableBroker: true });
+  backendStarted = true;
   let activeRequestSocket = null;
   let activeStreamSocket = null;
   let activeStreamThreadIds = null;
@@ -159,6 +166,12 @@ async function main() {
       // request continuation assigns ownership. Remember it, or that continuation would take
       // ownership of a turn that is already over and hold the broker busy until the client leaves.
       if (threadId) {
+        // Only ever consumed by a handoff that is racing this notification. One that never comes —
+        // a failed request, say — would otherwise sit here for the life of the broker, so keep the
+        // set to the handful of turns that could plausibly still be in flight.
+        if (completedBeforeHandoff.size >= COMPLETED_HANDOFF_MEMORY) {
+          completedBeforeHandoff.delete(completedBeforeHandoff.values().next().value);
+        }
         completedBeforeHandoff.add(threadId);
       }
     }
@@ -459,9 +472,12 @@ async function main() {
 
 main().catch((error) => {
   process.stderr.write(`${error instanceof Error ? error.message : String(error)}\n`);
-  // By the time most failures reach here the app-server and its MCP servers are already running.
-  // The broker is detached, so exiting alone would leave that tree with no parent and no record —
-  // the leak this script is supposed to prevent. Take the group down with us.
-  terminateProcessTree(process.pid);
+  // Once the app-server is up it has its own MCP servers under it, and this process is detached —
+  // exiting alone would leave that tree with no parent and no record, the leak this script exists
+  // to prevent. Only then, though: a bad argument fails before anything was spawned, and there is
+  // nothing to take down.
+  if (backendStarted) {
+    terminateProcessTree(process.pid);
+  }
   process.exit(1);
 });
--- a/plugins/codex/scripts/app-server-broker.mjs
+++ b/plugins/codex/scripts/app-server-broker.mjs
@@ -29,6 +29,9 @@ const SHUTDOWN_GRACE_MS = 5000;
 /** How many just-completed turns to remember while their handoff may still be racing us. */
 const COMPLETED_HANDOFF_MEMORY = 32;
 
+/** How many abandoned turns to keep discarding notifications for. */
+const ABANDONED_THREAD_MEMORY = 32;
+
 /** A deadline that never keeps the event loop alive on its own. */
 function grace(ms = SHUTDOWN_GRACE_MS) {
   return new Promise((resolve) => {
@@ -140,6 +143,11 @@ async function main() {
    */
   async function abandonStream(threadIds) {
     for (const threadId of threadIds) {
+      // Bounded: an interrupt need not be followed by a completion, and an entry left here
+      // silently discards every future turn on that thread.
+      if (abandonedThreadIds.size >= ABANDONED_THREAD_MEMORY) {
+        abandonedThreadIds.delete(abandonedThreadIds.values().next().value);
+      }
       abandonedThreadIds.add(threadId);
       try {
         await appClient.request("turn/interrupt", { threadId });
@@ -401,12 +409,17 @@ async function main() {
           if (isStreaming) {
             const threadIds = buildStreamThreadIds(message.method, message.params ?? {}, result);
             const finishedAlready = [...threadIds].some((id) => completedBeforeHandoff.delete(id));
-            if (!sockets.has(socket)) {
+            if (finishedAlready) {
+              // Over before we got here: nothing to hand over, and nothing to abandon. Marking it
+              // abandoned now would be permanent — interrupting a finished turn need not produce
+              // another completion to clear the mark, and every later turn on that thread would
+              // have its notifications discarded, including the one its caller is waiting for.
+            } else if (!sockets.has(socket)) {
               // The client left while the turn was starting. Taking ownership on its behalf would
               // hold the broker busy for a socket nobody reads; leaving the turn running would let
               // its notifications reach the next client. Stop it instead.
               await abandonStream(threadIds);
-            } else if (!finishedAlready) {
+            } else {
               activeStreamSocket = socket;
               activeStreamThreadIds = threadIds;
             }
--- a/plugins/codex/scripts/app-server-broker.mjs
+++ b/plugins/codex/scripts/app-server-broker.mjs
@@ -8,7 +8,11 @@ import process from "node:process";
 import { parseArgs } from "./lib/args.mjs";
 import { BROKER_BUSY_RPC_CODE, CodexAppServerClient } from "./lib/app-server.mjs";
 import { parseBrokerEndpoint } from "./lib/broker-endpoint.mjs";
-import { clearBrokerSession, loadBrokerSession } from "./lib/broker-lifecycle.mjs";
+import {
+  clearBrokerSession,
+  loadBrokerSession,
+  teardownBrokerSession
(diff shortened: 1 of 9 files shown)


principalwater:
Pushed one more fix that no review comment asked for, because it is the same defect the bot found, in the function next door.

### What

`ensureBrokerSession()` tore down the existing broker's socket, pid file, log and directory whenever the readiness probe failed, then spawned a replacement. On the normal path `options.killProcess` is `null`, so that teardown never stopped the old process — it only deleted the files that made it reachable and findable.

So a broker that was merely busy became an untracked, unreachable process still holding its `codex app-server` and every MCP server underneath, with a duplicate started next to it. That is precisely the leak this PR exists to close, produced by the cleanup path itself.

### How it turned up

While addressing "Recheck the broker record before deleting it" I added a liveness guard to the *reuse* path, then checked whether anything else made the same assumption. `ensureBrokerSession` did, and worse: the reuse path only declines to use a broker, while this one actively replaces it.

### Why it is fixed here and not left alone

I had held off on it earlier, because neither obvious option was safe on its own:

- Killing the recorded pid risks pid reuse — terminating a process group on the strength of a stale record can hit someone else's processes.
- Leaving a live broker alone used to mean stranding it forever.

The second option is safe now, and only because of this PR: a broker with no clients retires itself through the idle shutdown added here, tearing down its own socket, pid file and directory on the way out. Once the replacement takes over, the old one has no clients.

So teardown now waits for proof the process is gone, and a live one is left exactly as it is.

### Coverage

`tests/broker-reclaim.test.mjs` plants a session whose files exist on disk and runs `ensureBrokerSession` against an endpoint that never answers:

- with `pid: process.pid` — unmistakably alive — the socket, pid file and log must survive;
- with a pid that cannot be running, they must all be reclaimed along with the record.

Verified by reverting the guard: the first case then fails with `pid file was removed`.

One residual, deliberately left: a broker that is replaced while still alive keeps its `broker.log`, because the broker process is not told its own log path and its record by then points at the successor. It costs a few kilobytes and no process, and inventing a path convention inside the broker to chase it seemed worse than the leak.

principalwater:
@codex review

principalwater:
@codex review

principalwater:
I put this branch through a full review pass, which turned up more of the family it is about — cleanup and timers that leak on their own — including one this branch introduced. The tractable ones are fixed in the commit above. The rest are structural, predate this branch, and I have deliberately left them alone; they are listed here so they are not lost.

### Fixed in the commit above

**An abandoned turn followed whoever connected next.** This one was mine. Declining to hand a stream to a client that disconnected mid-turn stopped the broker from staying busy forever, but the turn itself kept running, and routing follows current ownership rather than the turn that produced the notifications — so they were delivered to the next client to connect. Such a turn is now interrupted and its notifications dropped.

**A turn that completed before its stream was handed over still took ownership.** The app-server can put a response and its `turn/completed` in one chunk, which lands before the request continuation runs. Ownership was then taken for a turn that was already over, and held until the client disconnected.

**Requests registered after the transport reported its exit never settled.** A socket can die between a successful connect and the next request; anything registered afterwards waited for an exit that had already been delivered.

**Shutdown could re-enter itself, be re-armed while running, and wait forever on a wedged app-server** — which is to say the exact backend this guard exists to reclaim could outlive the guard. It is now a single pass with a grace period.

**Fatal errors after the app-server was spawned left the tree behind.** A bare `null` is valid JSON and threw while being dereferenced, taking down this detached process and stranding its tree; `listen` failures never reached `main()`'s handler.

**The worker TTL wrote the day-old snapshot** the timer closed over, and one failed log write skipped the terminal status with it.

### Left for maintainers

These need decisions about the broker's concurrency model rather than another guard, and I did not want to make them inside a PR about lifetime limits.

**Session state has no locking or identity.** `load → probe → start → save` is not atomic, `saveBrokerSession` writes in place rather than via temp-and-rename, and a record is matched by endpoint string. Two callers racing can each end up deleting the other's record or reading a half-written one. A per-workspace lock plus a broker nonce would close the whole class; the conditional deletes in this branch narrow the windows but cannot remove them.

**`isProcessAlive` proves a pid exists, not that it is the broker.** After an abrupt exit the pid may have been reused, in which case artifacts are kept that should be reclaimed — and `SessionEnd` would signal a process group that is no longer ours. Identity needs to be recorded and verified.

**`SessionEnd` shuts down a workspace-shared broker unconditionally.** Two sessions in one workspace: when the first ends, it stops a broker the second is actively using. Now that idle shutdown exists, the forced shutdown may simply be unnecessary; otherwise this needs leases or a refcount.

**Idle is measured by connected sockets, not activity.** A client that connects and then goes quiet holds the whole tree indefinitely, because a socket in the set counts as busy before it has issued anything.

**Ownership is a single socket pointer.** Two concurrent requests on one socket, or a failed non-streaming request, can release ownership belonging to work still in flight. An operation token would model this properly.

**The worker TTL does not reach the shared broker,** contrary to what its comment implies: the broker runs in its own process group, so `terminateProcessTree` from the worker does not include it. Its turn keeps running until the broker's own idle timer, and a new client can cancel that timer. Interrupting the turn before terminating the worker would cover it, but killing a shared broker outright is not safe.

**A worker can read its job record before the parent writes it,** exit, and leave a `queued` job pointing at a dead pid, with no TTL ever armed.

### Note on coverage

None of the runtime races here have automated coverage, and I have not added any: exercising them needs a live broker and app-server, and there is no harness for that in this repo. What is covered is the pure logic — the limit parsing and clamping, the process-liveness predicate, connect-failure cleanup, and the reclaim guard around a live versus dead broker. Everything else was reasoned through by hand, and I would treat the structural list above as unproven until something can actually drive a broker under concurrency.

principalwater:
@codex review

principalwater:
@codex review

principalwater:
This is where I am stopping, and I think the reason is worth stating plainly, because it is also the argument for where the remaining work belongs.

### Where the branch stands

It closes the leak it set out to close — the detached worker and the broker both have a bounded lifetime now, and the shutdown path reclaims its own process, artifacts and record — plus every hang and orphan path reachable without changing how the broker models ownership. The runtime suite matches `main` exactly: 58 passing, the same 3 failing on both.

### Why I am not continuing

The review bot has now found a defect in freshly written code five rounds running, and the last two were mirror images of each other:

- A turn abandoned after it had already finished, poisoning its thread permanently.
- A completion remembered for a turn nobody was waiting on, poisoning the next turn on that same thread.

Both are the same shape. Ownership is a single socket pointer, and completion is matched against whatever that pointer happens to hold, so every guard that fixes one direction leaves the symmetric case open in the other. I can keep closing them one at a time, but the supply is a property of the model, not of my attention.

That is the argument for the structural list in my earlier comment rather than a caveat about it. An operation token — ownership belonging to a turn rather than to whichever socket is current — collapses most of that list at once, and it is the piece I deliberately did not build here: it changes whether the broker is single-client by design, and that is your call, not something a lifetime-limits PR should decide.

### What that leaves you

Everything in this branch is a bounded guard around behaviour you already have. Nothing changes the broker's concurrency model, the shared-broker semantics, or what `SessionEnd` means. If you would rather take the leak fix and rework ownership yourselves, the last few commits are separable — the worker TTL, the broker idle shutdown and the startup bound stand alone, and the ownership-race commits sit on top of them.

Two things worth knowing before merging:

Runtime races here have no automated coverage and I did not add any: exercising them needs a live broker against an app-server, and there is no harness for that. `tests/fake-codex-fixture.mjs` looks like most of what one would need, and if you would find such a harness useful I am happy to contribute it separately — it would also let the structural work be verified rather than reasoned about, which is the part I could not do here.

The three failures shared with `main` are environmental in my setup, not something this branch touches. Worth confirming they are green in your CI.

principalwater:
@codex review

principalwater:
@codex review

Edo771977:
**`isBrokerEndpointReady()` always `false` on Windows — the persisted endpoint is never reused**

`isBrokerEndpointReady()` (added in this PR) checks `target.kind === "unix" && fs.existsSync(target.path)`. On Windows the broker endpoint is a named pipe (`pipe:\.\pipe\...`, see `createBrokerEndpoint()`), so `target.kind` is `"pipe"`, not `"unix"` — the condition is always `false`, regardless of whether the broker is alive and listening.

Practical effect: in `CodexAppServerClient.connect()` with `reuseExistingBroker: true` (used by `interruptAppServerTurn`), `isBrokerEndpointReady(persisted)` always fails → the persisted endpoint is never reused → a direct `SpawnedCodexAppServerClient` is created instead, separate from the shared broker. A `turn/interrupt` sent this way reaches an app-server that never saw the original turn: it gets recorded as "received" but interrupts nothing, and the real turn keeps running on the shared broker until its own natural timeout.

Reproducible with a test that launches a background job, then calls `interruptAppServerTurn` with `reuseExistingBroker: true` on Windows — the turn never actually gets interrupted.

Suggested fix: have `isBrokerEndpointReady()` do a real cross-platform connection probe (e.g. reusing `waitForBrokerEndpoint()`/`probeEndpoint()`, already present in the file) instead of the filesystem check limited to Unix sockets.

PR #659, opened 2026-08-19
fix: SessionEnd/status lookups miss state written under a different CLAUDE_PLUGIN_DATA root, orphaning brokers

### LINKED ISSUE #636: SessionEnd cannot find the broker when CLAUDE_PLUGIN_DATA differs between spawn and teardown — same cwd, same hash, different state root
## Summary

`resolveStateDir()` picks the state root from `CLAUDE_PLUGIN_DATA`, falling back to
`$TMPDIR/codex-companion` when that variable is absent. The workspace slug and hash are
identical in both cases — only the **root** differs. When a broker is registered while
the variable is unset and `SessionEnd` later runs with it set (or vice versa),
`loadBrokerSession(cwd)` reads the *other* root, finds nothing, and the broker is
orphaned permanently.

This is **not** #380. There, a different cwd produces a different hash. Here the cwd and
the hash are identical — `wienerdog-cc76dfe898ef0cd4` exists in **both** roots on my
machine — and the lookup still misses.

## Mechanism

`scripts/lib/state.mjs`:

```js
const FALLBACK_STATE_ROOT_DIR = path.join(os.tmpdir(), "codex-companion");
// ...
const pluginDataDir = process.env[PLUGIN_DATA_ENV];      // CLAUDE_PLUGIN_DATA
const stateRoot = pluginDataDir ? path.join(pluginDataDir, "state") : FALLBACK_STATE_ROOT_DIR;
return path.join(stateRoot, `${slug}-${hash}`);
```

Same cwd, two answers:

```
$ CLAUDE_PLUGIN_DATA=~/.claude/plugins/data/codex-openai-codex \
    node -e 'import("…/lib/state.mjs").then(m=>console.log(m.resolveStateDir("…/wienerdog")))'
/Users/…/.claude/plugins/data/codex-openai-codex/state/wienerdog-cc76dfe898ef0cd4

$ env -u CLAUDE_PLUGIN_DATA \
    node -e 'import("…/lib/state.mjs").then(m=>console.log(m.resolveStateDir("…/wienerdog")))'
/var/folders/…/T/codex-companion/wienerdog-cc76dfe898ef0cd4
```

**There is no second chance.** `handleSessionEnd` falls back to the endpoint env var:

```js
const brokerSession =
  loadBrokerSession(cwd) ??
  (process.env[BROKER_ENDPOINT_ENV] ? { endpoint: process.env[BROKER_ENDPOINT_ENV], … } : null);
```

but `CODEX_COMPANION_APP_SERVER_ENDPOINT` is never exported to the hook environment —
`handleSessionStart` propagates only `CODEX_COMPANION_SESSION_ID`,
`CODEX_COMPANION_TRANSCRIPT_PATH` and `CLAUDE_PLUGIN_DATA` via `appendEnvVar`. So once
the root lookup misses, `endpoint` is `null`, no `broker/shutdown` is sent, and
`teardownBrokerSession` gets nothing to kill.

## Evidence from a real machine

One workspace, split across both roots:

```
A) $TMPDIR/codex-companion/wienerdog-cc76dfe898ef0cd4/
   broker.json   Jul 19 02:12     <- pid 68646
   jobs/         Jul 19 19:58     (64 entries)
   state.json    Jul 19 20:01     (32 KB)

B) ~/.claude/plugins/data/codex-openai-codex/state/wienerdog-cc76dfe898ef0cd4/
   jobs/         Aug 10 00:35     (empty)
   state.json    Aug 10 00:35     (80 bytes)
   (no broker.json)
```

`broker.json` in root A still records `"pid": 68646`, and that process **is alive after
23 days**, with zero clients connected (verified via `lsof` on its unix socket: only the
broker itself holds the listening fd).

The 64 job entries and the 32 KB `state.json` in root A show this was not a
first-run artifact — real work ran against the fallback root for ~18 hours.

**Inferred sequence** (consistent with all timestamps, not directly observed): the
workspace was revisited on Aug 10 with `CLAUDE_PLUGIN_DATA` set. `loadBrokerSession(cwd)`
resolved to root B, found nothing, so `existing` was `null` — `ensureBrokerSession`
skipped the teardown branch entirely and spawned a **second** broker. That one was reaped
normally at `SessionEnd` (no `broker.json` remains in root B), while pid 68646 in root A
was never referenced again by any code path.

So the root split does not merely orphan a broker — it also causes a duplicate broker to
be spawned for the same workspace, since `ensureBrokerSession` cannot see the existing one.

## Why the existing issues don't cover this

- **#380** — cwd mismatch → different hash. Here the hash matches exactly; the root differs.
- **#543 / #108** — broker never self-terminates / no idle timeout. Correct and complementary:
  an idle timer would eventually reap this orphan, but the lookup defect would still cause
  duplicate brokers and split job state on every affected session.
- **#288** — `sendBrokerShutdown` has no timeout. Different failure point; here the shutdown
  is never attempted at all because `endpoint` is `null`.

## Suggested fixes

1. **Make the state root deterministic.** Resolve it once and record the chosen root inside
   `broker.json`, or drop the `CLAUDE_PLUGIN_DATA` conditional and always use a fixed
   location. A root that depends on ambient environment cannot be relied on across a
   process boundary.
2. **Fall back across roots on lookup.** `loadBrokerSession(cwd)` could check the alternate
   root when the primary misses. Cheap, and fixes existing orphans on the next session.
3. **Export the broker endpoint to the hook environment** so the `BROKER_ENDPOINT_ENV`
   fallback in `handleSessionEnd` is actually reachable. Today it is dead code for hooks.
4. Independently, the idle self-exit proposed in #543 would bound the damage from this and
   every other lookup-side failure.

## Environment

- macOS (darwin 25.5.0), Node v24.18.0
- codex plugin 1.0.6 (`openai/codex-plugin-cc`)
- 9 orphaned broker chains total on this machine, ~5.1 GB RSS, 0 clients on all 9

mittalpk:
Fixes #636.

## What's wrong

`resolveStateDir()` picks the state root from `CLAUDE_PLUGIN_DATA` when set, falling back to `$TMPDIR/codex-companion` when it's absent — same workspace slug/hash either way, only the root differs. State written under one root (e.g. a broker registered while the var was unset) becomes invisible to any later lookup that resolves to the other root, since nothing checked both. For a broker specifically, that means `SessionEnd` never finds it to shut down — it's orphaned permanently, and since `ensureBrokerSession` also can't see it, the next session spawns a duplicate broker for the same workspace, compounding the leak. The same mechanism affects job/status state (`state.json`, individual job detail files), not just the broker.

## Fix

Reads now check every candidate root (current primary, then the tmpdir fallback), not just the current invocation's primary — writes are unchanged, still going to the primary root. Applied consistently to `loadState()` (job list, status, config), `readStoredJob()` (individual job detail lookups), and `loadBrokerSession()`/`clearBrokerSession()`.

**Known remaining asymmetry, called out honestly:** this fixes the direction with concrete evidence in the issue — state written while `CLAUDE_PLUGIN_DATA` was unset, later missed by a lookup that has it set. The reverse (written while set, later looked up unset) isn't fixable this way: an unset env var carries no trace of what value it previously held, so there's nothing to check beyond the always-known tmpdir fallback. Closing that direction fully would need something like your suggested fix #1 (persist the resolved root, e.g. inside `broker.json`) — happy to take a pass at that too if it's wanted, but wanted to keep this PR to the achievable, lower-risk fix first.

## Testing

Added regression tests for `loadState`, `readStoredJob`, and `loadBrokerSession`/`clearBrokerSession` each finding/clearing state across the root split; confirmed all fail on unpatched code. Full suite: 95 passed, 0 regressions.

### DIFF
--- a/plugins/codex/scripts/lib/broker-lifecycle.mjs
+++ b/plugins/codex/scripts/lib/broker-lifecycle.mjs
@@ -6,7 +6,7 @@ import process from "node:process";
 import { spawn } from "node:child_process";
 import { fileURLToPath } from "node:url";
 import { createBrokerEndpoint, parseBrokerEndpoint } from "./broker-endpoint.mjs";
-import { resolveStateDir } from "./state.mjs";
+import { resolveStateDir, resolveStateDirCandidates } from "./state.mjs";
 
 export const PID_FILE_ENV = "CODEX_COMPANION_APP_SERVER_PID_FILE";
 export const LOG_FILE_ENV = "CODEX_COMPANION_APP_SERVER_LOG_FILE";
@@ -73,17 +73,27 @@ function resolveBrokerStateFile(cwd) {
   return path.join(resolveStateDir(cwd), BROKER_STATE_FILE);
 }
 
-export function loadBrokerSession(cwd) {
-  const stateFile = resolveBrokerStateFile(cwd);
-  if (!fs.existsSync(stateFile)) {
-    return null;
-  }
+// The state root is derived from ambient environment (CLAUDE_PLUGIN_DATA),
+// which can differ between the invocation that registered a broker and a
+// later one that looks it up -- checking every candidate root, not just the
+// current invocation's primary, is what keeps a broker registered under one
+// root from being orphaned by a lookup that resolves to the other.
+function resolveBrokerStateFileCandidates(cwd) {
+  return resolveStateDirCandidates(cwd).map((stateDir) => path.join(stateDir, BROKER_STATE_FILE));
+}
 
-  try {
-    return JSON.parse(fs.readFileSync(stateFile, "utf8"));
-  } catch {
-    return null;
+export function loadBrokerSession(cwd) {
+  for (const stateFile of resolveBrokerStateFileCandidates(cwd)) {
+    if (!fs.existsSync(stateFile)) {
+      continue;
+    }
+    try {
+      return JSON.parse(fs.readFileSync(stateFile, "utf8"));
+    } catch {
+      continue;
+    }
   }
+  return null;
 }
 
 export function saveBrokerSession(cwd, session) {
@@ -93,9 +103,10 @@ export function saveBrokerSession(cwd, session) {
 }
 
 export function clearBrokerSession(cwd) {
-  const stateFile = resolveBrokerStateFile(cwd);
-  if (fs.existsSync(stateFile)) {
-    fs.unlinkSync(stateFile);
+  for (const stateFile of resolveBrokerStateFileCandidates(cwd)) {
+    if (fs.existsSync(stateFile)) {
+      fs.unlinkSync(stateFile);
+    }
   }
 }
 
--- a/plugins/codex/scripts/lib/broker-lifecycle.mjs
+++ b/plugins/codex/scripts/lib/broker-lifecycle.mjs
@@ -82,18 +82,31 @@ function resolveBrokerStateFileCandidates(cwd) {
   return resolveStateDirCandidates(cwd).map((stateDir) => path.join(stateDir, BROKER_STATE_FILE));
 }
 
-export function loadBrokerSession(cwd) {
+// The single source of truth for which candidate is "the" active broker
+// session: the first one that both exists *and* parses. loadBrokerSession()
+// and clearBrokerSession() both build on this so they always agree -- if
+// clearBrokerSession() instead selected by existence alone, a malformed
+// primary file next to a valid fallback one would make it delete the
+// (malformed, unused) primary while loadBrokerSession() actually returned
+// and a caller tore down the fallback broker, leaving that broker's now-
+// stale record behind.
+function selectBrokerState(cwd) {
   for (const stateFile of resolveBrokerStateFileCandidates(cwd)) {
     if (!fs.existsSync(stateFile)) {
       continue;
     }
     try {
-      return JSON.parse(fs.readFileSync(stateFile, "utf8"));
+      const session = JSON.parse(fs.readFileSync(stateFile, "utf8"));
+      return { stateFile, session };
     } catch {
       continue;
     }
   }
   return null;
+}
+
+export function loadBrokerSession(cwd) {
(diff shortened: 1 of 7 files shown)


mittalpk:
Thanks, all three caught real issues -- fixed in the two follow-up commits:

- `clearBrokerSession()` deleting every candidate instead of just the selected one — fixed.
- `loadState()` returning only the first candidate's job list instead of merging — jobs are a growing collection where both roots can legitimately have real, non-conflicting entries, so I merge them now (keeping the more recent copy if an id somehow collides).
- `clearBrokerSession()` selecting by file existence while `loadBrokerSession()` selects by existence *and* successful parse — they now share one selection helper so they can't disagree.

Added regression tests for each, confirmed they fail against the prior code. Full suite: 98 passed.

mittalpk:
Right again, and this one was more severe than the earlier finding it followed from — merging reads across roots wasn't enough on its own, since `saveState()` only ever wrote the new job list back to the current primary root. A job originating in the other root that got filtered out (e.g. `cleanupSessionJobs()` during `SessionEnd`) never actually disappeared: that root's own `state.json` kept its own untouched copy, and the next `loadState()` merged it right back in.

Fixed: `saveState()` now also prunes every other candidate root's own file down to the same retained job-id set. New/updated jobs are unaffected — still only ever written to the primary root — this only ever removes. Added a regression test mirroring `cleanupSessionJobs()`'s exact pattern (load merged view, drop one job that originated in the other root, save the remainder), confirmed it fails against the prior code. Full suite: 99 passed.

mittalpk:
Both correct, thanks. The SessionEnd check was a redundant primary-only pre-check ahead of the already-candidate-aware loadState() -- removed it, since the existing empty-jobs short-circuit already covers the same case correctly. For the config gate, merged it across roots the same way jobs already are, with an enabled boolean winning over a stale disabled one elsewhere -- a safety gate shouldn't get silently downgraded by which root happens to be primary for a given invocation.

mittalpk:
Following up here if there's time to take a look.

PR #660, opened 2026-08-19
Reap leaked broker sessions across working directories

siimvene:
## Problem

Broker sessions leak when the plugin is used across more than one working directory. A broker is spawned **per cwd** (`ensureBrokerSession` keys its state by cwd), but teardown only ran for the current cwd at `SessionEnd`, so brokers for every other cwd were never cleaned up, accumulating as orphaned `broker` + `app-server` processes plus leftover `os.tmpdir()/cxc-*` socket directories (tens of processes / ~1 GB after a few days of multi-repo use). They also contend on the shared codex log store, which can make later app-server startups hang.

## Approach

Give the broker ownership of its own lifecycle instead of having sessions guess when it is safe to kill it.

- **`app-server-broker.mjs`: the broker ends itself, and only itself.** It exits after `CODEX_BROKER_IDLE_MS` (default 30 min, strict-integer parsing, `<= 0` disables) with no connections, no in-flight work, and no active stream; it also exits when its codex app-server client exits (a dead child would otherwise leave an unkillable listening zombie) and when a forwarded request exceeds `CODEX_BROKER_REQUEST_TIMEOUT_MS` (default 10 min, a wedged app server). On every exit path it retires its own `broker.json` record (guarded on the endpoint still matching) and removes its own session directory; recursive removal is restricted to `cxc-` dirs directly under the OS temp dir. All shutdown paths share one promise, so a signal cannot abort cleanup mid-flight.
- **`broker-lifecycle.mjs`: dead-only GC.** `reapBrokerSessions()` removes only directories whose recorded broker PID is provably dead. It never inspects or signals a live PID, ignores directories without a readable valid `broker.pid` (an empty or torn file may be a broker mid-startup), and never touches non-broker `cxc-*` dirs.
- **`session-lifecycle-hook.mjs`: sessions leave shared state alone.** `SessionStart`/`SessionEnd` run the dead-only GC. `SessionEnd` no longer kills brokers or clears the shared `broker.json`; it interrupts the Codex turns owned by its own background jobs (mirroring `/codex:cancel`, bounded by a 2s budget under the hook's 5s timeout) before terminating their runners.
- **Consumers validate instead of trusting the record.** `reuseExistingBroker` probes the recorded endpoint before use; `getSessionRuntimeStatus` judges liveness by the recorded broker pid; `withAppServer` retries direct on any connect-phase failure, so racing an idle shutdown costs a retry rather than a failed command.

## Safety

Nothing ever kills or signals a live broker. A broker shared by a concurrent session is never interrupted, no PID is ever signalled by the GC, and a broker that can no longer serve (dead or wedged app server) removes itself rather than waiting to be found.

## Testing

`tests/broker-lifecycle.test.mjs` (new) covers the reaper GC matrix (dead pid removed; live pid, missing/torn/garbage/zero pid file, and non-`cxc-` dirs all left alone) and pid-based runtime-status liveness; the test env caps broker idle at 30s so suite runs no longer leave broker + fake-codex pairs behind. Functional runs against real brokers verified: idle exit removes process, session dir, and record; `SessionEnd` leaves a live shared broker and its record untouched while interrupting owned turns; a SIGKILLed app-server child takes its broker, record, and session dir down within ~2s. Full suite unchanged against the pre-change baseline.

_Supersedes an earlier owner-tracking revision; see the discussion for the review history (shared-broker ownership, `broker.json` lifecycle, orphaned turns, hook time-budget, and the idle-shutdown disarm paths found by a multi-lens adversarial review)._

### DIFF
--- a/plugins/codex/scripts/app-server-broker.mjs
+++ b/plugins/codex/scripts/app-server-broker.mjs
@@ -69,7 +69,21 @@ async function main() {
   let activeRequestSocket = null;
   let activeStreamSocket = null;
   let activeStreamThreadIds = null;
+  let inFlightRequests = 0;
   const sockets = new Set();
+
+  // Forward a request to the app server while counting it as in-flight, so idle
+  // self-shutdown can never fire while real work is running — even if the calling
+  // client disconnected mid-request (which clears activeRequestSocket).
+  async function forwardAppRequest(method, params) {
+    inFlightRequests += 1;
+    try {
+      return await appClient.request(method, params);
+    } finally {
+      inFlightRequests -= 1;
+      scheduleIdleShutdown();
+    }
+  }
 
   function clearSocketOwnership(socket) {
     if (activeRequestSocket === socket) {
@@ -105,17 +119,56 @@ async function main() {
     }
     await appClient.close().catch(() => {});
     await new Promise((resolve) => server.close(resolve));
-    if (listenTarget.kind === "unix" && fs.existsSync(listenTarget.path)) {
-      fs.unlinkSync(listenTarget.path);
-    }
-    if (pidFile && fs.existsSync(pidFile)) {
-      fs.unlinkSync(pidFile);
+    // Remove the broker's own session directory (socket, pid file, log) so a
+    // clean exit leaves nothing behind for the reaper to GC.
+    const sessionDir = pidFile
+      ? path.dirname(pidFile)
+      : listenTarget.kind === "unix"
+        ? path.dirname(listenTarget.path)
+        : null;
+    if (sessionDir) {
+      try {
+        fs.rmSync(sessionDir, { recursive: true, force: true });
+      } catch {
+        // Ignore an already-removed directory.
+      }
     }
   }
 
   appClient.setNotificationHandler(routeNotification);
 
+  // Idle self-shutdown: a broker is spawned per working directory and is reused
+  // across sessions, so no external actor can safely decide it is done. Instead
+  // the broker exits itself once it has had no connections and no in-flight work
+  // for CODEX_BROKER_IDLE_MS (default 30 min; <= 0 disables). Callers respawn one
+  // on demand, so exiting when idle is safe and stops brokers from accumulating.
+  const idleMs = Number.parseInt(process.env.CODEX_BROKER_IDLE_MS ?? "", 10);
+  const idleTimeoutMs = Number.isFinite(idleMs) ? idleMs : 30 * 60 * 1000;
+  let idleTimer = null;
+  function cancelIdleShutdown() {
+    if (idleTimer) {
+      clearTimeout(idleTimer);
+      idleTimer = null;
+    }
+  }
+  function isIdle() {
+    return sockets.size === 0 && inFlightRequests === 0 && !activeRequestSocket && !activeStreamSocket;
+  }
+  function scheduleIdleShutdown() {
+    cancelIdleShutdown();
+    if (idleTimeoutMs <= 0 || !isIdle()) {
+      return;
+    }
+    idleTimer = setTimeout(() => {
+      if (isIdle()) {
+        shutdown(server).finally(() => process.exit(0));
+      }
+    }, idleTimeoutMs);
+    idleTimer.unref(); // never keep the process alive solely to fire this timer
+  }
+
   const server = net.createServer((socket) => {
+    cancelIdleShutdown();
     sockets.add(socket);
     socket.setEncoding("utf8");
     let buffer = "";
@@ -183,7 +236,7 @@ async function main() {
 
         if (allowInterruptDuringActiveStream) {
           try {
-            const result = await appClient.request(message.method, message.params ?? {});
+            const result = await forwardAppRequest(message.method, message.params ?? {});
             send(socket, { id: message.id, result });
           } catch (error) {
             send(socket, {
@@ -198,7 +251,7 @@ async function main() {
         activeRequestSocket = socket;
 
         try {
-          const result = await appClient.request(message.method, message.params ?? {});
+          const result = await forwardAppRequest(message.method, message.params ?? {});
           send(socket, { id: message.id, result });
           if (isStreaming) {
             activeStreamSocket = socket;
@@ -225,11 +278,13 @@ async function main() {
     socket.on("close", () => {
       sockets.delete(socket);
       clearSocketOwnership(socket);
+      scheduleIdleShutdown();
     });
 
     socket.on("error", () => {
       sockets.delete(socket);
       clearSocketOwnership(socket);
+      scheduleIdleShutdown();
     });
   });
 
@@ -243,7 +298,9 @@ async function main() {
     process.exit(0);
   });
 
-  server.listen(listenTarget.path);
+  server.listen(listenTarget.path, () => {
+    scheduleIdleShutdown(); // exit if nobody ever connects
+  });
 }
 
 main().catch((error) => {
--- a/plugins/codex/scripts/app-server-broker.mjs
+++ b/plugins/codex/scripts/app-server-broker.mjs
@@ -2,6 +2,7 @@
 
 import fs from "node:fs";
 import net from "node:net";
+import os from "node:os";
 import path from "node:path";
 import process from "node:process";
 
@@ -71,7 +72,20 @@ async function main() {
   let activeStreamSocket = null;
   let activeStreamThreadIds = null;
   let inFlightRequests = 0;
+  let shuttingDown = false;
+  let shutdownPromise = null;
+  let serverRef = null;
   const sockets = new Set();
+
+  // Bound each forwarded request so a hung app server cannot pin
+  // inFlightRequests forever, which would permanently disarm idle
+  // self-shutdown and make the broker unkillable by anything but a signal.
+  // Requests resolve at acceptance (streams ride notifications), so the
+  // default is generous; <= 0 disables the bound.
+  const requestTimeoutRaw = (process.env.CODEX_BROKER_REQUEST_TIMEOUT_MS ?? "").trim();
+  const requestTimeoutMs = /^-?\d+$/.test(requestTimeoutRaw)
+    ? Math.min(Number(requestTimeoutRaw), 2 ** 31 - 1)
+    : 10 * 60 * 1000;
 
   // Forward a request to the app server while counting it as in-flight, so idle
   // self-shutdown can never fire while real work is running, even if the calling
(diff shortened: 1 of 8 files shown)


siimvene:
Thanks for the review — both points were spot on. I reworked the fix rather than patch around them.

Instead of an external reaper deciding when to kill a broker (which can't know about a concurrent session reusing it, and can mis-signal a reused PID), the broker now owns its own lifecycle:

- **Idle self-shutdown** (`app-server-broker.mjs`): the broker exits itself after `CODEX_BROKER_IDLE_MS` (default 30 min) with no connections and no in-flight work, and removes its own session directory. In-flight work is tracked with a counter around the app-server request (not `activeRequestSocket`, which a disconnecting client clears), so idle shutdown can't fire mid-request.
- **Reaper is now dead-only** (`broker-lifecycle.mjs`): it only removes the directory of a broker whose PID is already dead. It never inspects or signals a live PID, so it can't interrupt a shared broker, can't signal a reused PID, and can't race a broker that's still starting up.
- **SessionEnd no longer force-kills** the cwd broker (unsafe when shared) — it drops the session's pointer and lets the broker self-exit when idle.

Both original P1s (killing a broker a concurrent session reuses; signalling a possibly-reused PID) are resolved by never killing a live broker at all.

siimvene:
Both points confirmed and fixed in fa1e404. The root of both was the same: broker.json's lifecycle didn't follow the broker-owns-itself model the rest of the fix moved to.

**P1 (SessionEnd deleting the shared record):** `handleSessionEnd` no longer calls `clearBrokerSession`. The record is workspace-wide state shared by every session on the cwd, so a session ending now leaves it (and the broker) entirely alone. Peer sessions keep routing `turn/interrupt` and follow-up commands through the existing broker.

**P2 (stale record after idle shutdown):** three layers, in order of preference:
1. The broker retires its own record at the start of `shutdown()`, guarded on the endpoint still matching, so every clean exit path (idle, `broker/shutdown`, SIGTERM/SIGINT) removes the pointer along with the session dir. Clearing it first means a client probing mid-shutdown finds no record and starts a fresh broker instead of connecting to a dying socket; the guard keeps a record that was already replaced.
2. The `reuseExistingBroker` path in `CodexAppServerClient.connect` validates the recorded endpoint (150ms probe) before using it, so a record left by an uncleanly killed broker behaves like no record: `getCodexAuthStatus` and the interrupt path fall back to a direct app server instead of failing against a dead socket.
3. `getSessionRuntimeStatus` reports "shared" only when the recorded unix socket file still exists (the broker removes it on clean exit; the session-start reaper removes it after an unclean one).

Functionally tested: SessionEnd leaves a live broker and its record untouched; idle shutdown removes process, session dir, and record; a stale record reports direct startup and a reused connect degrades to direct spawn; a newer broker's record survives an older broker's shutdown.

siimvene:
Confirmed and fixed in 4cef183. This was a real regression the redesign introduced: previously the SessionEnd broker kill took any in-flight turn down with it, and removing the kill silently removed that guarantee.

`cleanupSessionJobs` now mirrors `handleCancel`: for each still-running job owned by the ending session it sends `turn/interrupt` with the `threadId`/`turnId` from the per-job file (falling back to the state-entry fields) and then terminates that job's runner immediately, before moving to the next job. Details:

- The interrupt goes through the broker's existing `allowInterruptDuringActiveStream` gate, which admits a second connection interrupting the stream holder's turn.
- It is skipped when no live broker endpoint exists (150ms probe): with no shared app server, the turn already died with the runner's direct one, and probing first avoids spawning a throwaway direct app server just to deliver a meaningless interrupt.
- Per-job interleaving (interrupt, then kill, per job) keeps the window between a runner exiting in response to its interrupt and its PID being signalled to a single await, so a reused PID cannot be killed while other jobs' interrupts are still pending.
- A corrupt or unreadable per-job file falls back to the state entry instead of failing the whole hook, and the final state save re-loads fresh state, since the awaited interrupts widen the old load-to-save window.

Functionally tested: with a live broker, the hook interrupts and exits 0, the job is removed, and the broker plus its record survive; with no broker, the fast path skips the interrupt (~150ms); missing thread/turn ids and corrupt job files degrade gracefully.

siimvene:
Following the round-4 fixes we ran an exhaustive multi-lens adversarial review of the cumulative diff (7 independent reviewers over consumer, concurrency, harness-constraint, portability, failure-injection, and maintainer lenses; every finding attacked by 2 skeptics). It confirmed several defects in this PR beyond what has been raised here, all fixed in this push.

**The broker's idle self-shutdown could be permanently disarmed**, making a wedged broker immortal now that no external path kills brokers:

- A client that disconnected while its streaming request was awaited could be assigned to `activeStreamSocket` after the await, stranding a dead socket that kept `isIdle()` false with no event left to clear it (and busy-rejecting other clients until turn completion). The assignment now checks `socket.destroyed`, and releasing stream ownership on `turn/completed` reschedules the idle timer.
- A codex app-server child that died left the broker listening: endpoint probes passed, the reaper skips live pids, and post-death requests hung forever (a callback-less write to destroyed stdin is a silent no-op), each one pinning `inFlightRequests`. The broker now observes the app-server client's exit and shuts itself down; callers respawn on demand.
- A hung (but alive) app server pinned `inFlightRequests` the same way. Forwarded requests are now bounded by `CODEX_BROKER_REQUEST_TIMEOUT_MS` (default 10 minutes; requests resolve at acceptance, streams ride notifications). Since the race cannot cancel the underlying request, a timeout terminates the whole broker (taking the wedged app server with it) rather than releasing ownership of a possibly still-executing turn to a later client.

**Other fixes in this push:**

- `shutdown()` recursively removed whatever directory contained the caller-supplied `--pid-file`; recursive removal is now restricted to `cxc-` directories directly under the OS temp dir (the plugin's own mkdtemp layout), with individual file unlinks otherwise. All shutdown paths also share one promise now, so a signal landing during an idle shutdown can no longer `process.exit()` mid-cleanup.
- The SessionEnd interrupt budget raced a timer against work that began with two synchronous `spawnSync` availability probes, which block the event loop and cannot be preempted. The hook now skips the availability check (a responding broker already proves the runtime exists), so the raced work is purely async.
- `withAppServer` never fell back to a direct app server when `connect()` itself threw. It now retries direct on any connect-phase failure (nothing ran yet, so a retry is always safe: this covers ENOENT, ECONNREFUSED, ECONNRESET, and initialize-phase closes when racing an idle shutdown). Post-connect retries stay limited to the broker-busy rejection, which the broker raises before any work runs; replaying `fn` after other mid-flight failures could repeat side effects such as an already-created thread.
- The reaper treated an empty or unparseable `broker.pid` as a dead broker and deleted the directory, which could race a broker mid-startup during the pid file's write window. Unparseable now means "leave it alone".
- `CODEX_BROKER_IDLE_MS` parsing: `parseInt` truncated unit suffixes ("30m" became 30ms) and values above 2^31-1 overflow `setTimeout` to ~1ms, both silently inverting "effectively never" configs into instant shutdown. Parsing is now strict-integer with a clamp.
- Connections landing between shutdown start and `server.close()` are destroyed instead of holding the close open; SessionEnd's broker probe now honors the `CODEX_COMPANION_APP_SERVER_ENDPOINT` precedence that `connect()` uses.
- New `tests/broker-lifecycle.test.mjs` covers the reaper's GC matrix (dead pid, live pid, missing/torn/garbage/zero pid file, non-`cxc-` dirs) and pid-based runtime-status liveness; the test env caps broker idle at 30s so suite runs no longer leave broker + fake-codex pairs alive for 30 minutes.

**Verified**: full test suite unchanged against the pre-change baseline; functional runs against real brokers confirm idle exit removes process, dir, and record; SessionEnd leaves live brokers untouched; and a SIGKILLed app-server child now takes its broker, record, and session dir down within ~2s.

**Known, intentionally accepted**: status/auth reads against a stale record now fall back to spawning a direct app server (correct answer at a cost, where pre-diff they failed fast against a dead socket); a broker reused across sessions holds its spawn-time environment until it idles out; sub-second probe-vs-shutdown windows remain and now cost a single direct retry.

siimvene:
Confirmed and fixed in 1cc837a. Name plus location was still an inference about ownership, not a record of it. `spawnBrokerProcess` now passes `--managed-session-dir` after creating the directory via `mkdtemp`, and the broker removes a directory recursively only when that flag was given and the path still matches the plugin's own layout. Any other invocation, including `--pid-file /tmp/cxc-work/broker.pid`, has only the broker's pid file and socket unlinked.

Functionally verified both ways: a manual serve into a user-created `cxc-` directory under the temp dir keeps the directory and its files on shutdown (pid file and socket unlinked); plugin-spawned brokers still remove their session dir on idle exit and on app-server-child death.

siimvene:
Confirmed and fixed in 77db721. `createBrokerSessionDir` now persists a `broker.managed` marker into the directory it mkdtemps, and the reaper deletes only marked directories; `teardownBrokerSession` unlinks the marker so its non-recursive rmdir still succeeds. A manual `--pid-file` location is now safe against every deleter regardless of how it is named or how its broker dies.

Ownership is persisted once at creation and checked by both recursive-removal paths (the broker's own shutdown via `--managed-session-dir`, the reaper via the marker), so there is no remaining code path that removes a directory the plugin did not provably create. Tests extended: an unmarked `cxc-` directory containing a dead pid file and a user file survives the reaper.

siimvene:
Confirmed and fixed. performShutdown now races appClient.close() against a 5s deadline that force-kills the child's process tree (the child's exit settles the dangling close), and every shutdown path carries a 15s process-exit backstop that previously existed only for the request-timeout path. A SIGTERM-immune app server can no longer wedge a shutdown after the record was cleared.

siimvene:
Confirmed and fixed in 8f4eec4, with the mechanics verified empirically this time: against a SIGTERM-trapping child, `terminateProcessTree` reported `delivered=false` (process-group ESRCH, no pid fallback) and the child stayed alive, while a direct `SIGKILL` to the pid took it down. The deadline now sends `SIGKILL` to the child pid on POSIX and keeps `taskkill /T /F` on Windows, where the child is a cmd.exe wrapper that needs the tree kill.

thossullivan:
## Independent Verification 
commit [8f4eec4](https://github.com/openai/codex-plugin-cc/commit/8f4eec49a1966dcbf97e67ec9f2b14b84c7c8bff)
### Test results

- [X] All 96 tests passed.
- [X] `npm run build` passed.

### Reproduction in production

Our production environment exposed the defect through a configured MCP server. A single app-server retained 5 MCP process pairs plus its code-mode host. Terminating that idle app-server reaped every descendant.

Tested standalone, the MCP server handled both stdin EOF and `SIGTERM` correctly.

### Conclusion

This is a generic Codex Companion lifecycle defect, not an MCP-specific one. PR #660 places idle shutdown and child supervision in the correct layer, and the fix requires no MCP-specific dependency or configuration change.

PR #667, opened 2026-08-22
Recover Codex jobs across workspace scopes

villilms:
## Summary
- scope companion commands to `CLAUDE_PROJECT_DIR` when `--cwd` is omitted
- persist background jobs before spawning detached workers
- recover explicit job IDs across workspace-scoped state directories
- add regression coverage for cross-workspace status/result and project-directory scoping

## Problem
A Claude rescue subagent can launch Codex from a parent/home directory while the target repository is only available through `CLAUDE_PROJECT_DIR`. The background job then lands in the wrong workspace registry. Later `status <job-id>` and `result <job-id>` from the target repository report that the job disappeared even though Codex completed it. The old enqueue order also spawned the worker before its job file existed.

## Verification
- targeted regression suite: 6 passed
- live Windows smoke test against the installed plugin: background job registered under the target project, recovered from another workspace, completed with `CHANNEL_OK`
- full upstream suite: 87 passed; 6 existing Windows-specific failures (Unix path assertion, symlink permissions, and transfer fixtures tied to the real Claude home)

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,7 +135,10 @@ Use it when you want Codex to:
 - take a faster or cheaper pass with a smaller model
 
 > [!NOTE]
-> Depending on the task and the model you choose these tasks might take a long time and it's generally recommended to force the task to be in the background or move the agent to the background.
+> Depending on the task and model, a rescue can take a while. `--background`
+> detaches the Claude Code subagent while its Codex command stays attached to
+> that subagent, so the final result is delivered back automatically when the
+> subagent completes.
 
 It supports `--background`, `--wait`, `--resume`, and `--fresh`. If you omit `--resume` and `--fresh`, the plugin can offer to continue the latest rescue thread for this repo.
 
--- a/plugins/codex/agents/codex-rescue.md
+++ b/plugins/codex/agents/codex-rescue.md
@@ -20,8 +20,12 @@ Selection guidance:
 Forwarding rules:
 
 - Use exactly one `Bash` call to invoke `node "${CLAUDE_PLUGIN_ROOT}/scripts/codex-companion.mjs" task ...`.
-- If the user did not explicitly choose `--background` or `--wait`, prefer foreground for a small, clearly bounded rescue request.
-- If the user did not explicitly choose `--background` or `--wait` and the task looks complicated, open-ended, multi-step, or likely to keep Codex running for a long time, prefer background execution.
+- Always invoke the companion `task` command in the foreground. Never add
+  `--background` to the companion command, even for a long or complicated task.
+- `--background` controls whether Claude Code runs this subagent in the
+  background; it does not control the companion process. Keeping the companion
+  call foreground-bound lets the subagent completion notification carry the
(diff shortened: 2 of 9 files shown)


villilms:
Addressed the cancellation race in e06cd2c:

- added a startup gate so the detached worker cannot enter runTrackedJob before its PID is persisted;
- the worker checks for a stored cancelled state after the gate opens;
- added a regression test that cancels immediately after enqueue and verifies the job cannot restart;
- targeted suite: 7/7 passed.

@codex review

villilms:
Addressed the cross-workspace cancellation finding in 93beb53:

- explicit cancel now falls back to the same cross-workspace lookup used by status/result;
- fallback is restricted to active queued/running jobs;
- added a deterministic regression fixture proving cancel from the wrong workspace updates the owning workspace record to cancelled.

Targeted cross-scope cancellation test passed.

@codex review

villilms:
Addressed the owning-broker finding in 7e107ec:

- handleCancel now calls interruptAppServerTurn with the resolved owning workspaceRoot, not the invocation cwd;
- added a two-workspace regression test that invokes cancel from the wrong scope and verifies the owning broker receives the expected threadId/turnId;
- targeted broker-routing test passed.

@codex review

villilms:
Addressed the startup cancellation race in 87db4e2.

The bootstrap job file is now kept out of the shared state index until the detached worker PID is known. The first publicly discoverable `queued` record therefore always has a usable PID, and the worker gate is released only after both the job file and shared index contain that launch record. There is no longer a visible `queued + pid:null` record for another process to cancel and have overwritten.

Validation:
- `node --test --test-name-pattern="task --background enqueues|immediately cancelled background task" tests/runtime.test.mjs` — 2/2 passed
- `git diff --check` — clean apart from the repository's Windows LF/CRLF warning
- Full `npm test` was attempted; the existing Windows-only Unix-socket and broken-symlink cases failed, then the suite stalled and was stopped. Neither failure touches this change.

@codex review

villilms:
The review requested for current head 87db4e2 is still pending after the initial monitoring window. Re-requesting review for that exact commit. @codex review

villilms:
Fixed rescue result delivery in 18cdf7a.

Root cause: the rescue subagent could infer companion `task --background` for long work. That detached the Codex worker from the subagent, so Claude received only the queued job ID and no completion result; the rescue contract also forbids polling `status`/`result`.

The contract now separates the two layers:
- Claude Code may background the rescue subagent when explicitly requested;
- the subagent's single companion `task` invocation must always remain foreground-bound;
- `--background`/`--wait` are stripped before the companion call;
- final Codex stdout therefore returns through the subagent completion notification.

Validation:
- `node --test tests/commands.test.mjs` — 8/8 passed
- foreground completion runtime tests — 2/2 passed
- fresh installed Claude session end-to-end smoke test returned exactly `HANDOFF_OK`
- local plugin updated and enabled as `1.0.6+local-20260823-072838`

@codex review

Issue #643, opened 2026-08-13
Review jobs on Windows leave junk files in the reviewed repo (pwsh treats ">" in quoted code as a redirect)


Kiborgik:
After review jobs on Windows I keep finding empty files in the repo root with names like `s.id`, `r.name`, `fraction)`, `tsx`, `process.exit(signal`, `7`, `71`.

Each name is a token that comes right after a `>` or `=>` somewhere in the code the reviewer was reading.

What happens: the reviewer runs its shell commands through PowerShell, like

    "pwsh.exe" -Command '$p='"'src/core/actions/index.ts'; "'$l...

The command strings embed pieces of source code, and the quoting often comes out broken (you can see the mixed quote segments in the job log). When a `>` from the code lands outside the quotes, PowerShell reads it as a redirect and creates an empty file named after the next token. So reviewing a diff with an arrow function like `(s) => s.id` can leave a file called `s.id` in the repo.

Setup: Windows 11, PowerShell 7, plugin 1.0.6, Claude Code.

Repro: on Windows, review any diff that contains arrow functions. Then look at the repo root for new empty files.

Two things would fix it:

1. Run review jobs in a read-only sandbox. A review never needs to write into the repo, but right now the reviewer's commands run with no write protection on Windows. This part probably needs a change in codex itself, not just the plugin. Related: #57 and #280 show the Windows sandbox used to block PowerShell commands in review mode. Whatever fixed that seems to have dropped the write protection with it.

2. Stop defaulting to `shell: true` on Windows in the plugin's own spawns (`scripts/lib/process.mjs` line 12, `scripts/lib/app-server.mjs` line 194). Node prints DEP0190 on every run because of it. `scripts/lib/git.mjs` already uses `shell: false` on purpose, with a comment saying why. Note that plain `shell: false` breaks `.cmd` resolution (#287), so the binary path has to be resolved first. Related: #452, #417.

Kiborgik:
## Where to fix it

**1. `scripts/lib/process.mjs`, line 12** — every command runs through a shell on Windows:

```js
shell: options.shell ?? (process.platform === "win32" ? (process.env.SHELL || true) : false),
```

Problems: Node prints DEP0190 on every run (args are concatenated, not escaped), and if `SHELL` happens to be set (Git Bash environments), the command runs through bash instead of cmd — that's what breaks in #336.

[...]

PR #669, opened 2026-08-22
fix: resolve executables instead of shell-wrapping SHELL on Windows spawns

### LINKED ISSUE #287: Windows: spawn("codex") in app-server.mjs throws ENOENT (Node does not try PATHEXT for .cmd shims)
## Environment
- OS: Windows 11
- Node: v22.x
- codex: 0.128.0 (npm global install)
- Plugin: openai-codex 1.0.1

## Symptom
On Windows, invoking `codex:rescue` (or any path that triggers `CodexAppServerClient.connect`) fails with:

```
Error: spawn codex ENOENT
```

`/codex:setup` reports `ready: true` because detection takes a different code path that already uses `shell: true`, masking the issue.

## Root cause
`scripts/lib/app-server.mjs:188` calls:

```js
this.proc = spawn("codex", ["app-server"], {
  cwd: this.cwd,
  env: this.options.env,
  stdio: ["pipe", "pipe", "pipe"]
});
```

On Windows, `child_process.spawn` does **not** consult `PATHEXT`, so it cannot resolve a `.cmd` shim. The npm global install of `@openai/codex` only ships:

```
codex       (sh shim, no extension)
codex.cmd
codex.ps1
```

There is no `codex.exe`, so spawn falls through with ENOENT.

By contrast, `scripts/lib/process.mjs:5` already uses `shell: process.platform === "win32"` for the same reason — that is why `binaryAvailable("codex", ...)` works during `/codex:setup`.

## Reproduction
On Windows with codex installed via `npm i -g @openai/codex`:

```js
// fails with ENOENT
require('child_process').spawn('codex', ['--version']).on('error', console.error);

// works
require('child_process').spawn('codex', ['--version'], { shell: true }).on('error', console.error);
```

## Diagnostic note
Stderr from a failed task-worker often shows a `DEP0190` deprecation warning **and** the ENOENT side by side. They come from *different* spawn points:

- `DEP0190` originates from `process.mjs` (uses `shell: true` + args array — Node 22+ warns on this combination, but the spawn itself succeeds).
- `ENOENT` originates from `app-server.mjs:188` (no `shell: true`, fails to resolve the `.cmd` shim).

Easy to misread as "shell:true is already set but spawn still fails."

## Suggested fix
Two options:

1. **Match the existing pattern in `process.mjs`** — add `shell: process.platform === "win32"` to the spawn options at `app-server.mjs:188`. One-line change. Caveat: on Windows this wraps the child in `cmd.exe`, so `this.proc.kill()` may leave the actual codex process as an orphan.

2. **Resolve the executable's full path before spawning** — try `codex`, `codex.cmd`, `codex.exe` against `PATHEXT` (or use a small helper / `cross-spawn`) and spawn the resolved path directly. No shell wrapper, kill targets the codex process directly. Cleaner long-term fix.

Option 2 is preferable; option 1 is a one-line patch that unblocks Windows users immediately.

mittalpk:
Fixes #287.

`scripts/lib/process.mjs`'s `runCommand` and `scripts/lib/app-server.mjs`'s `codex app-server` launch both defaulted to `shell: process.env.SHELL || true` on Windows. This has two problems:

- It's how #287 happens: `spawn`/`spawnSync` never consult `PATHEXT`, so a bare `codex` (which on Windows only ships as a `.cmd`/`.ps1` shim, no `.exe`) fails with `ENOENT` unless something resolves it first.
- Handing `process.env.SHELL` to `shell:` means Node passes the whole command line to whatever that variable points at, using Node's own quoting/escaping (tuned for `cmd.exe`). When `SHELL` is set to a PowerShell path, that escaping doesn't match PowerShell's own quoting rules.

This adds `resolveExecutablePath()`, which walks `PATH`/`PATHEXT` to find the concrete executable file on Windows (mirroring how a shell would resolve it), then spawns that resolved path with `shell: false`. Node still safely wraps a resolved `.cmd`/`.bat` target through `cmd.exe` internally when needed (hardened by the CVE-2024-27980 fix) — but this plugin's own spawns never hand an arbitrary shell string a raw command line to reinterpret again. This also removes the `DEP0190` warning these spawns triggered on every run.

**Re: #643** — this removes the same `shell: process.env.SHELL || true` pattern the issue flags in this plugin's own code (fix option 2 in the issue), and fixes the concrete `DEP0190`/`ENOENT` class of bug it's part of. It does **not** claim to fix the reported junk-file symptom itself: I traced every spawn in this plugin and none of them pass review-job source content as arguments, so the junk files most likely originate inside `codex app-server`'s own process — the compiled binary this plugin launches, not something this diff touches. That likely needs fix option 1 from the issue (sandboxed/read-only review execution in codex itself). Flagging this rather than claiming more than I can verify without a Windows machine.

**Testing:** 9 new unit tests for `resolveExecutablePath` (PATH/PATHEXT ordering, known-extension passthrough, custom `PATHEXT`, no-match fallback, non-Windows no-op) plus an end-to-end `runCommand` smoke test. Full suite: 100 passed, 0 regressions. `tsc -p tsconfig.app-server.json` clean.

### DIFF
--- a/plugins/codex/scripts/lib/app-server.mjs
+++ b/plugins/codex/scripts/lib/app-server.mjs
@@ -188,7 +188,7 @@ class SpawnedCodexAppServerClient extends AppServerClientBase {
 
   async initialize() {
     const env = this.options.env ?? process.env;
-    const invocation = resolveSpawnInvocation("codex", ["app-server"], { env });
+    const invocation = resolveSpawnInvocation("codex", ["app-server"], { env, cwd: this.cwd });
     this.proc = spawn(invocation.command, invocation.args, {
       cwd: this.cwd,
       env,
--- a/plugins/codex/scripts/lib/app-server.mjs
+++ b/plugins/codex/scripts/lib/app-server.mjs
@@ -14,7 +14,7 @@ import { spawn } from "node:child_process";
 import readline from "node:readline";
 import { parseBrokerEndpoint } from "./broker-endpoint.mjs";
 import { ensureBrokerSession, loadBrokerSession } from "./broker-lifecycle.mjs";
-import { resolveExecutablePath, terminateProcessTree } from "./process.mjs";
+import { resolveSpawnInvocation, terminateProcessTree } from "./process.mjs";
 
 const PLUGIN_MANIFEST_URL = new URL("../../.claude-plugin/plugin.json", import.meta.url);
 const PLUGIN_MANIFEST = JSON.parse(fs.readFileSync(PLUGIN_MANIFEST_URL, "utf8"));
@@ -187,11 +187,14 @@ class SpawnedCodexAppServerClient extends AppServerClientBase {
   }
 
   async initialize() {
-    this.proc = spawn(resolveExecutablePath("codex"), ["app-server"], {
+    const env = this.options.env ?? process.env;
+    const invocation = resolveSpawnInvocation("codex", ["app-server"], { env });
+    this.proc = spawn(invocation.command, invocation.args, {
       cwd: this.cwd,
-      env: this.options.env ?? process.env,
+      env,
       stdio: ["pipe", "pipe", "pipe"],
       shell: false,
+      windowsVerbatimArguments: invocation.windowsVerbatimArguments,
       windowsHide: true
     });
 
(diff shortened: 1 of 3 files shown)


mittalpk:
Both correct, thanks. The P1 was based on a wrong assumption on my part about Node auto-wrapping .cmd/.bat through cmd.exe even with shell:false — it doesn't, and the previous version would have broken the primary codex app-server launch outright. Fixed by explicitly spawning cmd.exe (never SHELL) for anything that isn't .exe/.com, using cross-spawn's escaping approach so the quoting is handled correctly. Also fixed the env-vs-process.env mismatch on PATH/PATHEXT/comspec resolution.

mittalpk:
Correct, thanks. Confirmed against CreateProcess's own documented search sequence (current directory comes before PATH) before fixing this one. Threaded options.cwd through resolution, searched before PATH, and relative PATH entries now resolve against it too.

mittalpk:
Both correct, thanks. Confirmed NoDefaultCurrentDirectoryInExePath's presence-not-value semantics against NeedCurrentDirectoryForExePath's docs before fixing — genuinely important here since cwd is an untrusted repo checkout. Skipped the cwd prepend when it's set (relative PATH entries still resolve against cwd, that's a separate mechanism). Also added a case-insensitive env lookup for PATH/PATHEXT/comspec so casing like your example (path/pathext) is honored the same way spawn() itself honors it.

mittalpk:
Correct, thanks. This was exactly the gap -- the check only consulted options.env like PATH/PATHEXT do, but the real preflight (codex.mjs) never passes env at all, so it was silently invisible on the one call path that matters. Falls back to process.env now, matching the adjacent lookups. Also worth noting for anyone debugging this locally: this variable can leak in from a WSL host's real Windows environment via interop, which is what surfaced a test-isolation gap while fixing this.

mittalpk:
Checking in on this one, happy to make any further changes if needed.

PR #711, opened 2026-09-02
fix(task): publish jobs before worker spawn

### LINKED ISSUE #620: task --background: detached worker is spawned before its job file is written, and a fast-failing worker leaves a permanently queued job that was reported as started
**Version:** plugin 1.0.6 (codex-companion.mjs), observed on Windows 11, Node 24, codex-cli 0.146.0-alpha.3.1.

**Summary.** `enqueueBackgroundTask` spawns the detached task worker before writing the job record the worker requires at startup. If the worker wins the race (or dies at startup for any reason), the parent still writes a `queued` record and reports the task as started. Nothing ever moves that record to a terminal state, and `status` reports it as active indefinitely.

**Where (scripts/codex-companion.mjs, v1.0.6):**

- `enqueueBackgroundTask` calls `spawnDetachedTaskWorker(cwd, job.id)` at line 688, then writes the job file with `writeJobFile(...)` at line 697 and upserts the index at line 698.
- The worker (`handleTaskWorker`, line 838) immediately reads that job file and throws if it is absent.
- The worker is spawned with `stdio: "ignore"` and `detached: true` (lines 673-679), so its startup failure is invisible to the parent and to the user.
- `renderQueuedTaskLaunch` (line 556) then reports the task as started in the background.
- `status` classifies jobs from the stored `status` field only (`buildStatusSnapshot`, lib/job-control.mjs line 213 onward); there is no PID liveness check, so the stuck `queued` record looks active forever.

**Impact.** A user who dispatches a background task gets a success message and a job id, but no work is running and no error ever surfaces. The failure mode is indistinguishable from a slow queue until the user inspects the log file (which contains only the "Queued for background execution." line).

**Suggested fix.** Write the job file and upsert the index BEFORE spawning the worker (swap lines 688-698 so `writeJobFile`/`upsertJob` precede `spawnDetachedTaskWorker`). The worker's read-at-startup then cannot race the write. Optionally: have `status` reconcile `queued`/`running` records whose `pid` is no longer alive.

**Related hardening idea (separate, smaller).** A cancel failure on Windows (`taskkill` returning an unrecognized nonzero, e.g. "operation not supported" under a restricted shell) throws in `terminateProcessTree` (lib/process.mjs line 97) before any cancellation bookkeeping, which also leaves the record active with no recovery path; a liveness reconciliation in `status` would cover both cases.

ALV0612:
## Summary

- publish the complete pid-less queued job record before starting its detached worker
- wait for the child `spawn` acknowledgement before reporting queued success
- let the worker exclusively own the running transition and PID registration, preventing parent/worker state regression
- persist a terminal failed job and return non-zero when spawn fails or yields no PID
- include initial `runTrackedJob` persistence inside its failure boundary

## Root cause

`enqueueBackgroundTask()` called `spawnDetachedTaskWorker()` before writing the job record consumed by `task-worker`. On Windows, parent-side `spawn()` was measured taking 716-878 ms while the child reached `readStoredJob()` in about 327 ms, so a normal single dispatch could lose the race, exit with “No stored job found,” and leave a permanent queued record.

This change removes the race by construction rather than adding sleeps or retries. The parent first publishes all worker input with `pid: null`; after process creation is acknowledged, the worker performs the existing queued → running transition and records `process.pid` itself.

## Testing

- `node --test tests/background-dispatch.test.mjs tests/tracked-jobs.test.mjs` — 5/5 pass
- existing background task integration test — pass
- `npm test` — 96/96 pass
- `npm run build` — pass
- `npm run check-version` — pass
- `git diff --check` — pass

Regression coverage includes:

- complete request exists synchronously before spawn
- worker-authored running state cannot be regressed by the parent
- spawn error converges job file and state index to failed
- spawn acknowledgement without a finite PID fails closed
- initial running-state persistence failure prevents runner execution
- normal detached task remains resolvable through status/result

Fixes #620

### DIFF
--- a/plugins/codex/scripts/codex-companion.mjs
+++ b/plugins/codex/scripts/codex-companion.mjs
@@ -22,6 +22,7 @@ import {
     runAppServerTurn
   } from "./lib/codex.mjs";
 import { resolveClaudeSessionPath } from "./lib/claude-session-transfer.mjs";
+import { dispatchBackgroundJob } from "./lib/background-dispatch.mjs";
 import { readStdinIfPiped } from "./lib/fs.mjs";
 import { collectReviewContext, ensureGitRepository, resolveReviewTarget } from "./lib/git.mjs";
 import { binaryAvailable, terminateProcessTree } from "./lib/process.mjs";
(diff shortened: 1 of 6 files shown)


ALV0612:
Closing this PR because a deeper backlog audit found that open PR #667 already implements the same publish-before-spawn ordering for background jobs. I missed it initially because #667 does not reference issue #620 in its title/body or timeline. Avoiding duplicate maintainer work; the independent tests here remain available if useful.

PR #719, opened 2026-09-03
feat: add `task --thread <id>` to resume a specific Codex thread

y-cruce:
## Summary

`task --resume-last` can only continue the newest finished task thread of the current Claude session in the repo. When a caller drives several Codex threads in one session (for example one thread per problem, or a review that ran as a task in between), the older threads become unreachable even though the app-server can resume them by id.

This adds `task --thread <id>` to resume a specific thread through the same path `--resume-last` uses.

## Changes

- `plugins/codex/scripts/codex-companion.mjs`: parse `--thread`, validate it (non-empty, not flag-like, mutually exclusive with `--resume`/`--resume-last`/`--fresh`), pass `resumeThreadId` through `buildTaskRequest`/`executeTaskRun` in both foreground and `--background` modes. The thread id is not pre-written into the job record; it is stored by the existing `Thread ready` progress handling after a successful resume, so a failed resume of an unknown id does not pollute the next `--resume-last`.
- `tests/runtime.test.mjs`: resume by id (foreground and background), flag not leaked into the prompt, conflicting and invalid values rejected, failed resume does not pre-write the id.
- `tests/commands.test.mjs`, `README.md`, `plugins/codex/commands/rescue.md`, `plugins/codex/agents/codex-rescue.md`, `plugins/codex/skills/codex-cli-runtime/SKILL.md`: document the flag and its routing.

No changes under `lib/`; `--resume-last`, `--resume`, and `--fresh` behave as before. Version not bumped.

## Verification

- `npm test`: the new tests pass; no new failures relative to the clean tree on the same machine (the remaining failures there are environment-related `setup`/`status`/`result`/`resolveStateDir` cases that fail identically without this change).
- Manual: with a shared app-server running and a newer thread being the `task-resume-candidate`, `task --thread <older-id> "..."` resumed the older thread (`Thread ready (<older-id>)`) and Codex answered from that thread's context.

## Notes

`--thread` intentionally bypasses the current-session filter used by `--resume-last`. It does not check repository ownership of the thread; the resume request carries the current `cwd`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,6 +74,25 @@ One simple first run is:
 
 ## Usage
 
+### Live task input
+
+Use the job ID from `/codex:status` to control an already running task:
+
+```text
+/codex:message <job-id> Use the latest plan from the plan center, not the stored chargeId.
+/codex:message <job-id> --interrupt Stop this approach and inspect the new requirement first.
+/codex:status <job-id> --wait
+/codex:answer <job-id> --request-id <id> --answers-file /absolute/path/answers.json
+```
+
+`message` uses native `turn/steer`: inputs stay ordered and are consumed before a later model request, not necessarily before tools already issued by the model. The acknowledgement means accepted, not executed. `--interrupt` cancels the current turn and starts a new turn in the same job and thread with the new input and the original write permission. Existing changes remain; its report includes observed file changes and the workspace's Git status, including pre-existing edits.
+
+`status <job-id>` shows pending messages, questions, and interruption state. Messages leave the pending list when their user-message event is observed; this confirms entry into thread history, not model execution. `status --wait` returns early for questions. Answer files contain an answers map such as `{"source":{"answers":["Use the latest plan."]}}`. Answers must match the pending request and question IDs. Questions time out after 10 minutes and interrupt the turn; the plugin does not fabricate an answer or auto-approve permissions. After answering, wait on the same job again and collect `/codex:result <job-id>`.
+
+Tasks require the shared broker; live commands never start a second runtime or silently create another thread. The broker enables `default_mode_request_user_input` for its Codex process without editing global configuration. Existing sessions must restart to load an updated broker. `task --thread <id> --write` applies workspace-write on `turn/start`; omitting `--write` explicitly restores read-only. Native `thread/queue/*` (follow-up turns after completion) is separate and is not exposed by these mid-turn controls.
+
+Run `npm test` and `npm run build` for local checks. The optional `CODEX_REAL_APP_SERVER_TEST=1 node --test tests/real-app-server.test.mjs` uses an installed Codex CLI with isolated temporary configuration and a local mock model, including actual file writing and syntax validation. It does not contact a real model service.
+
 ### `/codex:review`
 
 Runs a normal Codex review on your current work. It gives you the same quality of code review as running `/review` inside Codex directly.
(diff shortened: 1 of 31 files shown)


sylvesterkaczmarek (commented):
`--thread` accepts any known thread ID without checking the thread’s repository. That can resume context from repo A while executing with repo B’s current `cwd`, which risks cross-repo context leakage and wrong-repo edits. Please verify the saved thread cwd/repository before resuming, with an explicit override only if cross-repo resume is intended.

y-cruce:
Good catch, thanks. `--resume-last` is implicitly scoped to the current workspace through the job records, and `--thread` skipped that.

Added in the latest commit: before resuming, the thread id must appear in the current workspace's tracked task jobs (any Claude session). Otherwise the run fails before any resume request is sent, with an error pointing to `--allow-other-repo`, which is the explicit override for an intentional cross-repo resume. The check runs in `executeTaskRun`, so foreground and `--background` behave the same, and a rejected run leaves the job failed without a `threadId`.

Threads created outside the plugin (plain `codex` CLI) are not tracked and therefore also need the override; the error text and the docs say so.

Tests cover: tracked thread from another Claude session resumes; untracked thread rejected in foreground and background; `--allow-other-repo` resumes an untracked thread; `--allow-other-repo` with an unknown id still fails at Codex without pre-writing the id.

sylvesterkaczmarek (commented):
Rechecked the current `9b63301` head against the repository-scope guard. The original cross-repo concern is improved, but two ownership edges remain:

1. A foreign thread resumed once with `--allow-other-repo` is then written into this repository's bounded job history. A later `task --thread <id>` sees that local job record and accepts the thread without the override, even though its original provenance is still another repository. The override should not permanently convert foreign provenance into local ownership.
2. `handleTask()` creates/persists the new job before `executeTaskRun()` calls `requireTrackedThreadForWorkspace()`. If the requested thread exists only in the oldest record at the 50-job cap, creating the new job can prune the very evidence the validation needs and incorrectly reject a thread that was tracked at command start.

Could thread ownership/provenance be stored separately from bounded job history, or validated from a pre-insertion snapshot, with foreign-thread provenance retained across overridden resumes?

sylvesterkaczmarek (commented):
Rechecked current head `251ef06`. The two thread-ownership issues from my previous review are still present:

1. `requireTrackedThreadForWorkspace()` still treats any local job record containing the thread ID as repository ownership. A foreign thread resumed once with `--allow-other-repo` is persisted into this repositorys

sylvesterkaczmarek (commented):
Rechecked current head `251ef06`. The two thread-ownership issues from my previous review are still present:

1. `requireTrackedThreadForWorkspace()` still treats any local job record containing the thread ID as repository ownership. A foreign thread resumed once with `--allow-other-repo` is persisted into this repository's job history, so a later `task --thread <id>` can resume it without the override. The override still effectively converts foreign provenance into local ownership.
2. Validation still happens after job-state mutation. Foreground `runTrackedJob()` writes/upserts the new running record before `executeTaskRun()` calls `requireTrackedThreadForWorkspace()`, and background `enqueueBackgroundTask()` does the same before the worker validates. Because `upsertJob()` prunes the bounded history, the 50-job edge can still discard the oldest record that proves the requested thread was tracked at command start.

The new broker/liveness changes do not address either ownership boundary. I would keep thread provenance separate from bounded job history, or validate from an immutable pre-insertion snapshot and retain foreign provenance explicitly across overridden resumes.

PR #722, opened 2026-09-04
docs(rescue): document writable sandbox scope

### LINKED ISSUE #673: docs(codex-rescue): document the sandbox writable-root constraint (writes outside the repo are silently read as delegation failure)
### Summary

When a Claude Code session delegates to the `codex-rescue` agent and the task specifies an output path **outside the repository** (e.g. `~/.local/bin/foo`), Codex refuses to write:

```
writing outside of the project; rejected by user approval settings
```

Codex reports this clearly and even suggests the remedy. The problem is that **the constraint is not documented anywhere in the plugin** — `agents/codex-rescue.md`, the skills, and the README contain no mention of `writable` / sandbox roots.

### Why this is costly

The delegating agent typically checks whether the artifact files exist, sees nothing, and concludes "the delegation failed / the result could not be retrieved." It then redoes the work itself. The actual cause — an impossible spec — is only visible in the Codex CLI output, which the parent may never read. This is silent, repeats every session, and wastes both time and tokens.

### Reproduction

1. From Claude Code, delegate to `codex-rescue` with a prompt like: "create `~/.local/bin/mytool` and a test script next to it, then run the tests."
2. Codex reports the rejection above; no files are created.
3. Nothing in the plugin documentation would have warned the caller in advance.

### Suggested fix

Add a short note to `agents/codex-rescue.md` (and ideally the runtime/prompting skills), for example:

> **Sandbox scope**: Codex can only write inside the repository and temp directories. Tasks that must produce files elsewhere (`~/.local/bin`, `~/.config`, system paths) should write into the repo and let the caller install them.

Two smaller points that would help the same failure mode:

- Note that runs exceeding the foreground timeout move to the background, and that `codex-rescue` cannot retrieve its own background result — the caller must read the output file.
- Recommend that the caller read the Codex output rather than inferring success/failure from artifact presence.

### Environment

- codex plugin `1.0.6`
- Claude Code on macOS (arm64)

rksharma-owg:
## Summary

- document that write-capable rescue runs are limited to the current repository and temporary directories
- recommend staging artifacts in the repository when their final destination is a home, configuration, or system path
- keep the user-facing README, rescue agent, and runtime guidance aligned

## Testing

- `npm test` (91 tests passed)
- `npm run build`
- `npm run check-version`
- `git diff --check`

Fixes #673

### DIFF
--- a/README.md
+++ b/README.md
@@ -161,6 +161,7 @@ Ask Codex to redesign the database connection to be more resilient.
 - if you do not pass `--model` or `--effort`, Codex chooses its own defaults.
 - if you say `spark`, the plugin maps that to `gpt-5.3-codex-spark`
 - follow-up rescue requests can continue the latest Codex task in the repo
+- write-capable rescue runs can write only inside the current repository and temporary directories. For destinations such as `~/.local/bin` or `~/.config`, have Codex create the artifact in the repository, then install or copy it separately.
 
 ### `/codex:transfer`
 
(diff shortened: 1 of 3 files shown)


rksharma-owg:
@codex review

Issue #708, opened 2026-09-02
Windows: `cancel` silently fails — Git Bash mangles `taskkill /PID /T /F` flags, leaving jobs stuck as `running`


tylerurish54-svg:
## Summary

On Windows with Git Bash as the calling shell, `codex-companion cancel` never actually kills the job. It shells out `taskkill /PID <n> /T /F`, but MSYS (Git Bash) path-converts those flags before `taskkill` receives them. The command still exits cleanly, so the failure is **silent** — the job stays marked `running` in `state.json` indefinitely.

## Reproduction

In Git Bash on Windows, any leading-slash argument handed to a native Windows executable is rewritten:

```console
$ python -c "import sys; print(sys.argv[1:])" /PID 1234 /T /F
['C:/Program Files/Git/PID', '1234', 'T:/', 'F:/']
```

So the cancel path effectively invokes:

```
taskkill C:/Program Files/Git/PID 1234 T:/ F:/
```

Every flag is corrupted — `/PID` becomes a path inside the Git installation, and the single letters become drive roots.

Confirmed by the MSYS escape hatch, which passes them through intact:

```console
$ MSYS_NO_PATHCONV=1 python -c "import sys; print(sys.argv[1:])" /PID 1234 /T /F
['/PID', '1234', '/T', '/F']
```

## Impact

- Background `/codex:rescue` jobs that wedge **cannot be cancelled**.
- `codex-companion status` keeps reporting them as `running` (phase stuck at `starting`) indefinitely, and the stale record carries into later runs, so a relaunch inherits a phantom job.
- Observed twice in separate sessions, ~30 minutes lost each time, with `cancel` appearing to hang or no-op. Recovery required killing the PID manually from PowerShell and hand-editing `state.json`.

**Environment:** Windows 11, codex-cli 0.151.0, codex plugin 1.0.6, Git Bash (MSYS) as the calling shell.

## Suggested fix

Avoid handing MSYS-convertible flags through a shell. Options, roughly in order of robustness:

1. **Spawn `taskkill` with an argument array and no shell** — e.g. `child_process.spawn('taskkill', ['/PID', String(pid), '/T', '/F'], { shell: false })`. The CreateProcess path is not subject to MSYS conversion. (Verified in Python: an arg-list `subprocess` call passes the identical flags through intact even when launched from Git Bash.)
2. Set `MSYS_NO_PATHCONV=1` (or `MSYS2_ARG_CONV_EXCL=*`) in the child environment.
[...]

Issue #525, opened 2026-07-18
Windows: taskkill /PID is mangled by MSYS path conversion when SHELL is set (Git Bash under Claude Code) — cancel and SessionEnd cleanup never kill the process tree


danscMax:
### Summary

On Windows, `terminateProcessTree()` invokes `taskkill /PID <pid> /T /F` through a shell because `runCommand` defaults to `shell: process.env.SHELL || true` on win32 (`plugins/codex/scripts/lib/process.mjs`). When `SHELL` points to Git Bash — which is exactly the environment Claude Code provides on Windows (`SHELL=/bin/bash.exe`) — MSYS argument conversion rewrites the `/PID` switch into a filesystem path, so `taskkill` fails and the process tree is never killed.

### Reproduction

Windows 11, plugin 1.0.6, Node 24.16.0, `SHELL=/bin/bash.exe` (default under Claude Code's bash tool):

```js
const { spawnSync } = require("child_process");
const r = spawnSync("taskkill", ["/PID", "999999", "/T", "/F"],
  { shell: process.env.SHELL || true, encoding: "utf8", windowsHide: true });
console.log(r.status, r.stderr || r.stdout);
```

Output (status 1, cp866 mojibake from the Russian locale):

```
Ошибка: неправильный параметр или аргумент - 'C:/Program Files/Git/PID'.
```

MSYS has translated `/PID` → `C:/Program Files/Git/PID`. With `shell: false` the same call works correctly.

### Impact

- `/codex:cancel` marks a background job `cancelled`, but the detached task-worker tree survives (the soft `turn/interrupt` path still works when a `turnId` is already known, but the hard-kill never lands).
- The `SessionEnd` hook's job cleanup silently fails — `cleanupSessionJobs` swallows the error, leaking worker processes on every session end.
- `SpawnedCodexAppServerClient.close()` on Windows relies on the same `terminateProcessTree`, so the shell-wrapped `codex app-server` grandchild can be leaked too.
- Independently of MSYS: `looksLikeMissingProcessMessage()` matches English-only `taskkill` output (`not found|no running instance|...`). On non-English Windows locales `taskkill` prints localized (OEM-codepage) text, so the "process already gone" detection never matches and dead-process cases surface as thrown errors.

This also shows up when running the repo's own test suite on Windows: the cancel/session-end/status tests fail with the mangled-`/PID` assertion above (CI currently runs `ubuntu-latest` only, so it goes unnoticed).

### Suggested fix

[...]

PR #725, opened 2026-09-04
fix: avoid shell:true deprecation on Windows

### LINKED ISSUE #717: DEP0190 on Node 24: every companion command warns on stderr before its JSON
## What happens

On Node 24, any `codex-companion.mjs` invocation prints a deprecation warning to stderr before its output:

```
$ node ".../openai-codex/codex/1.0.6/scripts/codex-companion.mjs" setup --json
(node:70392) [DEP0190] DeprecationWarning: Passing args to a child process with shell option true can lead to security vulnerabilities, as the arguments are not escaped, only concatenated.
(Use `node --trace-deprecation ...` to show where the warning was created)
{
  "ready": true,
  ...
}
```

The JSON itself is correct — the warning is on stderr, and `setup` reported the plugin healthy (codex-cli 0.152.1, ChatGPT login active).

## Where

`--trace-deprecation` names the call site:

```
at runCommand      (scripts/lib/process.mjs:5:18)
at binaryAvailable (scripts/lib/process.mjs:39:18)
at buildSetupReport (scripts/codex-companion.mjs:184:22)
at handleSetup     (scripts/codex-companion.mjs:237:29)
```

`scripts/lib/process.mjs:12` defaults the shell option on Windows:

```js
shell: options.shell ?? (process.platform === "win32" ? (process.env.SHELL || true) : false),
```

so every bare command name spawned with args on Windows — `codex --version`, `node --version`, `npm --version` in `binaryAvailable` — trips DEP0190. `scripts/lib/git.mjs` already passes `shell: false` explicitly and does not.

## Why it is worth fixing rather than ignoring

Two reasons beyond noise:

1. **It is the warning's actual subject.** With `shell: true` the args are concatenated into a command line rather than escaped. These particular args are literals (`--version`), so nothing is exploitable today; the pattern is what Node is deprecating, and the same helper is what other call sites reach for.
2. **stderr belongs to the caller.** These scripts are invoked as subprocesses — by hooks, by slash commands, by other agents. A consumer that merges stderr into stdout (`2>&1`, or a hook harness that does it for you) hands its JSON parser a `(node:NNNN)` line and the call fails with nothing in the log to explain it. Structured output is a contract; anything unconditional on the shared stream is part of what the contract delivers.

## Possible fixes

- Resolve the binary path and spawn with `shell: false` (Windows can spawn a `.cmd`/`.exe` directly when the path is absolute), or
- keep the shell but pass a single pre-composed command string with no `args` array, or
- set `process.noDeprecation = true` in the entry scripts if the shell default is deliberate — cheapest, though it silences the warning rather than the pattern.

## Environment

- codex plugin 1.0.6 (`openai-codex` marketplace), installed at `.../cache/openai-codex/codex/1.0.6`
- Node v24.14.1, npm 11.11.0, Windows 11 26200
- codex-cli 0.152.1
- Reproduces on every `codex-companion.mjs` subcommand I ran, `setup --json` included

## Repro

```bash
node "<plugin>/scripts/codex-companion.mjs" setup --json
# warning appears on stderr before the JSON
node --trace-deprecation "<plugin>/scripts/codex-companion.mjs" setup --json
# names process.mjs:5 -> binaryAvailable -> buildSetupReport
```

fscfede-beep:
## Summary

Fixes #717.

Related: #708, #525.

- stop defaulting shared process launches to `shell: true` on Windows
- route `node` / `npm` / `codex` shim-style probes through an explicit `ComSpec /d /s /c call ...` invocation with `shell: false`
- use the same Windows invocation for direct `codex app-server` startup
- invoke `taskkill` directly with `shell: false`, avoiding Git Bash/MSYS conversion of `/PID`, `/T`, and `/F`
- treat `taskkill` exit 128 as an already-stopped process, so cleanup does not depend on English error text
- add regression coverage for the Windows shim invocation, taskkill options, and localized missing-process output

This avoids Node's DEP0190 warning without breaking `.cmd` shims such as `npm` on Windows. It also addresses the shell/argument-conversion root cause reported in #708 and #525 and makes the already-dead `taskkill` path locale-independent.

## Validation

On Windows 11 / Node 26.3.1:

- reproduced the pre-fix warning with `node --trace-deprecation plugins/codex/scripts/codex-companion.mjs setup --json`
- after the change, the same command exits 0 with `stderr_bytes=0`
- reproduced Git Bash/MSYS mangling on a nonexistent PID: `taskkill /PID 999999 /T /F` reaches `taskkill` as `C:/Program Files/Git/PID` and exits 1
- verified direct, no-shell `taskkill /PID 999999 /T /F` returns localized Spanish `ERROR: no se encontró el proceso "999999".` with exit 128
- `node --test tests/process.test.mjs tests/commands.test.mjs tests/bump-version.test.mjs` -> 14 passed, 0 failed
- `node --check plugins/codex/scripts/lib/process.mjs`
- `node --check plugins/codex/scripts/lib/app-server.mjs`
- `git diff --check`

I also verified on the same host that direct `spawnSync(..., { shell: false })` works for `.exe` binaries but returns `EINVAL` for an absolute `npm.cmd`, which is why the patch keeps explicit `cmd.exe` handling for Windows command shims instead of globally assuming every command is directly executable.

### DIFF
--- a/plugins/codex/scripts/lib/app-server.mjs
+++ b/plugins/codex/scripts/lib/app-server.mjs
@@ -14,7 +14,7 @@ import { spawn } from "node:child_process";
 import readline from "node:readline";
 import { parseBrokerEndpoint } from "./broker-endpoint.mjs";
 import { ensureBrokerSession, loadBrokerSession } from "./broker-lifecycle.mjs";
-import { terminateProcessTree } from "./process.mjs";
+import { commandWithWindowsShim, terminateProcessTree } from "./process.mjs";
 
 const PLUGIN_MANIFEST_URL = new URL("../../.claude-plugin/plugin.json", import.meta.url);
 const PLUGIN_MANIFEST = JSON.parse(fs.readFileSync(PLUGIN_MANIFEST_URL, "utf8"));
@@ -187,11 +187,12 @@ class SpawnedCodexAppServerClient extends AppServerClientBase {
   }
 
   async initialize() {
-    this.proc = spawn("codex", ["app-server"], {
+    const invocation = commandWithWindowsShim("codex", ["app-server"]);
+    this.proc = spawn(invocation.command, invocation.args, {
       cwd: this.cwd,
       env: this.options.env ?? process.env,
       stdio: ["pipe", "pipe", "pipe"],
-      shell: process.platform === "win32" ? (process.env.SHELL || true) : false,
+      shell: invocation.shell,
       windowsHide: true
     });
 
--- a/plugins/codex/scripts/lib/process.mjs
+++ b/plugins/codex/scripts/lib/process.mjs
@@ -100,7 +100,7 @@ export function terminateProcessTree(pid, options = {}) {
     }
 
     const combinedOutput = `${result.stderr}\n${result.stdout}`.trim();
-    if (!result.error && looksLikeMissingProcessMessage(combinedOutput)) {
+    if (!result.error && (result.status === 128 || looksLikeMissingProcessMessage(combinedOutput))) {
       return { attempted: true, delivered: false, method: "taskkill", result };
     }
 
--- a/plugins/codex/scripts/lib/process.mjs
+++ b/plugins/codex/scripts/lib/process.mjs
@@ -9,7 +9,7 @@ export function runCommand(command, args = [], options = {}) {
     input: options.input,
     maxBuffer: options.maxBuffer,
     stdio: options.stdio ?? "pipe",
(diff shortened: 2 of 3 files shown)


sylvesterkaczmarek (commented):
Explicit `ComSpec` handling is the right Windows boundary: `.cmd` shims still need command-processor semantics, but the child-process API no longer relies on `shell:true`. Keeping `taskkill` direct also avoids MSYS argument rewriting, and exit 128 removes the locale dependency.

PR #730, opened 2026-09-04
fix(task): optionally consume prompt files

### LINKED ISSUE #622: task --prompt-file: the prompt path is discarded after reading, so no plugin cleanup can ever remove the caller's file
**Version:** plugin 1.0.6 (`scripts/codex-companion.mjs`, `scripts/session-lifecycle-hook.mjs`), observed on Windows 11, Node 24.

**Summary.** When a prompt is supplied with `--prompt-file`, the companion converts the path into content and never retains the path. As a result no plugin lifecycle can ever remove that file — not the 50-job prune, not the SessionEnd cleanup. The caller is the only party able to delete it, and any cleanup the caller schedules runs in the very process that may die. An abrupt shell, session, or host termination between file creation and the caller's own cleanup step leaves the prompt on disk indefinitely, with nothing to scavenge it.

**Where:**

- `readTaskPrompt` (`codex-companion.mjs:643-645`) returns the file's contents as a string. The path is used for the read and then discarded; it is not stored on the job or anywhere else.
- `buildTaskRequest` (`codex-companion.mjs:604-610`) puts that CONTENT into `request.prompt`.
- `enqueueBackgroundTask` (`codex-companion.mjs:695-697`) writes that request into the job record.
- `cleanupSessionJobs` (`session-lifecycle-hook.mjs:42-75`) filters `state.jobs` by `sessionId` (line 54), terminates process trees for still-`queued`/`running` jobs (lines 59-69), and rewrites state without them (lines 71-74). Job records and their logs only — it has no prompt path to act on, so it cannot touch the caller's file.

**Impact.** Prompt files routinely carry sensitive material: repository excerpts, client content, internal notes. The plugin's own copies of that text have a retention story a user can reason about — pruned past 50 jobs, and wiped for the whole session at SessionEnd. The caller's file has no lifecycle at all. This is the part most likely to surprise: a user who sees the session-end cleanup reasonably assumes the dispatch was cleaned up, when the file the plugin told them to create is precisely what survives.

Note the asymmetry with the review path — `adversarial-review` takes its focus positionally and requires no file. Only the task path obliges the caller to materialize prompt text on disk, which is what creates the residue.

We currently document this to our own operators as best-effort cleanup that cannot survive abrupt termination, because no caller-side fix exists: cleanup code cannot run after its own process is gone.

**Suggested fixes** (the first is smallest and opt-in, so it cannot surprise existing callers):

1. **`--prompt-file-consume`** — unlink the file immediately after a successful read. Opt-in, one line at the read site, and it collapses the exposure window from "until the caller gets around to it" to "the moment the content is in memory".
2. **Record the resolved prompt path on the job record** and delete it in the same SessionEnd sweep that already removes the job, so an abrupt death is covered by the existing cleanup rather than by the caller.
3. **At minimum, document it** — state explicitly that the caller owns the file's lifetime and that no plugin cleanup will ever touch it. That alone would correct the reasonable assumption described above.

Options 1 and 2 are complementary: 1 covers the normal path, 2 covers a death between file creation and the read.

Filed alongside a companion report about the same flag's verification behaviour, and by the same reporter as #620.

fscfede-beep:
## Summary

Fixes #622 by adding an opt-in `--prompt-file-consume` flag to `task`.

When the flag is used with `--prompt-file`, the companion now reads the prompt, unlinks the caller-provided file, and only then proceeds to task dispatch. Without the flag, existing `--prompt-file` behavior is unchanged.

## Semantics

The ordering is deliberately:

`READ -> DELETE -> DISPATCH`

- if the read fails, no deletion is attempted;
- if the unlink fails, task dispatch does not proceed;
- without `--prompt-file-consume`, the prompt file remains caller-owned as before;
- `--prompt-file-consume` without `--prompt-file` fails with a clear error.

This implements the smallest opt-in fix proposed in #622 and does not mix in prompt digest verification or SessionEnd lifecycle changes.

## Regression coverage

Added three runtime tests:

1. consume mode deletes the prompt file after a successful read and the exact prompt still reaches `turn/start`;
2. ordinary `--prompt-file` preserves the file;
3. consume mode without `--prompt-file` is rejected.

RED against `upstream/main` (`db52e28f4d9ded852ab3942cea316258ae4ef346`): 1 pass / 2 fail, with the file remaining and the invalid flag combination exiting 0.

GREEN with this patch: 3/3 pass.

## Validation

- focused prompt-file runtime tests: 3/3 pass
- non-runtime suites (`bump-version`, `commands`, `git`, `process`, `render`, `state`): 28/28 pass
- `node --check plugins/codex/scripts/codex-companion.mjs`: pass
- `npx tsc -p tsconfig.app-server.json`: pass
- `npm run check-version`: pass
- `git diff --check`: pass
- high-confidence secret/private-marker scan of the diff: pass

I also attempted the full Windows suite. It does not terminate cleanly because the current test harness leaves app-server broker/fake-Codex processes behind; the first visible failure, `createBrokerEndpoint uses Unix sockets on non-Windows platforms`, reproduces identically on an untouched `upstream/main` worktree. I therefore am not claiming a full-suite pass from this Windows host.

## Scope

Two files only:

- `plugins/codex/scripts/codex-companion.mjs`
- `tests/runtime.test.mjs`

No version bump or unrelated lifecycle changes.

### DIFF
--- a/plugins/codex/scripts/codex-companion.mjs
+++ b/plugins/codex/scripts/codex-companion.mjs
@@ -79,7 +79,7 @@ function printUsage() {
       "  node scripts/codex-companion.mjs setup [--enable-review-gate|--disable-review-gate] [--json]",
       "  node scripts/codex-companion.mjs review [--wait|--background] [--base <ref>] [--scope <auto|working-tree|branch>]",
       "  node scripts/codex-companion.mjs adversarial-review [--wait|--background] [--base <ref>] [--scope <auto|working-tree|branch>] [focus text]",
-      "  node scripts/codex-companion.mjs task [--background] [--write] [--resume-last|--resume|--fresh] [--model <model|spark>] [--effort <none|minimal|low|medium|high|xhigh>] [prompt]",
+      "  node scripts/codex-companion.mjs task [--background] [--write] [--resume-last|--resume|--fresh] [--model <model|spark>] [--effort <none|minimal|low|medium|high|xhigh>] [--prompt-file <path> [--prompt-file-consume]] [prompt]",
       "  node scripts/codex-companion.mjs transfer [--source <claude-jsonl>] [--json]",
       "  node scripts/codex-companion.mjs status [job-id] [--all] [--json]",
       "  node scripts/codex-companion.mjs result [job-id] [--json]",
@@ -641,8 +641,18 @@ async function executeTransfer(cwd, options = {}) {
 }
 
 function readTaskPrompt(cwd, options, positionals) {
+  const consumePromptFile = Boolean(options["prompt-file-consume"]);
+  if (consumePromptFile && !options["prompt-file"]) {
+    throw new Error("--prompt-file-consume requires --prompt-file.");
+  }
+
   if (options["prompt-file"]) {
-    return fs.readFileSync(path.resolve(cwd, options["prompt-file"]), "utf8");
+    const promptPath = path.resolve(cwd, options["prompt-file"]);
+    const prompt = fs.readFileSync(promptPath, "utf8");
+    if (consumePromptFile) {
+      fs.unlinkSync(promptPath);
+    }
+    return prompt;
   }
 
   const positionalPrompt = positionals.join(" ");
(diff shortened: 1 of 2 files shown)


sylvesterkaczmarek (commented):
`READ -> DELETE -> DISPATCH` is the correct ownership order for an opt-in consume mode: a failed read preserves the caller's file, a failed unlink prevents execution, and successful dispatch cannot leave the supposedly consumed prompt behind. Keeping ordinary `--prompt-file` unchanged avoids surprising existing callers. The three regressions pin all of those branches cleanly.

PR #732, opened 2026-09-04
fix: make native review JSON contract explicit

### LINKED ISSUE #679: review --json emits no `result` (and no `parseError`): the built-in reviewer path never produces schema-shaped output, unlike adversarial-review
## Summary

In 1.0.6, `codex-companion.mjs review --json` returns a payload with **no `result` key at all** — and no `rawOutput` or `parseError` either, so a consumer cannot distinguish "structured output was not produced" from "structured output failed to parse". `adversarial-review --json` on the same repository, same working tree, returns `result` conforming to the plugin's own `schemas/review-output.schema.json`.

The two subcommands therefore emit two different `--json` shapes, and only one of them is machine-readable. Nothing in the payload marks which shape a consumer received.

I hit this building an automated benchmark that scores the plugin's reviewer against planted ground truth: the harness reads `payload.result`, and every case skipped with `no result in payload` while the companion itself exited 0 and looked healthy.

## Repro

Minimal git repo, one file, one breaking change in the working tree (the change deletes `module.exports`):

```bash
node scripts/codex-companion.mjs review --scope working-tree --json --cwd /path/to/repo
```

Top-level keys of the emitted JSON:

```
review, target, threadId, sourceThreadId, codex
```

`codex.stdout` holds prose:

> The change removes the module's only export, breaking all existing consumers that import the function.
> Review comment:
> - [P1] Restore the module export — …

Same repo, same working tree, other subcommand:

```bash
node scripts/codex-companion.mjs adversarial-review --scope working-tree --json --cwd /path/to/repo
```

Top-level keys:

```
review, target, threadId, context, codex, result, rawOutput, parseError, reasoningSummary
```

with `parseError: null` and

```json
{"verdict":"needs-attention","summary":"Do not ship: the change removes the module's public export…","findings":[{"severity":"high","title":"Function is no longer exported", …}]}
```

## Why the shapes differ

`executeReviewRun` branches on the review name (`scripts/codex-companion.mjs`, ~L357–L455):

- `reviewName === "Review"` → `runAppServerReview(...)`, i.e. codex's built-in reviewer. No `outputSchema` is passed. The payload is assembled as `{ review, target, threadId, sourceThreadId, codex: { status, stderr, stdout: result.reviewText, reasoning } }` — there is no `result` field in that object literal.
- any other review name (e.g. `Adversarial Review`) → the prompt-template path, which passes `outputSchema: readOutputSchema(REVIEW_SCHEMA)` and runs the response through `parseStructuredOutput`, producing `result` / `rawOutput` / `parseError`.

So the absence is by construction, not a parse failure. `parseStructuredOutput` itself always returns `parsed` and `parseError` (null on failure) — it is simply never called on the native path.

## Impact

- `schemas/review-output.schema.json` ships with the plugin but the default `review` never satisfies it.
- A `--json` consumer must special-case the review name to know whether findings are readable, and there is no field in the payload that says which path produced it.
- The failure is silent in the direction that matters: no `result`, and no `parseError` explaining its absence.

This looks distinct from #496, which is about the schema path *failing to conform* on long multi-tool-call turns; here the schema path is not taken at all. #496 also describes `review` as sharing the `runAppServerTurn` path, which no longer matches 1.0.6.

## Suggestions (any one would resolve it for a consumer)

1. Map the built-in reviewer's findings into the same `review-output.schema.json` shape and emit them as `result` on the `review` path.
2. Failing that, emit an explicit `result: null` plus a `parseError`-style reason (e.g. `"built-in reviewer returns prose; use adversarial-review for structured output"`) so the absence is stated rather than inferred.
3. And/or document the two `--json` shapes, keyed by subcommand, so consumers can branch deliberately.

## Environment

- codex plugin 1.0.6 (installed via the Claude Code marketplace)
- codex-cli 0.149.0
- Node v24.14.1
- Windows 11 (10.0.26200)

fscfede-beep:
## Summary

Fixes #679.

The built-in `review` path intentionally returns prose rather than schema-shaped findings, but its `--json` payload previously omitted `result`, `rawOutput`, and `parseError` entirely. Consumers could not distinguish "no structured output exists" from a parse failure.

This makes the contract explicit without heuristically parsing reviewer prose:

- `result: null`
- `rawOutput`: the native review text
- `parseError`: an actionable explanation that built-in review returns prose and `adversarial-review` should be used for structured output
- `reasoningSummary`: exposed consistently with the structured review path

Human-readable review output is unchanged.

## Validation

Windows 11 / Node 26.3.1:

- test-first reproduction failed on `main` because `payload.result` was `undefined`
- native JSON contract + existing review regressions: **3 passed, 0 failed**
- `node --check plugins/codex/scripts/codex-companion.mjs`
- `git diff --check`

### DIFF
--- a/plugins/codex/scripts/codex-companion.mjs
+++ b/plugins/codex/scripts/codex-companion.mjs
@@ -382,7 +382,11 @@ async function executeReviewRun(request) {
         stderr: result.stderr,
         stdout: result.reviewText,
         reasoning: result.reasoningSummary
-      }
+      },
+      result: null,
+      rawOutput: result.reviewText,
+      parseError: "The built-in reviewer returns prose; use adversarial-review for structured output.",
+      reasoningSummary: result.reasoningSummary
     };
     const rendered = renderNativeReviewResult(
       {
--- a/tests/runtime.test.mjs
+++ b/tests/runtime.test.mjs
@@ -157,6 +157,23 @@ test("review renders a no-findings result from app-server review/start", () => {
   assert.match(result.stdout, /No material issues found/);
 });
 
+test("review --json explicitly reports that native review output is unstructured", () => {
+  const repo = makeTempDir();
+  const binDir = makeTempDir();
+  installFakeCodex(binDir);
+  initGitRepo(repo);
+  fs.writeFileSync(path.join(repo, "README.md"), "hello\n");
+  run("git", ["add", "README.md"], { cwd: repo });
+  run("git", ["commit", "-m", "init"], { cwd: repo });
+  fs.writeFileSync(path.join(repo, "README.md"), "hello again\n");
+  const result = run("node", [SCRIPT, "review", "--json"], { cwd: repo, env: buildEnv(binDir) });
+  assert.equal(result.status, 0, result.stderr);
+  const payload = JSON.parse(result.stdout);
+  assert.equal(payload.result, null);
+  assert.equal(payload.rawOutput, payload.codex.stdout);
+  assert.match(payload.parseError, /built-in reviewer.*prose.*adversarial-review/i);
+});
+
 test("task runs when the active provider does not require OpenAI login", () => {
   const repo = makeTempDir();
   const binDir = makeTempDir();


sylvesterkaczmarek (commented):
Making the JSON contract explicit is preferable to heuristic prose parsing. `result: null` distinguishes “no structured result exists” from a successful empty findings set, while `rawOutput` preserves the native reviewer output and `parseError` tells structured consumers which command owns that contract. Mirroring `reasoningSummary` at the same level also keeps the native and adversarial paths easier to consume consistently.

PR #734, opened 2026-09-04
fix: ignore recommended plugin context in thread titles

### LINKED ISSUE #726: Codex history shows <recommended_plugins> instead of session titles
## Description

CC GUI displays Codex history entries with titles such as:

```text
<recommended_plugins> Here is a list of plugi...

fscfede-beep:
## Summary

Fixes #726.

Codex Companion task threads are explicitly named from the task prompt. When Claude prepends its injected `<recommended_plugins>...</recommended_plugins>` context block, that metadata currently becomes the visible Codex history title.

This patch strips only a leading `recommended_plugins` block before deriving the thread-name excerpt. The task prompt itself is unchanged and still reaches Codex verbatim; only the display-name source is sanitized.

## Validation

Windows 11 / Node 26.3.1:

- test-first reproduction via `--prompt-file`: **RED on `upstream/main`** with thread name `Codex Companion Task: <recommended_plugins> Here is a list of plugins. </re...`
- after the patch: `Codex Companion Task: Fix the flaky login test`
- targeted naming / resume-last / background runtime: **3 passed, 0 failed**
- `node --check plugins/codex/scripts/lib/codex.mjs`
- `git diff --check`

Fresh collision check immediately before publication found no open PR for #726.

### DIFF
--- a/plugins/codex/scripts/lib/codex.mjs
+++ b/plugins/codex/scripts/lib/codex.mjs
@@ -105,7 +105,11 @@ function looksLikeVerificationCommand(command) {
 }
 
 function buildTaskThreadName(prompt) {
-  const excerpt = shorten(prompt, 56);
+  const titleSource = String(prompt ?? "").replace(
+    /^\s*<recommended_plugins>[\s\S]*?<\/recommended_plugins>\s*/i,
+    ""
+  );
+  const excerpt = shorten(titleSource, 56);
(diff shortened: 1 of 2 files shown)


sylvesterkaczmarek (commented):
Keeping the sanitisation title-only is the correct boundary: the task prompt still reaches Codex verbatim, so execution context is unchanged. Anchoring at the beginning prevents removing user text later in the prompt, and the non-greedy close-tag match limits stripping to the injected `recommended_plugins` block. The regression exercises the real task/thread path rather than only the helper. No blocker.

PR #736, opened 2026-09-04
Fix: retry stdin reads on EAGAIN instead of crashing the hook

anthonyytao:
Root-caused from a live SessionStart hook failure: `Failed with non-blocking status code: EAGAIN: resource temporarily unavailable, read`.

`fs.readFileSync(0, "utf8")` throws `EAGAIN` when stdin is piped (non-TTY) and the fd is still non-blocking at the moment of the read, before the writer has finished flushing — a known Node/libuv race, not a Claude Code daemon issue. It hit `session-lifecycle-hook.mjs` directly, and `stop-review-gate-hook.mjs` / `lib/fs.mjs`'s `readStdinIfPiped()` shared the identical unguarded pattern.

## Fix
Adds `readStdinSync()` to `lib/fs.mjs`: a `fs.readSync` loop that catches `EAGAIN`, waits briefly (`Atomics.wait`, 20ms), and retries, rather than letting it escape as an uncaught exception. Routed all three call sites through it instead of the bare `readFileSync(0)`.

## Testing
- `node --check` on all three edited files
- Unit test mocking `fs.readSync` to throw `EAGAIN` once then succeed — confirms the retry loop recovers and returns the correct data
- Functional smoke test: valid piped JSON through `session-lifecycle-hook.mjs` and `stop-review-gate-hook.mjs` — unchanged behavior, exit 0

### DIFF
--- a/plugins/codex/scripts/lib/fs.mjs
+++ b/plugins/codex/scripts/lib/fs.mjs
@@ -32,9 +32,35 @@ export function isProbablyText(buffer) {
   return true;
 }
 
+// fs.readFileSync(0) can throw EAGAIN on a piped (non-TTY) stdin whose
+// underlying fd is still non-blocking when the read fires, before the
+// writer has finished flushing. Retry on EAGAIN instead of letting it
+// escape as a hook failure.
+export function readStdinSync() {
+  const chunks = [];
+  const buffer = Buffer.alloc(65536);
+  for (;;) {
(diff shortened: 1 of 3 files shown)


anthonyytao:
Opened by mistake — meant to target our own fork (anthonyytao/codex-plugin-cc), not upstream. Closing; the change will be re-opened there instead. Apologies for the noise.

PR #739, opened 2026-09-05
Add an explicit --network flag for read-only tasks

naniastudios:
# Title

Add an explicit network flag for read-only mandates

## Problem

In Codex CLI 0.147.0, the read-only sandbox has no network access. As a
result, read-only mandates that need to reach the network fail with a
misleading DNS error. The existing `--write` flag cannot express a mandate
that needs network access but must not write to the workspace.

## Solution

Add an explicit `--network` boolean flag, separate from `--write`. When
`--network` is selected, task startup uses `workspace-write` and explicitly
enables `sandbox_workspace_write.network_access`; `--write` remains
independent, and the default behavior is unchanged when neither flag is
provided.

## Verification

- Red: before the patch, a task started without `--network` uses the
  `read-only` sandbox and a network probe fails with the sandbox's DNS error.
- Green: after the patch, a task started with `--network` uses
  `workspace-write` with network access enabled and the same network probe
  succeeds; a task without `--network` remains network-disabled.
- `node --check` passes for `scripts/codex-companion.mjs` and
  `scripts/lib/codex.mjs`.

### DIFF
--- a/plugins/codex/scripts/codex-companion.mjs
+++ b/plugins/codex/scripts/codex-companion.mjs
@@ -79,7 +79,7 @@ function printUsage() {
       "  node scripts/codex-companion.mjs setup [--enable-review-gate|--disable-review-gate] [--json]",
       "  node scripts/codex-companion.mjs review [--wait|--background] [--base <ref>] [--scope <auto|working-tree|branch>]",
       "  node scripts/codex-companion.mjs adversarial-review [--wait|--background] [--base <ref>] [--scope <auto|working-tree|branch>] [focus text]",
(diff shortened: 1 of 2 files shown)


sylvesterkaczmarek (commented):
This makes `--network` violate the read-only guarantee it is intended to preserve. When `network=true` and `write=false`, `executeTaskRun()` selects the `workspace-write` sandbox, and the added config only enables network access inside that sandbox; it does not remove workspace write permission. A read-only task that asks for network can therefore modify the workspace. I would not use `workspace-write` as the transport for this flag unless the permission profile explicitly denies writes. If Codex 0.147.0 cannot express read-only plus network directly, this should fail closed or use a request-scoped filesystem profile that keeps the workspace read-only while enabling network. A regression should attempt an actual workspace write under `--network` without `--write` and require it to fail.

Issue #671, opened 2026-08-23
Shared per-workspace broker is torn down by any session's SessionEnd, killing other sessions' in-flight jobs


bwish71:
## Summary

The companion broker is keyed on the **workspace**, but its teardown is triggered by an **individual session** ending, with no check for other sessions' in-flight work. When several Claude Code sessions run in the same checkout — normal on a shared repo — any session closing kills whatever job the others are running. The victim sees `codex app-server connection closed.`, and for the stop-review gate this surfaces as its session-stop being blocked by a failure it did not cause.

Observed as an intermittent stop-gate failure: **3 of 7 runs died over one day, with no correlation to run duration** — a 6-minute run survived while ~5-minute runs died. The variable was never the run; it was whether another session happened to close during it.

## Environment

- plugin **1.0.6**, Node v24, Linux
- 7 Claude Code sessions live in one git checkout (verified: every one of their `/proc/<pid>/cwd` resolves to the same repo root)
- one broker serving them all: `app-server-broker.mjs serve --endpoint unix:/tmp/cxc-XXXXXX/broker.sock --cwd <repo>`, reparented to PID 1

## Bug 1 (primary) — unguarded workspace-wide teardown

`scripts/session-lifecycle-hook.mjs` → `handleSessionEnd()` resolves the broker via `loadBrokerSession(cwd)`, which reads `broker.json` from the **workspace-keyed** state dir (`lib/state.mjs` `resolveStateDir`). It then unconditionally calls `sendBrokerShutdown(endpoint)` and `teardownBrokerSession({... killProcess: terminateProcessTree})`, which kills the pid, unlinks socket/pid/log files, and rmdirs the session dir.

`lib/broker-lifecycle.mjs` `teardownBrokerSession` has **no refcount, no owning-session check, and no in-flight-job check**. `cleanupSessionJobs()` correctly scopes itself to the ending session's own job rows — but the broker kill beside it is workspace-wide.

Victim path: `BrokerCodexAppServerClient`'s socket fires `close` → `handleExit(this.exitError)` with `exitError === null` → every pending request rejects with `new Error("codex app-server connection closed.")` (`lib/app-server.mjs`). **No reconnect, no retry.**

**Expected:** a session ending does not disturb work owned by another session.
**Actual:** it terminates it, surfacing as an opaque connection-closed.

## Bug 2 — the broker honours `broker/shutdown` mid-turn

[...]

Issue #474, opened 2026-07-10
Increase `SessionEnd` hook timeout to prevent premature cancellation


0x221A:
The `SessionEnd` lifecycle hook currently has a timeout of only 5 seconds in [hooks.json](https://github.com/openai/codex-plugin-cc/blob/main/plugins/codex/hooks/hooks.json#L21). In some environments, the hook takes longer than this to complete, causing it to time out and be cancelled before session cleanup finishes.

Please increase the `SessionEnd` timeout, for example, from 5 seconds to 30 seconds while leaving the `SessionStart` timeout unchanged.

**Expected behavior**

The `SessionEnd` hook has enough time to complete normally without being prematurely terminated.

PR #755, opened 2026-09-09
fix(hooks): align SessionEnd timeout with Codex runtime cap

james-meneses:
## Summary
- set the SessionEnd hook timeout to the runtime hard cap of 3 seconds
- bound the broker shutdown request at 750 ms so teardown stays inside that budget
- add focused manifest and non-responsive-broker regression tests

## Evidence
- RED manifest: SessionEnd timeout 5s exceeds the runtime 3s hard cap
- RED stalled broker: focused test remained pending until external timeout, rc=124
- `node --test tests/broker-lifecycle.test.mjs tests/hooks.test.mjs`: 2 pass
- `npm test`: 93 pass
- Desktop 0.153.4 `hooks/list`, before: effective `timeoutSec: 3` plus the clamp warning
- fresh isolated home from this head, after: effective `timeoutSec: 3`, `warnings: []`, `errors: []` in two project roots

## Why
The runtime enforces `SESSION_END_MAX_TIMEOUT_SEC = 3` to retain headroom inside its five-second app-server shutdown bound. A plugin value of 5 is valid input but is always clamped, producing a warning in every project where the globally enabled plugin is loaded. The prior shutdown request could also wait indefinitely after a broker accepted without replying, so lowering only the manifest value would create a false-green contract.

## Scope
This bounds the request and removes the manifest/runtime mismatch. It does not change shared-broker ownership or cross-session shutdown policy.

Related: #474, #540, #671

### DIFF
--- a/plugins/codex/hooks/hooks.json
+++ b/plugins/codex/hooks/hooks.json
@@ -18,7 +18,7 @@
           {
             "type": "command",
             "command": "node \"${CLAUDE_PLUGIN_ROOT}/scripts/session-lifecycle-hook.mjs\" SessionEnd",
-            "timeout": 5
+            "timeout": 3
           }
         ]
       }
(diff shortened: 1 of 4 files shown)


james-meneses:
@codex review

